PR #2, opened 2017-05-26
feat(site): independently scrolling content and side-nav

dborodyansky:


packages/site/layouts/chrome.module.css
```diff
@@ -23,10 +23,14 @@
   }
   .side {
     min-width: 200px;
+    height: calc(100vh - 8px);
```
jaketrent: Isn't there a var(--topBarHeight) var we could use here?

packages/site/layouts/chrome.module.css
```diff
@@ -23,10 +23,14 @@
...
+    overflow: auto;
   }
   .main {
     flex: 1;
     border-left: 1px solid var(--psColorsGray01);
+    height: calc(100vh - 8px);
```
jaketrent: ditto

jaketrent:
I love it!

jaketrent:
Dmitry and I decided not to do independent nav scrolling because it's a document with natural scrolling expectations and we don't want to style windows scrollbars.

PR #54, opened 2017-08-03
text styles fix

iammerrick:


jaketrent:
Nice catch!  Thanks for the fix.

Just an fyi on the commit message format: https://github.com/pluralsight/design-system/blob/master/CONTRIBUTING.md#commit-messages

I need to automate that convention so we don't have to worry about it.

PR #82, opened 2017-08-18
refactor(row): spacing and line-height adjustment

dborodyansky:
A few spacing and line-height adjustment to match design.

There was a typo (missing .core.) that was causing title line-height not work.

Also, suggesting flex box center VS. padding for `words`, so that the text is centered regardless of string length or metadata inclusion.

### DIFF
--- a/packages/row/src/react/index.js
+++ b/packages/row/src/react/index.js
@@ -225,13 +225,15 @@ const formatActionBarWidth = ({ actionBar }) =>
     ? `(${actionBar.length} * ${actionBarActionWidth} + ${actionBar.length} * ${actionBarActionMarginLeft})`
     : '0px'
 
-const Words = glamorous.div({ flex: 1 }, props => ({
+const Words = glamorous.div({
+    display: 'flex',
+    flex: 1,
+    flexDirection: 'column',
+    alignSelf: 'center'
+  }, props => ({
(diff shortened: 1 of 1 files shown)


packages/row/src/react/index.js
```diff
@@ -228,25 +228,29 @@ const formatActionBarWidth = ({ actionBar }) =>
 const Words = glamorous.div({ flex: 1 }, props => ({
   maxWidth: `calc(100% - ${formatImageWidth(props)} - ${formatActionBarWidth(
     props
-  )})`
+  )})`,
```
jaketrent: these static styles can go on line 228, but `flex: 1`.  The later args to glamorous components are overrides.  I think this'll work.  Just convention-wise/readability-wise, I think we move it up.

packages/row/src/react/index.js
```diff
@@ -228,25 +228,29 @@ const formatActionBarWidth = ({ actionBar }) =>
...
+  alignSelf: 'center'
 }))
 
 const styleTitleSize = ({ size }) =>
   ({
     [sizes.small]: {
+      // marginTop: core.layout.spacingTiny,
       fontSize: core.type.fontSizeSmall,
-      lineHeight: core.lineHeightTight
+      lineHeight: core.type.lineHeightTight
```
jaketrent: Ah!  nice catch.

PR #114, opened 2017-09-25
refactor(button): make flat button hover color alpha based

dborodyansky:
Changing the background color of the flat button to white with alpha, so it looks better on image backgrounds etc.

### DIFF
--- a/packages/button/src/react/index.js
+++ b/packages/button/src/react/index.js
@@ -1,6 +1,7 @@
 import core from '@pluralsight/ps-design-system-core'
 import glamorous from 'glamorous'
 import React from 'react'
+import { transparentize } from 'polished'
 
 const appearances = { stroke: 'stroke', flat: 'flat' }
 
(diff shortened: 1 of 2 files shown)


packages/button/src/react/index.js
```diff
@@ -59,7 +59,7 @@ const styleAppearance = ({ appearance }) =>
       background: 'none',
       ':hover': {
         color: core.colors.white,
-        background: core.colors.gray04
+        background: 'rgba(255, 255, 255, 0.15)'
```
jaketrent: @dborodyansky , Do you want to base this on the hard-coded rgb values, or would you rather base this on `core.colors.white`?  If the later, that's possible.  See the card's usage of:

```
import { transparentize } from 'polished'
```

If the former, let's merge this.

dborodyansky: The former is preferred, just didn't know it was possible :) let me give it a go.

jaketrent:
Looks awesome!  Merged.

PR #133, opened 2017-11-01
link: add typescript typings for react component

emandres:
### What You're Solving

- This change adds typescript typings for the React Link component

### Design Decisions

- I used a union type for two `appearance` values. I'm pretty new to typescript, so I'm not sure if that's a good/bad pattern. The downside to it is that if an appearance style is added, it will need to be updated in the type definition. Upside is compile time assurance that the value is correct.

### How to Verify

- The only way I really know to verify this is to pull it into a typescript/react app and see that it compiles. I've done this on the project I'm using the design system on and it seems to be working.

### DIFF
--- a/packages/link/react.d.ts
+++ b/packages/link/react.d.ts
@@ -1,0 +1,10 @@
+/// <reference types="react" />
+
+export type Appearance = 'default' | 'subtle'
+
+export interface LinkProps {
+  appearance?: Appearance
+}
+
+export default function Link(props: LinkProps): React.Component<LinkProps>
+
--- a/packages/link/react.d.ts
+++ b/packages/link/react.d.ts
@@ -1,6 +1,6 @@
 /// <reference types="react" />
 
-export type Appearance = 'default' | 'subtle'
+type Appearance = 'default' | 'subtle'
 
 export interface LinkProps {
   appearance?: Appearance


packages/link/react.d.ts
```diff
@@ -0,0 +1,10 @@
+/// <reference types="react" />
+
+export type Appearance = 'default' | 'subtle'
+
+export interface LinkProps {
```
jaketrent: Is the *Props name a convention?

packages/link/react.d.ts
```diff
@@ -0,0 +1,10 @@
+/// <reference types="react" />
+
+export type Appearance = 'default' | 'subtle'
```
jaketrent: What is conventionally exported from a typings file like this?  Everything?

packages/link/react.d.ts
```diff
@@ -0,0 +1,10 @@
+/// <reference types="react" />
```
jaketrent: what does `types=react` control?

packages/link/react.d.ts
```diff
@@ -0,0 +1,10 @@
...
+
+export type Appearance = 'default' | 'subtle'
+
+export interface LinkProps {
+  appearance?: Appearance
+}
+
+export default function Link(props: LinkProps): React.Component<LinkProps>
```
jaketrent: Is `React.Component` known within this file?  How is this resolved?

jaketrent:
also, btw, regarding the commit message, which follows this convention: https://conventionalcommits.org/

My thought is that what you want is something like `feat`, since this is a new user-exposed api.  This will help the versions get bumped appropriately.

packages/link/react.d.ts
emandres: That's the way I've seen it everywhere else

packages/link/react.d.ts
emandres: My understanding is that this references the types declared in the `@types/react`, making it so I can use `React.Component` down below.

packages/link/react.d.ts
emandres: See comment on the `/// <reference>` line

packages/link/react.d.ts
emandres: That I wasn't quite sure on. I think we could probably leave off Appearance and it would be fine. My experience with typescript type definitions is pretty sparse. Basically just got it to compile :smile:. I'll remove `Appearance` from exports.

emandres:
Wasn't sure on feat/chore. It seemed like a small thing, but since it does alter the published interface, feat does make more sense.

packages/link/react.d.ts
emandres: Did a little more research, just to make sure that `*Props` wasn't a homegrown convention. Every place I researched that gives examples of React + Typescript are using this naming convention.

emandres:
Also, on the commit message, do you normally squash before merging? I'm not actually seeing the option right now

jaketrent:
Merged!  I squashed to make a clean commit msg.  Sometimes I just rebase if multiple commits would be nice.

Also, fyi, in a later commit, I added a `typings` entry in the link/package.json file, which points to the .d.ts file.  Apparently this will help tooling find the definition.

PR #140, opened 2017-11-10
Row: prevent row image from shrinking

KaidenR:
### What You're Solving

In certain cases, the row image and actions are shrinking when the Words
section runs out of space (see attached screenshot). This fix makes sure the Words section will collapse all the way to zero before the other flex items start collapsing.

### How to Verify

We've spent a while trying to reproduce this exact behavior and have not found a consistent way to do so outside our specific use case. We tested to make sure the fix did not change any of the examples in the design system site.

#### Before Fix
 screen shot

#### After Fix
screen shot 2017-11-10 at 2 12 40 pm

### DIFF
--- a/packages/row/src/react/index.js
+++ b/packages/row/src/react/index.js
@@ -261,7 +261,8 @@ const Words = glamorous.div(
     display: 'flex',
     flex: 1,
     flexDirection: 'column',
-    alignSelf: 'center'
+    alignSelf: 'center',
+    minWidth: 0
   },
   props => ({
     maxWidth: `calc(100% - ${formatImageWidth(props)} - ${formatActionBarWidth(


jaketrent:
@KaidenR Thanks for finding this and finding a solution accommodate the case.  The solution is non-intuitive to me at first.  What makes this work?  Or do you have a url you could point me to to learn about it?

KaidenR:
@jaketrent The Flexbox spec says 
> By default, flex items won’t shrink below their minimum content size (the length of the longest word or fixed-size element). To change this, set the min-width or min-height property. (See §4.5 Automatic Minimum Size of Flex Items.)

So I'm pretty sure that was causing the Words section to reach it's automatic min size and then the image would start shrinking. So setting `min-width: 0;` would let it shrink all the way. 

You can check out the bug here: https://app-stage.pluralsight.com/channels/details/1eda599f-aefa-4372-a1a5-2d54a9088b3d
Shrink the window down and the row image will start collapsing. But add `min-width:0` to the Words section (`class="css-1kx0lgw"`) and it will be fixed.

PR #144, opened 2017-11-17
badge: extract css and vars modules

jaketrent:
Proof of concept for extracting css (and exported vars) from a design system component (badge, as the simple example).  It works in storybook and local reference site.  Feedback welcome.  

### Questions I'm Thinking About

- Is it worth it?  Good investment?  
- Are there better ways to do a similar thing?

### What You're Solving

- A potential way to allow sharing css between diff side-by-side or future component techs
- Or building/publishing raw css

### Design Decisions

- Plain JavaScript, compatible with that, at least
- JS could be built to JSON, converted to any other formats, including *.css

### DIFF
--- a/packages/badge/css.js
+++ b/packages/badge/css.js
@@ -1,0 +1,1 @@
+module.exports = require('./dist/css')
--- a/packages/badge/index.js
+++ b/packages/badge/index.js
@@ -1,3 +1,5 @@
+const css = require('./dist/css')
 const React = require('./dist/react')
+const vars = require('./dist/vars')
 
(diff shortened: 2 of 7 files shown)


jaketrent:
another note on the src lang: authoring in js is kinda awkward.  But a nice feature of it is that you can name the selectors based on the js-based prop enumerations that many of the styles are based on.  Easier to keep in sync this way.  Hopefully someday some better refactor tools too. :)

jaketrent:
best idea so far.  will support q1 alongside tech experimentation. merging.

Issue #166, opened 2018-01-19
button: innerRef prop is being ignored


jimboslicethat:
### Expected behavior
`Button.innerRef` should use the function passed to `props.innerRef`

### Actual behavior
`Button.innerRef` is statically set to `innerRef={el => (this.el = el)}`

It looks like this property is overwriting what the user of the component passed in. We also noticed that on line **328**, `this.props.innerRef` gets set to ref on the button anyways:
`...(this.props.innerRef ? { ref: this.props.innerRef } : {}),`

### Steps to repro

1. Pass in a custom function to Button.innerRef _(with something like a console log)_
2. Mount the component in the DOM and nothing happens.
3. Edit the package source code, removing line **353**.
4. Re-fresh the DOM and watch the magic happen.

### Related packages
n/a
### Environment
- OS - MacOS 10.12
- Browser version - Chrome v63

jaketrent:
Fixed in #168.

PR #167, opened 2018-01-19
fix(button): remove innerRef override

jimboslicethat:
### What You're Solving

- Fix for issue #166

### How to Verify

-  Pass in a custom function to `Button.innerRef` _(something like a console log)_

### DIFF
--- a/packages/button/src/react/index.js
+++ b/packages/button/src/react/index.js
@@ -350,7 +350,6 @@ class Btn extends React.Component {
       <Button
         {...props}
         iconOnly={React.Children.count(props.children) <= 0}
-        innerRef={el => (this.el = el)}
         style={isLoadingWithNoText ? { width: this.nonLoadingWidth } : {}}
         themeName={context.themeName}
       >


jaketrent:
Hey @jimboslicethat and @zlintz , I'm just getting a chance to look at the entire button src code, and I'm realizing that the button component needs to retain its own innerRef on the element because it is getting at the DOM node to measure its width in some cases.  

Because of this, I'm not going to be able to merge the PR as-is.  Rather, I need to do two things:

1. Keep the internal innerRef
2. Figure out how to allow the passing in of an innerRef for clients to use for their purposes.

I think I'll just do what is usually done in cases similar to this where there's something like an internal onClick that needs overrides:

```
const Button = props =>
  <button ...
     {...props}
     innerRef={el => {
        this.el = el
        if (typeof props.innerRef === 'function') props.innerRef(el)
     }} />
```

What do you think about this plan?

For reference, the line showing the need for an internal innerRef and saving of the node is here: https://github.com/pluralsight/design-system/blob/master/packages/button/src/react/index.js#L341

jimboslicethat:
Hey Jake. I think we both feel pretty good about your solution. Sorry, we probably could have done a little more digging to figure out what `this.el` was responsible for. Next week we will try that code out locally just to double check that things are working as expected, then update you! Thanks.

jaketrent:
Fixed in https://github.com/pluralsight/design-system/pull/168.

PR #170, opened 2018-01-22
fix(drawer): Add aria-label for expand/collapse

jim-cooper:
adds accessibility label to drawer toggle arrow.

### DIFF
--- a/packages/drawer/src/react/index.js
+++ b/packages/drawer/src/react/index.js
@@ -125,6 +125,8 @@ class Drawer extends React.Component {
 
   render() {
     const { isOpen, context: { themeName } } = this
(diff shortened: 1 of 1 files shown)


packages/drawer/src/react/index.js
```diff
@@ -125,6 +125,8 @@ class Drawer extends React.Component {
 
   render() {
     const { isOpen, context: { themeName } } = this
+    const ariaLabel = isOpen ? 'Collapse' : 'Expand'
```
jaketrent: @dborodyansky, quick ux question, is "collapse" and "expand" the language you'd like to go with here?

dborodyansky: 👍 Yes. Thank you.

PR #178, opened 2018-01-26
Avatar: Style changes and Gravatar friendly

assisantunes:
This PR implements the following:

- Change initials font-weight from `ExtraLight` to `Book`
- Adds a transparent image as default fallback on gravatar urls.
  - Add `[?|&]d=https://cdn/transparent.gif` param at the end of gravatar urls.

### Usage
```jsx
<Avatar src="https://secure.gravatar.com/..." name="Anakin Skywalker" />
```

When you pass a gravatar link that doesn't have a picture

Before             |  After
:-------------------------:|:-------------------------:
image | image


jaketrent:
just fyi, but that first commit that is a change of the font weight is probably a `refactor(avatar)` commit instead of `feat(avatar)`.  In this case, it'll make not diff, because the next commit is a legit `feat`.  But just to be clear, those help drive the semver version bumping.

PR #179, opened 2018-01-26
drawer: remove first drawer's border top

jaketrent:


jaketrent:
Hey @KaidenR and @bressain, what do you think about the movement of the borderTop  on the Drawer?  

This is being done to match the style of the Row, where in a sequence, the first Drawer doesn't have a borderTop.  The subsequent Drawers do.

Reviewing the last commit should be sufficient for this question.  

https://github.com/pluralsight/design-system/pull/179/commits/aeb127bacd379b4176039b151a86e3a88141796a

I accidentally put other local commits in this PR.  But for the sake of time I'm not removing them.

PR #181, opened 2018-01-27
Fix disabled button click

jim-cooper:
### What You're Solving

- Button: If you provide an href for a disabled Button (i.e. <Button href="..." disabled={true}>), the button click doesn't actually get disabled

### Design Decisions

- Used the `whitelistProps` to remove the onClick handler if `disabled` and `href` both exist


jim-cooper:
This was a bad branch (due to a bad rebase). Closing this in favor of a cleaner PR.

PR #182, opened 2018-01-29
Disabled button click fix

jim-cooper:
### What You're Solving

- Fix: Disabled buttons don't get disabled if they have an href prop
- When an href prop is passed, the button is rendered using an anchor tag. Anchor tags ignore the disabled attribute.

### Design Decisions

- I implemented this by removing (without mutating) onClick from the whitelisted props only when disabled and href are both passed.

### How to Verify

- Add a disabled prop to a Button that also has an href prop

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,4 +40,7 @@ jspm_packages
 # Optional REPL history
 .node_repl_history
 
+# VsCode Settings
+.vscode/
+
 .env*
--- a/packages/button/src/react/__specs__/index.spec.js
+++ b/packages/button/src/react/__specs__/index.spec.js
@@ -18,6 +18,26 @@ test('click on button triggered once', () => {
   expect(callCount).toBe(1)
 })
 
+test('click on disabled button with href does not trigger onClick', () => {
+  let callCount = 0
+  const onClick = _ => {
+    callCount += 1
+  }
+  const button = mount(
+    <Button
+      disabled={true}
+      onClick={onClick}
+      href="https://foo.com"
+      icon={<Icon id={Icon.ids.check} />}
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -40,4 +40,7 @@ jspm_packages
 # Optional REPL history
 .node_repl_history
 
-.env*
\ No newline at end of file
+# VsCode Settings
+.vscode/
```
jaketrent: philosophically, I think these kinds of "my environment" ignores should be in your own system dotfiles, not in the project's.  A la https://jaketrent.com/post/gitignore-system-files/

packages/button/src/react/__specs__/index.spec.js
```diff
@@ -18,6 +18,26 @@ test('click on button triggered once', () => {
   expect(callCount).toBe(1)
 })
 
+test('click on disabled button with href does not trigger onClick', () => {
+  let callCount = 0
+  const onClick = _ => {
+    callCount += 1
+  }
+  const button = mount(
```
jaketrent: `shallow` is faster and can be used for event simulation.

packages/button/src/react/index.js
```diff
@@ -322,12 +322,17 @@ const whitelistProps = (props, whitelist) =>
 
 class Button extends React.Component {
   render() {
+    const whitelistedProps =
```
jaketrent: I like that this is created as a local var.  Even better might be some "code as docs" with a function extraction called something like `preventOnClickForDisabledAnchorTag`.

jaketrent:
Published in v10.3.6

Issue #188, opened 2018-02-07
tag: change children propType from string to node


KaidenR:
### Desired behavior

We need to be able to control the way the text inside a tag is ellipsified to prevent a number at the end from overflowing:

![image](https://user-images.githubusercontent.com/679346/35938616-e26f5750-0c06-11e8-9421-728626739d66.png)

We can do this pretty easily by wrapping the text in a flex container:

![image](https://user-images.githubusercontent.com/679346/35938753-5da45948-0c07-11e8-8967-048277e2a0fb.png)

But the tag complains because it requires `children` to be a string

### The Tradeoffs

Only thing I can think of is the potential for tags that don't adhere to the design standards.

jaketrent:
I think the tradeoff that you point out is real.  We want users to only send in strings.  If we open our proptype to say something as generic as node, we're relying on users to be cognizant of the ability to now display a tag that doesn't match the styleguide.

I like how you've shown your use case could be a lot better if this was supported.

I think I'm in favor of merging your PR that addresses this.

@dborodyansky how do you like the tradeoffs?

PR #189, opened 2018-02-07
tag: change children propType from string to node

KaidenR:
### What You're Solving

- Allows passing a node as children instead of just a string. Makes it easier to
control the way children are ellipsified.
- Resolves issue #188 

### How to Verify

- Verify you can pass a node in as the children to a `tag` component


jaketrent:
Can I also just say that Kaiden, you and Zach here, and everyone on Devstar has been awesome at using and contributing back enhancements and fixes.  Thank you!  It's awesome to sense your investment.

jaketrent:
Chatted with Dmitry.  We're thumbs up.

PR #198, opened 2018-03-06
Change avatar image fallback strategy

jaketrent:


jaketrent:
why do I have 2 extra commits here? closing.

PR #199, opened 2018-03-06
fix(avatar): hide img or initials based on img load state

### LINKED ISSUE #191: avatar: color halo
### Expected behavior

No initials state color bleed through if there's a user profile displayed.  Just the user profile.

### Actual behavior

There's a halo that's visible if the profile pic is light (eg, white bg)

image 2

### Related packages

- avatar

jaketrent:
Help remove color halo that bled out from behind loaded images.  Had to
change the gravatar fallback strategy to support this.  Otherwise, the
rule was that we always show initials in the case of gravatar images.
But we don't want this, because depending on the color combo, the halo
is visible.  Instead, now we fail gravatar images like any other.

Closes #191

### DIFF
--- a/packages/avatar/src/react/index.js
+++ b/packages/avatar/src/react/index.js
@@ -16,25 +16,52 @@ const styles = {
   initials: ({ name }) => glamor.css(css['.psds-avatar__initials'])
 }
 
-const Avatar = ({ className, name, size, src, style }) => {
-  const avatarProps = {
-    ...styles.avatar({ size }),
-    ...(className ? { className } : null),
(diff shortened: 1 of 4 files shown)


packages/avatar/src/react/index.js
```diff
@@ -16,25 +16,52 @@ const styles = {
   initials: ({ name }) => glamor.css(css['.psds-avatar__initials'])
```
jaketrent: lines 40-41 show main checks
lines 26-31 are the new callbacks for 
lines 48-49 the new image loading event listeners

PR #200, opened 2018-03-06
refactor(button): whitelist role for button

QuietOmen:
In a spa we often wrap the button within a Link tag, and need a way to assign a role to the button. This pull-request adds `role` to the whitelist and allows it to be attached to the generated comp.


.gitignore
```diff
@@ -40,4 +40,7 @@ jspm_packages
 # Optional REPL history
 .node_repl_history
 
+# IDE files
```
jaketrent: I would propose backing out personal editor ignores based on this philosophy https://jaketrent.com/post/gitignore-system-files/

PR #201, opened 2018-03-07
refactor(avatar): color algo for variance in alpha-order names

### LINKED ISSUE #192: avatar: adjust color algo to be varied in alpha order
### Desired behavior

More random colors in an alpha-ordered avatar list

### Today's behavior

Color is based on name initials, so A.F. and A.S. will get the same default initial color.

### The Value Add

Random per person.  Should change more often between person name.

### The Tradeoffs

More work to get done.  Base on probably a hash of the entire name contents.  Maybe heavier codebase for the hashing function.  Nbd.

jaketrent:
One team found that we showed the same color over and over in a row when a list of avatars was shown for people in an alpha-ordered list.  This attempts to mix that color up.

Closes #192

### DIFF
--- a/packages/avatar/src/react/utils.js
+++ b/packages/avatar/src/react/utils.js
@@ -1,4 +1,4 @@
-import { colorByLetter, defaultColor, defaultGravatarImage } from '../vars'
+import { colorByLetter, defaultGravatarImage } from '../vars'
 
 export const getInitials = fullname => {
   if (!fullname || !fullname.length) return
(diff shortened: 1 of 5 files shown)


jaketrent:
from this:
screen shot 2018-03-07 at 11 32 44 am

to this:
screen shot 2018-03-07 at 11 32 31 am

packages/avatar/src/react/utils.js
```diff
@@ -12,12 +12,76 @@ export const getInitials = fullname => {
 }
 
 export const getColorByName = name => {
-  const initial = name[0].toUpperCase() || null
-  return colorByLetter[initial] || defaultColor
+  const stableRandLetter =
+    sha1(name)
+      .toUpperCase()
+      .replace(/[^A-Z]/, '')[0] || 'A'
+  return colorByLetter[stableRandLetter] || defaultColor
```
assisantunes: I think `defaultColor` probably never gonna be used on this case, once we have `'A'` as fallback for `stableRandLetter`, If so maybe remove it?

jaketrent: yes, good idea!

PR #202, opened 2018-03-08
Table: new component

rafaelcardoso:
This PR adds the Table component: `@pluralsight/ps-design-system-table`

### Features

- Column/header sorting
- Nested tables (or any component)
- Themes (dark/light)
- Custom components / Icons compatibility

### Features to be add in the future

- Responsive / locked column
- Column alignment
- Column width priority

### Design Decisions

- **Sort**: Our first approach was provide internal functions to handle sorting but we can't cover all sort possibilities so we chose to provide a `onSort()` prop (that is called when the user clicks on a sortable table header), where you can plug your own sort method and handle the data as you wish. Another factor that contributed to our decision is the way how we designed the API, where the `<Table>` doesn't know the content of the rows that are being rendered, it just prints it's `{children}`, so we don't have access to the data.

- **Pagination**: In order to keep the component api/usage simple, we decided to not handle the pagination inside the component. It might be useful to have a pagination in some situations, but we can handle it outside the table component when necessary. it's very unlikely we'll need this feature often.

- **Tag**: Initially we used the `<table>` element but we had a problem to render nested tables, it was necessary to create another `<tr>` on the same level of the previous one and that broke the render because we couldn't wrap it with any other valid html element. We could overcome it using [React Fragments](https://reactjs.org/docs/fragments.html) but it's only available on React v16.2.0. So we sticked with regular `<div>`. Another factor that made us chose `<div>` instead of `<table>` is the locked columns feature that seems to be more feasible with regular flex-box divs.

### Usage

```javascript
<Table onSort={this.handleSort} headers={[
  { key: 'name', label: 'Name', sortable: true },
  { key: 'age', label: 'Age', sortable: true },
  { key: 'amount', label: 'Amount', sortable: true }
  { key: 'city', label: 'City', sortable: true }
]}>
  <Table.Row data={{ name: 'John', age: 19, amount: 100, city: 'Seoul'}} />
  <Table.Row data={{ name: 'Bob', age: 15, amount: 200, city: 'Los Angeles'}} />
  <Table.Row data={{ name: 'Maria', age: 27, amount: 300, city: 'Paris'}} />
</Table>
```

### preview:

**sorable columns**:
table-pr

**nested components**:
nested-pr

### DIFF
--- a/packages/table/.babelrc
+++ b/packages/table/.babelrc
@@ -1,0 +1,11 @@
+{
+  "presets": [
+    "react",
+    "stage-2",
+    ["env", {
+      "targets": {
+        "browsers": ["last 2 versions", "ie >= 10"]
+      }
+    }]
+  ]
+}
--- a/packages/table/.npmignore
+++ b/packages/table/.npmignore
@@ -1,0 +1,3 @@
+__specs__
+src
+.babelrc
--- a/packages/table/.npmrc
+++ b/packages/table/.npmrc
@@ -1,0 +1,1 @@
+package-lock=false
--- a/packages/table/.storybook/addons.js
+++ b/packages/table/.storybook/addons.js
@@ -1,0 +1,5 @@
+/* eslint-disable import/no-extraneous-dependencies, import/no-unresolved, import/extensions */
+import addons from '@storybook/addons'
+import '@storybook/addon-actions/register'
+import register from '@pluralsight/ps-design-system-storybook-addon-theme/register'
+register(addons)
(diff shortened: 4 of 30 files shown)


rafaelcardoso:
I'm closing this PR due an existing `@pluralsight/ps-design-system-table` component recently implemented, so I believe this PR shouldn't be merged.

rafaelcardoso:
Cool, no problem @jaketrent 
I'm glad that you used some of the ideas from this PR, and I totally get your points, thank you for explaining them.
Let me know if you need anything from me, I'll be happy to help.

PR #209, opened 2018-04-14
fix(layout): fixing a unresolved conflicting file

rafaelcardoso:
### What You're Solving

- Fix: The layout component has a unresolved conflicting `package.json` file from the last publish commit


rafaelcardoso:
Awesome, no problem dude 😃

Issue #212, opened 2018-05-03
switch:input should be tabIndex="-1"


deltaepsilon:
## Bug

I can't easily take the checkbox input within `<Switch />` out of the tab order.

### Expected behavior

I should be able to use `<Switch tabIndex="-1" />` and have that propagate down to the checkbox... or even better, the checkbox input should never get focus.

### Environment

- OS - MacOS 10.11.6
- Browser version - Chrome v66

### The Value Add

Managing tabindex is critical for accessibility.

I was able to implement a workaround for now, but it's ugly as sin and involves a bunch of manual focus management.

### The Tradeoffs

There are no tradeoffs

jaketrent:
Released in switch @1.3.0.  See #213 for eventual approach.

PR #213, opened 2018-05-03
switch:Remove switch checkbox input from the tab order

deltaepsilon:
### What You're Solving

Solves [issue #212](https://github.com/pluralsight/design-system/issues/212)

### Steps to verify

Attempt to tab to the input. You can't. That's what tabIndex="-1" does!

### DIFF
--- a/packages/switch/src/react/index.js
+++ b/packages/switch/src/react/index.js
@@ -63,6 +63,7 @@ const Switch = (props, context) => {
         type="checkbox"
         readOnly
         checked={allProps.checked}
+        tabIndex="-1"
         {...styles.checkbox(allProps)}
       />
       <label {...styles.label(allProps)}>{allProps.children}</label>


jaketrent:
I think you've pointed out something valid: that this component needs some tabbing/accessibility help.

I'm not sure that I understand the proposed change in the PR.  What is your specific case for wanting to permanently remove the switch from being tabbable/focusable?

And do you think that applies generally to the Switch always?  Seems unlikely -- that you'd probably usually *want* your form components to be tabbable/focusable.

But what we have now could use an improvement.

Current state: onClick trigger is on outer/visible/pretty div that looks and acts like a switch.  Input checkbox is the native storage for the checked state, which seems right.  It is set as screenreader only, since sighted users want to see the pretty switch div only.

But what we seem to be disregarding is the *visual* focus/tab state of the pretty div.  If a sighted user tabs through, the input that gets the focus is not visible, thus not visually-discernible as focused.

Better state?: We put a `tabIndex=0` default on the pretty switch div so that it's visually discernible.  We allow a tabIndex override for when you don't want to tab to it.  We would remove the input from the tabIndex.  But then we'd want to make the pretty div look like a checkbox to non-sighted users.  We could do that by adding `role="checkbox"` and `aria-check=true|false` to it.

Thanks for shining a light on this.  What do you think of the above proposal?

deltaepsilon:
I like your better state :)

My suggestion that we never focus on the input is because it can never hold an outline because it's hidden.

I ended up setting the input's tabIndex to -1 and wrapping the switch with a button. I'm trying to optimize for both Voice Over and a keyboard-only sighted navigation... so I'm hitting all of the edge cases 🌮

jaketrent:
Released in switch @1.3.0

PR #221, opened 2018-06-08
actionmenu: allow custom icons

KaidenR:
### What You're Solving

- Allow passing custom icons to action menu items via an icon prop
- Needed custom social media icons for share menu

### How to Verify

- Pass a custom icon component to an action menu item and verify it shows up

### DIFF
--- a/packages/actionmenu/CHANGELOG.md
+++ b/packages/actionmenu/CHANGELOG.md
@@ -3,6 +3,18 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+<a name="2.6.2"></a>
+## [2.6.2](https://github.com/pluralsight/design-system/compare/@pluralsight/ps-design-system-actionmenu@2.6.1...@pluralsight/ps-design-system-actionmenu@2.6.2) (2018-06-06)
+
+
+### Bug Fixes
+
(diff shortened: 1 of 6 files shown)


KaidenR:
We didn't update the site. We'll do that when you give the green light on this change.

jaketrent:
Hey Kaiden, is this by chance related to the work that @leo-duran is doing, using the Card for the Paths context?

jaketrent:
Hey Kaiden, I don't have a problem with this approach.  In fact, again, @leo-duran is chatting me separately that he needs to use a custom icon for a Card.Action.  And this makes sense.  

Instead of supporting both, however, I think I'd rather just support `icon`.  I'm thinking this because:

- It'd support this custom feature
- It'd be more in line with other components' apis
- It'd keep the interface to this component as simple as possible

KaidenR:
@jaketrent Yep makes sense to me. Do you want me to pull out the `iconId` prop or do you want to do that?

KaidenR:
Oh, and no this isn't related to the stuff Leo is working on. We're using the action-menu for a new dropdown in Channels.

jaketrent:
Just serendipitous work then, I guess.  

Yes, will you please pull out (replace) iconId with an icon prop?

Unless you're in a hurry to ship something, in which case we can have both props in there for a moment, let you move forward, then drop the iconId in a major rev to come.

What would you prefer?

KaidenR:
I'll go ahead and update my PR to pull that prop out in a breaking-change commit.

PR #230, opened 2018-06-26
WIP: button,icon: add custom element conversion script

lizardruss:
## Instructions

- fill in title: format of "button,icon: add custom element conversion script"
- label: enhancement

### What You're Solving

- This is a work in progress, and should be treated as a starting point for discussion
- This uses [web-react-components](https://github.com/ChristophP/web-react-components) to convert react components into custom elements in order to consume them in non-react applications
- Separate `/web` scripts have been created to allow opting in to custom elements
- [Example angular application](https://github.com/lizardruss/angular-react-interop) showing how the custom elements can be consumed

### Design Decisions

- This uses a [fork](https://github.com/lizardruss/web-react-components) of [web-react-components](https://github.com/ChristophP/web-react-components) that allows inheriting of styles into the custom element's shadow dom. This is enabled by passing the `inheritStyles` option when calling `register()`.
- Styles could also be included in the custom element, however accomplishing that would require revising the current build process.
- Mirroring the React practice of passing components in attributes isn't easily mapped to custom elements. The custom element way would be to design  the component with `<slot name="foo"></slot>` elements in it's template, and provide the slotted content when using the custom element. I've included a factory function, `createButtonWithIcon()`, to demonstrate how one might work around this limitation by first generating a react component and converting it into a custom element.
- Any feedback on whether this approach is practical would be appreciated.

### How to Verify

- [Example angular application](https://github.com/lizardruss/angular-react-interop)

### DIFF
--- a/packages/button/package.json
+++ b/packages/button/package.json
@@ -30,7 +30,8 @@
     "@pluralsight/ps-design-system-theme": "^1.3.1",
     "@pluralsight/ps-design-system-util": "^1.2.1",
     "polished": "^1.7.0",
-    "prop-types": "^15.5.10"
+    "prop-types": "^15.5.10",
+    "web-react-components": "lizardruss/web-react-components#master"
   },
   "devDependencies": {
     "@pluralsight/ps-design-system-build": "^1.8.1",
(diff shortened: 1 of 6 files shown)


jaketrent:
Triaged: Super cool, but we're not actively working on this.  Would love to revisit.

Issue #194, opened 2018-02-20
site: tooltip flicker on hover [chrome]


jaketrent:
in-app doc for tooltip.  in chrome.  hover, move cursor, get flicker of tooltip.

jaketrent:
Fixed!

PR #245, opened 2018-08-14
fix(site): tooltip flicker on hover

JustAboutJeff:
### Instructions

- site: fix tooltip flicker on hover [chrome]
- label: bug

### What I'm Solving

- see #194 

### Design Decisions

- After a bit of debugging, the issue here appeared to be driven by mouse events firing too quickly. One solution to this problem is limiting the rate of invocation for callbacks that `setState` within the underlying component. To achieve this, I put together a small `debounce` util function that will not invoke the provided callback prior to a delay timeout.

- There are lots of other implementations of `debounce` out there readily available via `npm`, however I saw from the existing documentation here that `util` may be the place to drop something like this. In addition, there is value in limiting the number of dependencies we take on as we are shipping a library here!
 
- I also decided to throw immediately from `debounce` if the callsite provides incorrectly typed arguments. My thinking was that developers would like to know sooner rather than later that they are not going to get a debounced invocation.

- There may be some additional discussion warranted around using a `throttle` implementation here as opposed to `debounce`. Or reworking the `debounce` implementation to fire immediately, then using timeouts for subsequent invocations. These types of features would have added additional complexity so I figured I would leave that decision to follow up discussion here on github 😅 

### How to Verify

- See `Hover me` example under http://localhost:1337/components/tooltip#in-app-example while 
   running the design system locally with this branch checked out.

### DIFF
--- a/packages/site/pages/components/tooltip.js
+++ b/packages/site/pages/components/tooltip.js
@@ -4,6 +4,7 @@ import Icon from '@pluralsight/ps-design-system-icon/react'
 import React from 'react'
 import Theme from '@pluralsight/ps-design-system-theme/react'
 import Tooltip from '@pluralsight/ps-design-system-tooltip/react'
+import * as funcUtil from '@pluralsight/ps-design-system-util/func'
 
 import {
   Chrome,
@@ -108,8 +109,11 @@ class InAppExample extends React.Component {
       isHovered: false,
       isClicked: false
     }
-    this.handleMouseOver = this.handleMouseOver.bind(this)
-    this.handleMouseOut = this.handleMouseOut.bind(this)
+    this.handleMouseOver = funcUtil.debounce(
+      this.handleMouseOver.bind(this),
+      100
+    )
+    this.handleMouseOut = funcUtil.debounce(this.handleMouseOut.bind(this), 100)
     this.handleClick = this.handleClick.bind(this)
   }
   componentDidMount() {
--- a/packages/util/src/__specs__/func.spec.js
+++ b/packages/util/src/__specs__/func.spec.js
@@ -1,0 +1,42 @@
+import { debounce } from '../func'
+
+describe('#debounce', () => {
+  describe('with valid arguments', () => {
+    beforeEach(() => {
+      jest.useFakeTimers()
+    })
+
+    test('with func and delay arguments', () => {
+      const spy = jest.fn()
+      const result = debounce(spy, 100)
+      expect(result).toBeInstanceOf(Function)
+      result('foo')
+      jest.runTimersToTime(100)
+      expect(spy).toHaveBeenCalledTimes(1)
+      expect(spy).toHaveBeenCalledWith('foo')
+    })
+  })
+
+  test('empty', () => {
+    expect(debounce).toThrow(/^debounce expects a function/)
+  })
+
+  test('with func and no delay', () => {
+    expect(_ => debounce(Function.prototype)).toThrow(
+      /^debounce expects a number/
+    )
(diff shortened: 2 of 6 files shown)


packages/util/src/func.js
```diff
@@ -0,0 +1,15 @@
+export const debounce = (func, delay) => {
+  if (typeof func !== 'function') {
```
jaketrent: Since this util package is one used internally by the DS, I'd prefer to leave these in-src usage helps out.  My main reason would be to rely on us as expert users of the DS and not ship this code to consumers of things like the Tooltip component.

packages/util/src/__specs__/func.spec.js
```diff
@@ -0,0 +1,42 @@
+import { debounce } from '../func'
+
+describe('#debounce', () => {
+  describe('with valid arguments', () => {
+    beforeEach(() => {
+      jest.useFakeTimers()
+    })
+
+    test('with func and delay arguments', () => {
```
jaketrent: Great tests.  Thank you!

packages/site/pages/components/tooltip.js
```diff
@@ -4,6 +4,7 @@ import Icon from '@pluralsight/ps-design-system-icon/react'
 import React from 'react'
 import Theme from '@pluralsight/ps-design-system-theme/react'
 import Tooltip from '@pluralsight/ps-design-system-tooltip/react'
+import * as funcUtil from '@pluralsight/ps-design-system-util/func'
```
jaketrent: Do we get better tree-shaking if we just import the function instead of the whole module with `* as`?  I'm not sure, but seems like I remember that... :)

JustAboutJeff:
Thanks for the review @jaketrent will follow up :soon:

packages/util/src/func.js
JustAboutJeff: Makes sense - I can adjust

packages/site/pages/components/tooltip.js
JustAboutJeff: Great question, there are a few variables that would affect the final outcome of this:

* The first being consumer applications and their module bundler. Are they using a tool that is ES2015 module compliant and one that can perform tree-shaking (webpack, rollup, etc)?
* The second being how we build and publish our library. Do we include a `module` entry in the `package.json` of our published packages that points to a static ES2015 module compliant build?

As you probably know, libraries are often transpiled down to a dynamic module system like Common JS to support the widest possible execution target. We could look into publishing a second build that includes a static build that would take advantage of the ES2015 module system's killer feature: exporting more than a single thing! In our case here that would be a single `debounce` function, or in the future, other util functions that would be useful in isolation. This is in contrast to the Common JS single export workaround which involves exporting an object with keyed properties. This makes it impossible to determine what will not be used in the importing module, and thus prevents any optimization!

For now, I can easily rework the import/exports here to use only the function we care about. However it is a bit harder to say right now if end users in the browser will feel the effects of any optimization for the reasons I mention above!

JustAboutJeff:
thoughts on closing this in favor of [the fix over here](https://github.com/pluralsight/design-system/pull/252)?

packages/site/pages/components/tooltip.js
jaketrent: Oh man!  Great reminder.  I forgot about this.  And yes, we compile to cjs for our design system build.  And it's only using babel.  Yes, many more factors here to consider.  Fine for now.  That said, I'm going to go check out your other PR now.

Sorry for the slowness.  I've been out at a conf all week.

PR #246, opened 2018-08-16
datepicker: remove console.log statement from parseDate

thatchej:
### What You're Solving

This has been logging quite a bit in our tests and elsewhere. I'm assuming it's an oversight so submitting a quick PR -- if not let me know :)

### DIFF
--- a/packages/datepicker/src/js/index.js
+++ b/packages/datepicker/src/js/index.js
@@ -59,7 +59,6 @@ export const getMonthName = mm =>
 
 export const parseDate = value => {
   const [mm, dd, yyyy] = (value || '').split('/')
-  console.log('parseDate', mm, dd, yyyy)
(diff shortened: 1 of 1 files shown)


JustAboutJeff:
@thatchej what do you think about taking this one step further and including some protections for future developers and releases?

After looking around a bit I am seeing that `gulp` is used for the build and publish step. Because `debugger` and `console` statements can often be useful for developers while working prior to release, and because linting does not appear to be setup (perhaps because the precommit `prettier` hook is enough) we might solve this at the build stage [by including a tool like this](https://www.npmjs.com/package/gulp-strip-debug). And then within `gulpfile.js`:

```
.pipe(stripDebug())
```

Thoughts?

jaketrent:
@thatchej Thanks for noticing this and cleaning it up!

Issue #268, opened 2018-10-01
drawer: improve aria-labeling


bressain:
### Desired behavior

Move the collapsible button aria-label to the button element and allow for optional label text.

### Today's behavior

Right now there's an aria-label on the inner `div` that describes whether the collapsible button is expanded or not. When tabbing around, we land on the enclosing button, not the inner div.

The label is either Collapse or Expand which while describes the action, doesn't necessarily describe what the user is collapsing or expanding.

### The Value Add

This would provide better visibility and context for those that are reading the aria-label text.

### The Tradeoffs

Yet another optional prop.

PR #269, opened 2018-10-01
drawer: allow custom aria-label

bressain:
### What You're Solving

See #268
This moves the aria-label to the button element and allows for adding a custom label suffix.

### Design Decisions

- This was pointed out from the recent a11y audit (item 122). Allowing for a custom label allows us to give more context for what's in the drawer.
- This _could_ be seen as a breaking change but unsure if anyone is relying on that label being on the other element. We suspect that this shouldn't be a problem.
- We considered making this a callback to allow for more control but felt that this would be easier to set up.

### How to Verify

- Added a story to verify.


jaketrent:
Great catch.  Thanks for verifying.  Worked for me too.  Merging.

PR #271, opened 2018-10-03
dropdown: add ability to override selectedLabel

KaidenR:
### What You're Solving

When you're using a dropdown outside a form it's useful to be able to give more context about the field and selected value inside the dropdown itself. I've come across a few cases where I've been unable to use the Dropdown component without the ability to do this. Right now I'm trying to build this:

image

In order to prefix the selected value with `Sort:` I need to be able to override the selected label.

### Design Decisions

We should get Dmitry's approval on this design. I understand that most of the times, when you're using a dropdown inside a form, you're going to want to use the regular field label and sublabel to communicate, but when you're outside a form it's useful to be able to provide more context without a label.

### How to Verify

The dropdown storybook is updated with a new story. Make sure that when a `selectedLabel` is provided it is always show even after selecting an item from the menu.

### DIFF
--- a/packages/dropdown/src/react/__specs__/__snapshots__/storyshots.spec.js.snap
+++ b/packages/dropdown/src/react/__specs__/__snapshots__/storyshots.spec.js.snap
@@ -860,7 +860,9 @@ exports[`Storyshots labels label 1`] = `
           <span
             aria-hidden={true}
             data-css-e6fvyw=""
-          />
+          >
+            Three item
+          </span>
           <span
             data-css-5m5yd6=""
           />
@@ -934,7 +936,9 @@ exports[`Storyshots labels label and subLabel 1`] = `
           <span
             aria-hidden={true}
             data-css-e6fvyw=""
-          />
+          >
+            Three item
+          </span>
           <span
             data-css-5m5yd6=""
           />
@@ -1008,7 +1012,9 @@ exports[`Storyshots labels none 1`] = `
           <span
             aria-hidden={true}
             data-css-e6fvyw=""
-          />
+          >
+            Three item
+          </span>
           <span
             data-css-5m5yd6=""
           />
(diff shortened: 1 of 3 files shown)


dborodyansky:
Thank you @KaidenR! You mentioned multiple cases this may address. Would you please point us to another one for reference? We try to consider DS for wide, or multiple team, applications (https://whimsical.co/GncwqZbmGjHBKT9srDYBSj), so there are may be a trade off here in adding a prop vs. alternate solutions like A) an external `Sort` label on the left of the dropdown, or B) appending `Sort:` to each menu item. Those alternates are certainly not as elegant, but worth discussing around how the solution may pertain to other use cases.

KaidenR:
@dborodyansky Another case is the "Add content" dropdown on a channel details page:

image

This is another situation where the user is taking an action rather than selecting a value from a list to populate a form. 

One more case is the "Role" dropdown in a channel's members modal:

image

We need to be able to be more succinct with the label on the button ("Can edit") but more descriptive inside the menu ("Can edit (Contributor)")

dborodyansky:
Thank you, @KaidenR. This does help explain wider application. Much appreciated. In reviewing, I think this enhancement does seem suitable. Jake and I will touch base and follow up.

jaketrent:
Hey @KaidenR , sorry to be equivocating a bit and not making this higher priority.  We have had other items that have come first.

To at least give you an update...

Dmitry and I have followed up and talked about this as a feature.  We're still not super clear on what the best course of action is.  It's easy to add things.  It's hard to take them away.  Since there are ways to accomplish your specific use case (get a label inside the DS component by using it in the options label as well), we feel like there's a path forward, perhaps not ideal, for the time being.  

The other variants that you have shown of a dropdown in our app have been eye opening.  Thank you for including that survey of those usages.  They raise another question that makes the way forward not any clearer: that is, each of these other dropdowns use very different looks/UIs, and should we support this?  The one is a button with a dropdown drawer beneath.  The other is a button(?) with a very different dropdown drawer beneath.  

Thus far, the DS has provided the looks for the UI of Pluralsight.  The focus has been on visual consistency.  Obviously, there are other things that have come with that: one is a dropdown component with a certain amount of visual state and interactivity patterns.  That obviously has value.  And we could publish that as a general thing, allowing a developer to customize the look of the thing while it still retaining the properties of a dropdown.  Think the patterns from Kent Dodd's talk from React Rally this year https://www.youtube.com/watch?v=AiJ8tRRH0f8  But this jump is a big one for the Design System from a purpose and a "where are we going to play" standpoint.  

So, the team is going to look at this again next week when we are all together again, but unfortunately for your present need, this PR is still in a holding pattern.  I hope you'll understand.  I also welcome your input on the further questions raised and mentioned above.  Thanks, Kaiden your constant support.

jaketrent:
Triage: We looked at this as a team.  We were really torn with this one.  We want to support you in your dev work on this page.  We also looked at what the DS enables today, and we at least think there's a workaround here without adding to the DS.  You could use a Button with an ActionMenu in combo (as shown in the other example above).  The button style does not quite match the "flat" appearance of Button today, so you could override the style.  But Dmitry reports that this is an emerging button style amongst the designs of late, so it's possible that it'll make its way into Button in the future.  Closing for now.  But open, as always, to more discussion about what you're seeing and what would be helpful.

PR #279, opened 2018-10-11
dropdown: increase dropdown overlay z-index

### LINKED ISSUE #278: dropdown: clicking outside the open menu doesn't always close the dropdown
### Expected behavior
I should be able to click *anywhere* outside of the open dropdown to close it.

### Actual behavior
There are certain sections of the page (anywhere in prism, a channel card list) that do not close the dropdown. The problem is that the page overlay that intercepts clicks does not have a high enough z-index to cover all parts of the page. The screenshot below shows an open dropdown in the top-right and the page overlay in blue. You can see that it isn't covering the entire page:

image

### Steps to repro

1.  Add a dropdown to a page
2. Add another element later in the DOM with a `position:relative`
3. Open the dropdown and click on the other element - the dropdown does not close

KaidenR:
### What You're Solving

- Fixes #278 
- Fixes issue where the dropdown overlay was not covering the entire page so
the dropdown would not always close when clicking outside of the open menu.

### Design Decisions

- Chose a z-index of `970` for the overlay because we want it to cover prism (the sidebar has a z-index of `960`)
- Chose a z-index of `980` for the menu so it appears above the overlay.

### How to Verify

1. Add a dropdown to a page
2. Add an element later in the DOM with `position: relative`
3. Open the dropdown
4. Click the other element and/or Prism - the dropdown should close

### DIFF
--- a/packages/dropdown/src/css/index.js
+++ b/packages/dropdown/src/css/index.js
@@ -213,13 +213,14 @@ export default {
   // __menu
   '.psds-dropdown__menu': {
     position: 'absolute',
-    zIndex: '1',
+    zIndex: 980,
     marginTop: core.layout.spacingXXSmall
   },
 
   // __page-overlay
   '.psds-dropdown__page-overlay': {
     position: 'fixed',
+    zIndex: 970,
     top: 0,
     right: 0,
     bottom: 0,


jaketrent:
Hey Kaiden, I want your opinion on the other layer z-indexes that prism sits on (some well above 980).  Some might not matter.  

https://github.com/ps-dev/prism/blob/master/client/src/common/styles/layers.css

But z-indexes are just mayhem anyway.  

Feels like we need some more clear guidelines on what plays in what z space across our product.  Might be nice to be more comprehensively sure about these things.

That might be a bigger effort.  

It's probably the case that we just merge 980 for this particular issue and time and go from there if we have another issue.  But, who knows, like I said, mayhem.  

Anywho, a quick comment on these other prism layers based on that knowledge would be helpful for me.

KaidenR:
I was looking for that list of Z Indexes in Design system but didn't think to look in Prism. After looking at it, I think I'd still be comfortable using the 970 & 980 in my PR. Those higher z-indexes shouldn't ever interfere with dropdowns because it would take some other click in Prism to get them to show up and the click should be intercepted by the dropdown overlay.

KaidenR:
Oh and yeah, it would be great to have a single file somewhere that allocates specific z-indexes across the app so we can make more informed decisions when picking them. But, depending on where your app's stacking contexts are, those values may or may not help.

PR #281, opened 2018-10-17
Errors: Low contrast in light theme (below WCAG contrast ratio)

danethurber:
### Desired behavior

The contrast in the body text of the error page for the light theme meets the WCAG contrast ratio.

### Today's behavior

The white body text of the error message is low contrast compared to the light grey background.

### The Value Add

Users will find it easier to read.  We will meet the accessibility guidelines.


getschomp:
🎉

Issue #162, opened 2018-01-10
card: add required 'title' attribute to Card.Action to display tooltip for icon


dborodyansky:
### Desired behavior

Encourage developers to add `title` / tooltip for non labeled buttons, so that a default browser tooltip appears on hover.

### Today's behavior

No default browser tooltip appears when hovering on a card action

### The Value Add

Accessibility, best practices.

### The Tradeoffs

More required props

jaketrent:
Triage: team convo.

PR #292, opened 2018-10-18
card: require title prop on Card.Action

danethurber:
fixes: #162

### DIFF
--- a/packages/card/src/react/index.js
+++ b/packages/card/src/react/index.js
@@ -83,12 +83,6 @@ const ActionBar = glamorous.div(
     actionBarVisible ? css['.psds-card__action-bar--actionBarVisible'] : null
 )
 
-const ActionBarAction = props => {
-  const filteredProps = {}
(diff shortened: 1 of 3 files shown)


packages/card/src/react/index.js
```diff
@@ -105,8 +99,16 @@ const ActionButton = glamorous.button(
...
+const ActionBarAction = props => {
+  const filteredProps = { title: props.title }
+  if (props.onClick) filteredProps.onClick = props.onClick
+  return <ActionButton {...filteredProps}>{props.icon}</ActionButton>
+}
+
 ActionBarAction.propTypes = {
-  icon: PropTypes.element.isRequired
+  icon: PropTypes.element.isRequired,
+  title: PropTypes.string.isRequired
```
jaketrent: k, we're going with required, people will get this minor upgrade automatically in many cases and start to get this error.  That's the plan, right?  I'm fine with that.

Might be worth putting that fact in the commit message so it shows up in the changelog.  Well, I guess the title will do that.

Issue #290, opened 2018-10-17
button: shows black background on focus


KaidenR:
### Expected behavior

The background should remain the same when focused.

### Actual behavior

When a flat button on a colored background is focused the background of the button changes to black because the focus ring adds the background.

### Steps to repro

1. Add a flat button to a container with a background
2. Focus the button - the button's background will go black

![image](https://user-images.githubusercontent.com/679346/47105010-0cb7b880-d200-11e8-8167-e9efc7509afc.png)
![image](https://user-images.githubusercontent.com/679346/47105021-13dec680-d200-11e8-8a81-5172cfb90a23.png)

Issue #289, opened 2018-10-17
form-related: focus ring not showing up


KaidenR:
### Expected behavior

Focus ring should always show when button is focused.

### Actual behavior

When buttons are inside a container with a background, the focus ring is rendered behind the background and so does not show up. This may be an issue with all elements that have focus rings and not just button.

### Steps to repro

1. Add a button to a container that has a background.
2. Focus the button - the focus ring does not show up

jaketrent:
Kaiden noted yesterday when talking to me about this that it's likely that ALL the form-related components are probably going to need an adjustment here.  TextInput has been fixed already ec73ec67666999412bb9cc900a8d5bda025c3ebd

jaketrent:
This even more ready to work now that the Halo component PR has been merged.   But I'm going to wait a day to pick it up until the other jest/hoisting/touch everything PRs merge so-as to avoid being hammered at merge time.  There's other stuff to do.

PR #297, opened 2018-10-19
button: update focus styles to allow transparent backgrounds

danethurber:
fixes: #290

### Design Decisions
The button had two pseudo elements that were used to display our focus state. One of the pseudo elements was the entirety of the button background and defined a theme specific background color. This caused the `flat` appearance to have a background color on focus that was unwanted (#290).

I removed the `:after` pseudo element and updated the styles for the `:before` pseudo element to use a border only. Originally, through the use of two pseudo elements, we were able to achieve a nice rounded inner border on the focus outline that matched the rounded corner of the button. This change results in a less than ideal inner border rounding because i don't have a second pseudo element to overlay over the first to give the illusion of a rounded inner border.

screen shot 2018-10-19 at 1 51 14 pm

I also introduced a new storybook decorator to center the story in the viewport. I needed this to see the fill focus state of the button and I think it looks better. I would like to separate that into a separate package and use in other components - i'll need some help to accomplish that :)

fixes: #290

selfie-0
[_GitHub Selfies_](https://github.com/thieman/github-selfies/)

### DIFF
--- a/packages/button/src/css/index.js
+++ b/packages/button/src/css/index.js
@@ -46,37 +46,13 @@ export default {
   '.psds-button:focus:before': {
     content: ' ',
     position: 'absolute',
-    top: '-1px',
-    left: '-1px',
-    right: '-1px',
-    bottom: '-1px',
-    background: core.colors.black,
-    zIndex: '-1',
-    borderRadius: '2px'
-  },
-  [`.psds-button.psds-theme--${themeNames.light}:focus:before`]: {
-    top: '-2px',
-    left: '-2px',
-    right: '-2px',
-    bottom: '-2px',
-    background: core.colors.bone
-  },
-  '.psds-button:focus:after': {
-    content: ' ',
-    position: 'absolute',
     top: '-4px',
     left: '-4px',
     right: '-4px',
     bottom: '-4px',
-    zIndex: '-2',
-    borderRadius: '4px',
-    background: core.colors.blue
-  },
-  [`.psds-button.psds-theme--${themeNames.light}:focus:after`]: {
-    top: '-5px',
-    left: '-5px',
-    right: '-5px',
-    bottom: '-5px'
+    zIndex: '-1',
+    border: `3px solid ${core.colors.blue}`,
+    borderRadius: '4px'
   },
 
   // --size
--- a/packages/button/src/css/index.js
+++ b/packages/button/src/css/index.js
@@ -50,7 +50,6 @@ export default {
     left: '-4px',
     right: '-4px',
     bottom: '-4px',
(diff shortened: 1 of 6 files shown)


packages/button/src/react/__specs__/storyshots.spec.js
```diff
@@ -1,3 +1,11 @@
+import React from 'react'
 import initStoryshots from '@storybook/addon-storyshots'
 
+jest.mock('../../../.storybook/center-decorator.js', () => {
```
danethurber: i'm mocking out the decorator to make the storyshot snapshots less noisy

packages/button/src/css/index.js
```diff
@@ -44,39 +44,15 @@ export default {
...
-    borderRadius: '4px',
-    background: core.colors.blue
-  },
-  [`.psds-button.psds-theme--${themeNames.light}:focus:after`]: {
-    top: '-5px',
-    left: '-5px',
-    right: '-5px',
-    bottom: '-5px'
+    zIndex: '-1',
+    border: `3px solid ${core.colors.blue}`,
```
jaketrent: I love the simplification of this.  I am concerned that we've diverged from the original requirements of the focus indicator.  I'm also concerned that I don't remember them 100%.  It was quite the process to get them exactly right.  Perhaps we need to review what exactly they are.  This would be good to do before I tackle #289, which is going to address focus ring visibility for the other form-related components.  I propose a 15 minute meeting to review today.  @dborodyansky would you be able to walk us through the viz specs of the focus indicator today?

packages/button/src/css/index.js
```diff
@@ -44,39 +44,15 @@ export default {
...
-    zIndex: '-2',
-    borderRadius: '4px',
-    background: core.colors.blue
-  },
-  [`.psds-button.psds-theme--${themeNames.light}:focus:after`]: {
-    top: '-5px',
-    left: '-5px',
-    right: '-5px',
-    bottom: '-5px'
+    zIndex: '-1',
```
jaketrent: Is z-index: -1 required now?  I used it in the past to hide the bg of the pseudo element *behind* the button.  But now it's just a border.  If not z-index: -1, maybe need pointer-events: none?

packages/button/src/css/index.js
danethurber: totally agree. a quick design meeting would be helpful.

i could tell there was a clear reasoning for the complexity :)

packages/button/src/css/index.js
danethurber: makes sense. i'll re-address after we go over the requirements

PR #298, opened 2018-10-22
lint and format on precommit hook

danethurber:
### What You're Solving

Adds a precommit hook to lint and format `.js` files. Also adds a hook to format `package.json` files. This is to ensure consistent style and automate the process. This also will make vscode/prettier play nicer with the project and give helpful hints in editors with eslint integrations.

As usual the precommit can be bypassed with the `-n` flag but my hope that in the future this is unnecessary - more of just an escape hatch while we're in transition.

### Design Decisions

- I chose to adopt the standardjs style - this is so we don't have to spend time defining our own rules
- I just integrated the tooling but did not fix all the linting errors. my reasoning is that we can fix them as we work on or refactor code. This allows a more incremental(but longer) transition.
- I added a hook to format package.json files and made the appropriate changes. This is more a style design I make but it is very low impact and i've liked having it automated in the past.
- i added a `.editorconfig` file for editors that support this.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,0 +1,7 @@
+module.exports = {
+  extends: ['standard', 'standard-react', 'plugin:jest/recommended'],
+  plugins: ['jest'],
+  env: {
+    'jest/globals': true
+  }
+}
--- a/commitlint.config.js
+++ b/commitlint.config.js
@@ -1,7 +1,7 @@
 const Repository = require('lerna/lib/Repository')
 const PackageUtilities = require('lerna/lib/PackageUtilities')
 
-function getPackageAbbreviations() {
+function getPackageAbbreviations () {
   const cwd = process.cwd()
   const repo = new Repository(cwd)
   return PackageUtilities.getPackages({
--- a/package.json
+++ b/package.json
@@ -1,13 +1,15 @@
 {
   "name": "design-system",
   "version": "1.0.0",
-  "private": true,
   "description": "Pluralsight design system",
-  "main": "packages/site/index.js",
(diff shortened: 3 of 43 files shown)


.eslintrc.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
```
danethurber: i'm keeping a simple as possible and just adopting other maintained rules.

commitlint.config.js
```diff
@@ -1,7 +1,7 @@
 const Repository = require('lerna/lib/Repository')
 const PackageUtilities = require('lerna/lib/PackageUtilities')
 
-function getPackageAbbreviations() {
+function getPackageAbbreviations () {
```
danethurber: I did format the few .js files and the root level - just because

package.json
```diff
@@ -1,39 +1,65 @@
 {
```
danethurber: example of the `format-package` script that organizes `package.json` files. there is some noise in this commit because of the changes but I thought it was warranted.

package.json
```diff
@@ -1,39 +1,65 @@
...
     "start": "npm run site",
     "test": "jest",
-    "test:watch": "npm run test -- --watchAll",
-    "test:updateSnapshot": "npm run test -- --updateSnapshot"
+    "test:updateSnapshot": "npm run test -- --updateSnapshot",
+    "test:watch": "npm run test -- --watchAll"
+  },
+  "lint-staged": {
+    "**/*.js": [
+      "eslint --fix",
```
danethurber: eslint with attempt to fix staged files and recommit them

PR #299, opened 2018-10-22
storybook: create new package for center addon

danethurber:
this will allow the addon to be used by other packages in the design system

### DIFF
--- a/commitlint.config.js
+++ b/commitlint.config.js
@@ -2,16 +2,18 @@ const Repository = require('lerna/lib/Repository')
 const PackageUtilities = require('lerna/lib/PackageUtilities')
 
 function getPackageAbbreviations() {
+  const prefix = `ps-design-system-`
+
   const cwd = process.cwd()
   const repo = new Repository(cwd)
+
   return PackageUtilities.getPackages({
     packageConfigs: repo.packageConfigs,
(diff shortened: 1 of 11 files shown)


commitlint.config.js
```diff
@@ -2,16 +2,18 @@ const Repository = require('lerna/lib/Repository')
...
+
   return PackageUtilities.getPackages({
     packageConfigs: repo.packageConfigs,
     rootPath: cwd
   })
     .map(pkg => pkg.name)
     .map(name => (name.charAt(0) === '@' ? name.split('/')[1] : name))
-    .map(name => (/-/.test(name) ? name.split('-').pop() : name))
-    .concat(['storybook-addon-theme'])
+    .map(name => (name.includes(prefix) ? name.replace(prefix, '') : name))
```
danethurber: i updated this to just check and strip the design system prefix.

packages/button/src/react/__specs__/storyshots.spec.js
```diff
@@ -1,11 +1,6 @@
+jest.mock('@pluralsight/ps-design-system-storybook-addon-center')
```
danethurber: this mock now lives next to the implementation

PR #300, opened 2018-10-23
<Halo />

danethurber:
- [x] setup new `<Halo />` component scaffolding
- [x] `<Halo />` implementation
  - [x] default appearance
  - [x] error appearance
  - [x] show on focus
  - [x] dark/light theme support
  - [x] config prop for rounding edges
  - [x] fix `:focus-within` in ie/edge
- [x] use in other components
  - [x] `<Switch />`
  - [x] `<TextInput />`

questions:
  - `<Switch />` allow focus if disabled?

--- 

reference design: 
image


commitlint.config.js
```diff
@@ -13,6 +13,7 @@ function getPackageAbbreviations() {
     .map(pkg => pkg.name)
     .map(name => (name.charAt(0) === '@' ? name.split('/')[1] : name))
     .map(name => (name.includes(prefix) ? name.replace(prefix, '') : name))
+    .concat(['project'])
```
danethurber: i added a new `project` scope for commitizen so we can do things outside the packages

packages/halo/src/css/index.js
```diff
@@ -0,0 +1,90 @@
+import core from '@pluralsight/ps-design-system-core'
+
+export const BASE_CLASSNAME = '.psds-halo'
+
+// TODO: make sure this is building/exported correctly
```
danethurber: i need more info about how the css is built so i can verify

packages/halo/src/react/index.js
```diff
@@ -0,0 +1,73 @@
...
+import polyfillFocusWithin from 'focus-within'
+import PropTypes from 'prop-types'
+import React from 'react'
+
+import css, { BASE_CLASSNAME } from '../css'
+import * as vars from '../vars'
+
+import withDefaultTheme from './with-default-theme'
+
+polyfillFocusWithin(document)
```
danethurber: there are other ways to do this. i went with this package/approach because it was the most widely used

packages/halo/src/react/index.js
```diff
@@ -0,0 +1,73 @@
...
+
+const styles = {
+  halo: props => {
+    const theme = `${BASE_CLASSNAME}--theme-${props.themeName}`
+    const appearance = `${BASE_CLASSNAME}--appearance-${props.appearance}`
+    const gapSize = `${BASE_CLASSNAME}--gap-size-${props.gapSize}`
+    const shape = `${BASE_CLASSNAME}--shape-${props.shape}`
+
+    return glamor.css(
```
danethurber: there seems to be some hurdles to getting the write css for the components. also i think this is generating a classname for every combination of possible variations. i assume there is a reason i was just curious if there is a simplier way

packages/halo/src/react/with-default-theme.js
```diff
@@ -0,0 +1,28 @@
...
+
+const getDisplayName = Component => {
+  if (typeof Component === 'string') return Component
+
+  if (!Component) return undefined
+
+  return Component.displayName || Component.name || 'Component'
+}
+
+export default function withDefaultTheme (BaseComponent) {
```
danethurber: I don't love this name - to me it might mean the component is explicitly using the `default` theme. what could be a better name, `withThemeDefault`? `withFallbackTheme`?

packages/halo/stories/focused.js
```diff
@@ -0,0 +1,62 @@
...
+    background: core.colors.gray02,
+    borderColor: core.colors.gray02,
+    color: core.colors.white
+  }),
+  pill: glamor.css({
+    borderRadius: 1000
+  })
+}
+
+class Focused extends React.Component {
```
danethurber: this is just used in the story to have an element that autofocuses

packages/storybook-addon-center/__mocks__/index.js
```diff
@@ -1,10 +1,5 @@
 var React = require('react')
 
-var CenterDecorator = props => {
-  return React.createElement('div', {
-    ...props,
-    'data-storybook-center-decorator': true
-  })
-}
+const MockCenter = props => React.createElement(React.Fragment, props)
```
danethurber: this makes even smaller snapshots

packages/switch/src/react/index.js
```diff
@@ -71,69 +33,84 @@ const styles = {
...
 
     return (
       <button
         {...switchProps}
-        tabIndex={allProps.tabIndex || '0'}
-        role="checkbox"
         aria-checked={allProps.checked}
-        onFocus={this.handleFocus}
         onBlur={this.handleBlur}
+        onFocus={this.handleFocus}
```
jaketrent: love me some alpha ordering!

packages/switch/src/react/index.js
```diff
@@ -71,69 +33,84 @@ const styles = {
...
+          </div>
+
+          <input
+            tabIndex='-1'
+            type='checkbox'
+            readOnly
+            checked={allProps.checked}
+            {...styles.checkbox(allProps)}
+          />
+        </Halo>
```
jaketrent: nbd, but putting Halo around the div only might be a subtle indicator as to what is visible vs hidden. 🤷‍♂️

packages/switch/src/react/index.js
```diff
@@ -71,69 +33,84 @@ const styles = {
...
+        >
+          <div {...styles.track(allProps)}>
+            <div {...styles.thumb(allProps)} />
+          </div>
+        </Halo>
+
         <input
-          tabIndex="-1"
-          type="checkbox"
+          tabIndex='-1'
```
jaketrent: funny, I either missed this and didn't say anything or it wasn't in the previous commits.  Are there new eslint rules on the scene midway through this PR?  I'm still trying to figure out what new tooling has its ladles in the kitchen.

PR #302, opened 2018-10-26
[WIP] feat(project): add script to scaffold new components

danethurber:
### What You're Solving

I found there were many steps to setup a new component. I wanted an automated way to save myself time and errors - as well and to make it easier for outside contributors to get started.

### How to Verify

- `npm run component:create`

### Outstanding tasks

- [ ] glamor integration
- [ ] setup jest/enzyme
- [ ] update contributor docs

### DIFF
--- a/scripts/create-component/index.js
+++ b/scripts/create-component/index.js
@@ -1,0 +1,92 @@
+#!/usr/bin/env node
+
+const fs = require('fs')
+const path = require('path')
+
+const inquirer = require('inquirer')
+const copyTemplateDir = require('copy-template-dir')
+
+const { getPackages, toPackageName } = require('../utils/packages')
+const { smush, pascalize } = require('../utils/string')
+
+const MINIMUM_REACT_VERSION = '^16.2.0'
+
+const packagesDir = path.resolve(__dirname, '..', '..', 'packages')
+const templateDir = path.resolve(__dirname, 'package-template')
+
+const copyPackageFromTemplateDir = (source, dest, vars) => {
+  return new Promise((resolve, reject) => {
+    copyTemplateDir(source, dest, vars, (err, createdFiles) => {
+      if (err) reject(err)
+      else resolve(createdFiles)
+    })
+  })
+}
+
+const allPackages = getPackages()
+
+const suggestedDependencies = [
+  'button',
+  'core',
+  'icon',
+  'normalize',
+  'theme',
+  'util'
+].map(toPackageName)
+
+const questions = [
+  {
+    type: 'input',
+    name: 'name',
+    message: 'Component Name',
+    validate: name => (name.length > 0 ? true : 'invalid name')
+  },
+  {
+    type: 'checkbox',
+    name: 'dependencies',
+    message: 'Dependencies',
+    choices: suggestedDependencies.map(name => ({ name })),
+    filter: keys => {
+      const depMap = keys.reduce((acc, key) => {
+        const version = `^${allPackages[key]}`
+        return { ...acc, [key]: version }
+      }, {})
+
+      return JSON.stringify(depMap)
+    }
+  }
+]
+
+async function main () {
+  const answers = await inquirer.prompt(questions)
+
+  const componentName = pascalize(answers.name)
+  const folderName = smush(answers.name)
+  const packageName = toPackageName(answers.name)
+
+  const folderPath = path.resolve(packagesDir, folderName)
+
+  const vars = {
+    ...answers,
+    componentName,
+    folderName,
+    folderPath,
+    packageName,
+    reactVersion: MINIMUM_REACT_VERSION
+  }
+
+  if (fs.existsSync(folderPath)) {
+    throw new Error(`folder ${folderName} already exists`)
+  }
+
+  await copyPackageFromTemplateDir(templateDir, folderPath, vars)
+
+  return vars
+}
+
+main()
+  .then(vars => {
+    console.info(`\n${vars.componentName} created at: ${vars.folderPath}\n`)
+    console.info('Next run: npm run bootstrap\n')
+  })
+  .catch(console.error)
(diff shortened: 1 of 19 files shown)


scripts/create-component/index.js
```diff
@@ -0,0 +1,95 @@
...
+  const packagesMap = packages.reduce((acc, pkg) => ({
+    ...acc,
+    [pkg.name]: pkg.version
+  }))
+
+  const suggestedDependencies = [
+    'button',
+    'core',
+    'icon',
+    'normalize',
```
jaketrent: This is a peerDep.  probably don't want it listed here.

scripts/create-component/index.js
```diff
@@ -0,0 +1,95 @@
+#!/usr/bin/env node
```
jaketrent: are there cross-OS problems with this.  Not familiar enough to know.  would the script reading `node thisscript.js` be more sure-fire compatible?

scripts/create-component/index.js
```diff
@@ -0,0 +1,95 @@
...
+
+  const folderPath = path.resolve(packagesDir, folderName)
+
+  const vars = {
+    ...answers,
+    componentName,
+    folderName,
+    folderPath,
+    packageName,
+    reactVersion: MINIMUM_REACT_VERSION
```
jaketrent: I'm not sure where this is actually used, but as far as DS support goes, this min version is too high.  Min version is 15.?.? somewhere.

scripts/create-component/package-template/.storybook/config.js
```diff
@@ -0,0 +1,16 @@
+import addons from '@storybook/addons'
+import { addDecorator, configure } from '@storybook/react'
+
+import centerDecorator from '@pluralsight/ps-design-system-storybook-addon-center'
+import themeDecorator from '@pluralsight/ps-design-system-storybook-addon-theme'
+
+addDecorator(centerDecorator)
+addDecorator(themeDecorator(addons))
+
+const req = require.context('../stories', true, /\.story\.js$/)
```
jaketrent: Is the *.story.js a new convention?  Looks like it's in the latest halo-touching stuff.  I think I'm ok with it.  Just didn't notice it before.

jaketrent: Perhaps related to this question and resolvable later... perhaps we should move *.story.js files out of '/stories' and into '/.storybook'.  Not sure if we need two storybook dirs in every package.

scripts/create-component/index.js
```diff
@@ -0,0 +1,95 @@
...
+    [pkg.name]: pkg.version
+  }))
+
+  const suggestedDependencies = [
+    'button',
+    'core',
+    'icon',
+    'normalize',
+    'theme',
+    'util'
```
jaketrent: So few things use this, and I think I'd rather component devs not try to get too fancy about splitting out reusable things.  I think we as product owners are probably better equipped with a system vision to do that once we have a working component added to the library.

scripts/create-component/index.js
danethurber: i don't quite know the difference. you'll have to teach me

scripts/create-component/index.js
danethurber: i will remove - this list is mostly a shot in the dark. let me know if there is anything that would be valuable here

scripts/create-component/package-template/.storybook/config.js
danethurber: sounds good to me. I can get it in place with this pr for new packages and we can refactor old packages down the road

scripts/create-component/index.js
danethurber: this isn't actually used anymore. will remove

scripts/create-component/package-template/.storybook/config.js
danethurber: I hit a snag. storybook doesn't like stories in the .storybook folder. I decided to instead move the stories next to the component in src. thoughts?

scripts/create-component/index.js
jaketrent: just that the npm-script for create:component could have the value: "node scripts/create-component/index.js"

but again, not 100% sure that this current version is not windows compat.

scripts/create-component/package-template/.storybook/config.js
jaketrent: hmmm, seems a little odd.  And it's not really src.  We make the exception for specs.  We could do the same for __stories__, I suppose.

what's your fav location for these stories?

scripts/create-component/package-template/.storybook/config.js
```diff
@@ -7,10 +7,10 @@ import themeDecorator from '@pluralsight/ps-design-system-storybook-addon-theme'
...
 
-const req = require.context('../stories', true, /\.story\.js$/)
+const req = require.context('../src/react', true, /\.story\.js$/)
 
 function loadStories() {
   req.keys().forEach(filename => req(filename))
 }
 
-configure(loadStories, module)
+configure(loadStories, module);
```
jaketrent: semi: false

scripts/create-component/package-template/.storybook/config.js
```diff
@@ -7,10 +7,10 @@ import themeDecorator from '@pluralsight/ps-design-system-storybook-addon-theme'
 addDecorator(centerDecorator)
 addDecorator(themeDecorator(addons))
 
-const req = require.context('../stories', true, /\.story\.js$/)
+const req = require.context('../src/react', true, /\.story\.js$/)
```
jaketrent: does this allow __stories__ subdirs?  If we can't get to that, I think I prefer keeping packages/X/stories dirs.

scripts/create-component/index.js
danethurber: oh gotcha - yeah i can change that

scripts/create-component/package-template/.storybook/config.js
danethurber: 👍

scripts/create-component/package-template/.storybook/config.js
danethurber: yeah it will find any .story file in src/react

PR #309, opened 2018-10-31
Fix/open overlay on focus

jaketrent:
### What You're Solving

Make keyboard nav better on cards.  What you focus on in overlay will make the overlay visible (analogous to the overlay mouse over currently)

### Design Decisions

- Use :focus-within because we don't control the contents
- But :focus-within not supported in IE/Edge, so js polyfill, which feels sad

### How to Verify

- New combo story with hidden overlay by default.  Click in storybook space and tab around.  Stuff should appear!


packages/card/src/css/index.js
```diff
@@ -104,6 +104,9 @@ export default {
     pointerEvents: 'none',
     opacity: 0
   },
+  '.psds-card__full-overlay:focus-within': {
```
danethurber: you also need to define the `[focus-within]` attr selector. that's what is added by the polyfill in unsupporting browsers

packages/card/src/react/index.js
```diff
@@ -61,7 +61,10 @@ const renderImage = props => (props.image ? props.image : null)
 
 const FullOverlay = glamorous.div(
   css['.psds-card__full-overlay'],
-  { ':focus-within': css['.psds-card__full-overlay:focus-within'] },
+  {
+    ':focus-within': css['.psds-card__full-overlay:focus-within'],
+    '[focus-within]': css['.psds-card__full-overlay:focus-within']
```
jaketrent: I suppose the question will be: Why not add `[focus-within]` on the css side as well?  As I see it, [focus-within] is more a feature of the js implementation (that's where the polyfill is, after all).  If someone was to use the stylesheet, they'd need their own markup/code implementation.

jaketrent:
Hey @danethurber , would you mind giving this another look?

packages/card/src/react/index.js
danethurber: that makes sense to me

PR #310, opened 2018-10-31
lerna hoist

danethurber:
### What You're Solving

- get packages in similar dependency ranges so we can use --hoist
- updated bootstrap command to use --hoist
- added a new commitlint plugin to handle verifying the scope


commitlint.config.js
```diff
@@ -1,24 +1,6 @@
...
-    rootPath: cwd
-  })
-    .map(pkg => pkg.name)
-    .map(name => (name.charAt(0) === '@' ? name.split('/')[1] : name))
-    .map(name => (name.includes(prefix) ? name.replace(prefix, '') : name))
-}
-
 module.exports = {
-  extends: ['@commitlint/config-angular'],
+  extends: ['@commitlint/config-angular', '@commitlint/config-lerna-scopes'],
```
jaketrent: Fancy!

lerna.json
```diff
@@ -1,5 +1,4 @@
 {
-  "lerna": "2.0.0",
```
jaketrent: not required now, apparently?

package.json
```diff
@@ -61,7 +61,7 @@
     "pattern library"
   ],
   "dependencies": {
-    "glamor": "^2.20.39",
+    "glamor": "^2.20.0",
```
jaketrent: why the downgrade?

packages/circularprogress/package.json
```diff
@@ -28,7 +28,7 @@
   "dependencies": {
     "@pluralsight/ps-design-system-core": "^4.3.4",
     "@pluralsight/ps-design-system-theme": "^1.3.3",
-    "polished": "^1.7.0",
+    "polished": "^1.9.2",
```
jaketrent: just an errant thought:  this is pretty large for what we're using it for (I believe, just transparentize, ever).  Would be cool to maybe hork that function and drop this package someday (not this PR).

lerna.json
danethurber: yeah it's not in the docs - i forgot to change it with the upgrade an now complaints

package.json
danethurber: just to keep the versions in the same range of everything else. i figured it would resolve to the same major version but lerna likes these version in the package.json files to be exact

packages/circularprogress/package.json
danethurber: i don't know much about it - i'm curious to see what else it does

PR #312, opened 2018-10-31
Fix/document in server

jaketrent:
verified document problem and fix in local build of site package.

### DIFF
--- a/commitlint.config.js
+++ b/commitlint.config.js
@@ -1,6 +1,18 @@
+const scopes = require('@commitlint/config-lerna-scopes')
+
+const prefix = 'ps-design-system-'
+
 module.exports = {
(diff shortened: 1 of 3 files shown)


commitlint.config.js
```diff
@@ -1,6 +1,18 @@
...
+  extends: ['@commitlint/config-angular'],
   rules: {
-    lang: [0]
+    lang: [0],
+    'scope-enum': ctx =>
+      scopes.rules['scope-enum'](ctx).then(([level, applicable, packages]) => [
+        level,
+        applicable,
+        packages.map(
+          pkg => (pkg.includes(prefix) ? pkg.replace(prefix, '') : pkg)
```
danethurber: 👍

PR #313, opened 2018-11-01
fix: upgrade storybook and fix issues with require.context

danethurber:
- revert storybook configs that used `require.context` to the older pattern
- upgrade to storybook@4

storybook uses a webpack specific `require.context` method(and storyshots polyfills it - poorly) to allow more advanced import patterns. this was mostly working before we hoisted the packages through lerna - but it is now broken. I was hoping that by upgrading storybook and using their suggested [babel config](https://storybook.js.org/basics/writing-stories/#loading-stories-dynamically) we could continue to use it - that did not work...maybe an upgrade to babel@7 would help but i didn't want to go that far. instead i removed the need to use `require.context` and upgraded storybook to get everything working again.


.gitignore
```diff
@@ -41,3 +41,5 @@ jspm_packages
 .node_repl_history
 
 .env*
+
+.cache
```
danethurber: storybook creates a `.cache` folder :(

packages/halo/.storybook/config.js
```diff
@@ -7,10 +7,8 @@ import themeDecorator from '@pluralsight/ps-design-system-storybook-addon-theme'
 addDecorator(centerDecorator)
 addDecorator(themeDecorator(addons))
 
-const req = require.context('../stories', true, /\.story\.js$/)
-
 function loadStories() {
-  req.keys().forEach(filename => req(filename))
+  require('../stories')
```
danethurber: i learned this only can import a single file...not a dir of files. fyi

jaketrent: probably ok in practice, but maybe want to rename the function .... `loadStoriesFile`?  Maybe?

PR #314, opened 2018-11-01
fix(switch): disable focus and toggle when disabled

### LINKED ISSUE #307: switch: shows focus visual state even when disabled
### Expected behavior

Cannot be focused nor look like it when disabled

### Actual behavior

Can be, yo!

### Steps to repro

1. Disable and click.

danethurber:
fix #307


packages/switch/src/react/index.js
```diff
@@ -116,6 +117,15 @@ class Switch extends React.Component {
   }
 }
 
+Switch.defaultProps = {
```
jaketrent: doesn't matter a bunch to me, but `defaultProps` have (probably) always come after `propTypes` in these components, at least the ones I just spot-checked.

danethurber: 👍 i'll move

PR #319, opened 2018-11-02
Refactor/text area halo

jaketrent:
Reuse new, simplified Halo focus indicator.


packages/halo/src/react/index.js
```diff
@@ -63,7 +63,6 @@ Halo.defaultProps = {
   appearance: vars.appearances.default,
   gapSize: vars.gapSizes.default,
   shape: vars.shapes.default,
-  className: '',
```
jaketrent: removed empty className attributes where no overrides given (see snapshot)

Issue #260, opened 2018-09-12
Modal Dialog focus accessibility issues


thatchej:
### Desired behavior

Users are unable to use Tab or Shift+Tab to navigate focus outside of a Modal Dialog

### Today's behavior

Currently it is possible to use Tab or Shift+Tab to move focus outside of the design system modal dialog. Technically speaking, this makes sense and is exactly the behavior you'd expect for a custom dialog/modal. However that so happens to be an accessibility violation

### The Value Add

This would remove the aforementioned accessibility issue across all BCs that use the design system. 

### The Tradeoffs

Must bring in additional 3rd party library or spend a non-trivial amount of time coding a solution. From my research this is a somewhat difficult issue to solve.

jaketrent:
A potential implementation from Andy L. https://pluralsight.slack.com/archives/C70HPSJPP/p1539732119000100?thread_ts=1539730876.000100&cid=C70HPSJPP

PR #321, opened 2018-11-02
dialog: focus lock

danethurber:
fixes: #260 

- [x] new `<FocusLock />`
- [x] `<Dialog />` uses focus lock

TODO: 

- [x] rename `<FocusLock />` to `<FocusManager />`
- [x] add `trapped` prop to activate/deactivate functionality
- [x] remove `as` prop
- [x] with no focusable children - make container focusable 
- [x] update `<Dialog />` to always use the `<FocusManager />`
- [x] `<Dialog />` is `trapped` if has `onClose` or `modal` props


packages/dialog/.storybook/config.js
```diff
@@ -1,9 +1,22 @@
...
+import core from '@pluralsight/ps-design-system-core'
+import centerDecorator from '@pluralsight/ps-design-system-storybook-addon-center'
+import themeDecorator from '@pluralsight/ps-design-system-storybook-addon-theme'
+
+const paddingDecorator = storyFn => (
+  <div style={{ padding: core.layout.spacingXLarge }}>{storyFn()}</div>
+)
+
+addDecorator(paddingDecorator)
+// addDecorator(centerDecorator)
```
jaketrent: rm? and import?

packages/dialog/src/react/index.js
```diff
@@ -51,39 +49,59 @@ const CloseButton = props => (
...
+    if (evt.keyCode === ESCAPE_KEYCODE) this.props.onClose(evt)
+  }
+
   render() {
     const { props } = this
     const dialogProps = {
       ...styles.dialog(props),
       ...(!this.props.disableCloseOnEscape &&
       typeof this.props.onClose === 'function'
-        ? { onKeyUp: handleKeyUp.bind(this, this.props) }
```
jaketrent: Why change?  Dislike the style?  Perf tweak?

packages/focuslock/.babelrc
```diff
@@ -0,0 +1,14 @@
+{
```
jaketrent: why create a whole new package for this?  It's only used in the Dialog.  Why not colocate?

packages/focuslock/src/react/index.js
```diff
@@ -0,0 +1,101 @@
+import PropTypes from 'prop-types'
+import React from 'react'
+
+const FOCUSABLE_SELECTORS = [
```
jaketrent: where did this list come from?

packages/focuslock/src/react/index.js
```diff
@@ -0,0 +1,101 @@
...
+    return (
+      <Tag ref={this.bindElement} onKeyDown={this.handleKeyDown}>
+        {children}
+      </Tag>
+    )
+  }
+}
+
+FocusLock.propTypes = {
+  as: PropTypes.oneOf(['div']).isRequired,
```
jaketrent: This is a pattern I've seen in Material UI, but we've yet to use it in our components.  

Thus far, when we've wanted to supply semantics, we've passed it as a first child, something like:

```
<FocusLock>
  <div>...
```

I like that approach better in general.  I'd rather have one approach than many.  But I haven't considered long enough whether that approach even works or feels appropriate in this case.  So, we might need another approach.  Or the advantages of this approach, so far unapparent to me, could be enough the we should migrate to it.  What is your take on this subject?

packages/focuslock/src/react/index.js
```diff
@@ -0,0 +1,101 @@
...
+  componentDidUpdate() {
+    this.focusableNodes = getFocusableChildNodes(this.el)
+  }
+
+  componentWillUnmount() {
+    const { prevFocusedElement: prev } = this
+    if (this.props.returnFocus && prev && prev.focus) prev.focus()
+  }
+
+  handleKeyDown(event) {
```
jaketrent: beautiful implementation

packages/focuslock/test/setup.js
```diff
@@ -0,0 +1,4 @@
+import Adapter from 'enzyme-adapter-react-16'
```
jaketrent: what would you think about migrating to react-testing-library?  Pros: We don't have to do version adapters.  We avoid the ginormous enzyme api.  We are forced more into blackbox testing and dom interrogation.

packages/focuslock/package.json
```diff
@@ -17,6 +17,9 @@
     "test:updateSnapshot": "npm run test -- --updateSnapshot",
     "test:watch": "npm run test -- --watchAll"
   },
+  "publishConfig": {
```
jaketrent: fancy!

packages/focuslock/src/react/__specs__/index.spec.js
```diff
@@ -3,15 +3,113 @@ import { mount } from 'enzyme'
...
 
+describe('FocusLock', () => {
   beforeEach(() => {
-    props = {
-      children: 'children'
-    }
+    document.activeElement.blur()
+  })
+
+  it('has defaultProps', () => {
```
jaketrent: I have never tested default props as a structure.  I would maybe rather test the outcome that the existence of a defaultProp is meant to create.  What value have you found in testing this in the past?

packages/focuslock/src/react/__specs__/index.spec.js
```diff
@@ -3,15 +3,113 @@ import { mount } from 'enzyme'
...
+    </li>
+  </ul>
+)
 
+describe('FocusLock', () => {
   beforeEach(() => {
-    props = {
-      children: 'children'
-    }
+    document.activeElement.blur()
```
jaketrent: oh, I love this.  Meant to "not focus on anything", right?

packages/focuslock/src/react/__specs__/index.spec.js
```diff
@@ -3,15 +3,113 @@ import { mount } from 'enzyme'
...
+`)
+  })
+
+  describe('WITH focusable children', () => {
+    it('renders successfully', () => {
+      expect(() => mount(<FocusLock>{children}</FocusLock>)).not.toThrow()
+    })
+  })
+
+  describe('WITHOUT focusable children', () => {
```
jaketrent: I love these test cases.  In other components, I may have put these in snapshot test, generated from storybook.  But on thinking about what you're asserting here, that different children scenarios are *possible* without some child state to check against probably fits well into this more traditional unit test.

packages/focuslock/src/react/__specs__/index.spec.js
```diff
@@ -3,15 +3,113 @@ import { mount } from 'enzyme'
...
+
+      const wrapper = mount(<FocusLock>{children}</FocusLock>)
+      const lastEl = wrapper.find('a[href]').last()
+
+      lastEl.getDOMNode().focus()
+      wrapper.unmount()
+
+      expect(document.activeElement).toEqual(button)
+    })
+  })
```
jaketrent: Do we need anything that tests the limits of the "focusable elements whitelist" in the component?

packages/dialog/src/react/__stories__/index.story.js
```diff
@@ -8,28 +9,95 @@ import * as Text from '@pluralsight/ps-design-system-text/react'
...
+    closeAction(event)
+  }
+
+  render() {
+    const { modalIsOpen } = this.state
+
+    return (
+      <div>
+        <Text.P>
+          Macaroon danish I love candy macaroon danish toffee muffin. Cotton
```
jaketrent: Mmmmm...

packages/dialog/src/react/index.js
```diff
@@ -87,32 +90,29 @@ class Dialog extends React.Component {
...
-  }
-
   handleKeyUp(evt) {
     if (evt.keyCode === ESCAPE_KEYCODE) this.props.onClose(evt)
   }
 
   render() {
     const { props } = this
+
+    const hasFocusLock = !!props.onClose
```
jaketrent: should include `|| props.modal` ? (the case where we have a modal that doesn't feature a close button/callback)

packages/dialog/src/react/index.js
```diff
@@ -87,32 +90,29 @@ class Dialog extends React.Component {
     this.handleKeyUp = this.handleKeyUp.bind(this)
   }
 
-  componentDidMount() {
-    if (this.el && !this.props.disableFocusOnMount) this.el.focus()
```
jaketrent: What was your reasoning in getting rid of this?  I can't see where its function is handled elsewhere.

packages/dialog/src/react/index.js
```diff
@@ -60,17 +60,16 @@ class ModalOverlay extends React.Component {
...
   }
 
   render() {
-    const { disableCloseOnOverlayClick } = this.props
-
     return (
       <div
         {...styles.overlay(this.props)}
         id={MODAL_OVERLAY_ID}
-        onClick={!disableCloseOnOverlayClick && this.handleOverlayClick}
```
jaketrent: I like the change.  Just interested in what the warning was that was being addressed.

packages/dialog/.storybook/config.js
danethurber: i'll fix

packages/focuslock/.babelrc
danethurber: i have plans for reuse

packages/focuslock/src/react/index.js
danethurber: i will removed - we add in support for additional tags if needed later

packages/focuslock/src/react/__specs__/index.spec.js
danethurber: let's leave them out

packages/dialog/src/react/index.js
danethurber: i will add that prop support

packages/dialog/src/react/index.js
danethurber: i will update the `FocusLock` to make it's self focusable if no focusable children

danethurber: Dialog(that aren't modal/onClose) should also focus on mount

danethurber: update FocusLock to have active/inactive or locked/unlocked. rename FocusManager? trapped(bool?)

danethurber:
@jaketrent this is ready for you to give it a second pass

jaketrent:
Added one relevant story.  Please have a look.  Overall, looks great to me, plz merge.

jaketrent:
Now that I've committed, I'm not sure I have an approve/review button any more.

danethurber:
that's weird

PR #324, opened 2018-11-02
StarRating: New Component for rating and displaying ratings

TravisBumgarner:
### What You're Solving

- Creating a component that didn't exist previously

### Design Decisions

- This problem was a bit difficult to solve because the logic was a bit difficult to grok. The logic being whether something is being hovered or not, whether it is dark or light, whether the value to display included full stars/empty stars/half star. 
- The `HalfStarIcon` was also difficult to incorporate due to being part of the `StarRating` and not an `Icon`.
- I've gotten feedback from a few people already and think it's about time to open this up for a more general review.


packages/starrating/package.json
```diff
@@ -0,0 +1,48 @@
...
+    "@pluralsight/ps-design-system-build": "^1.8.3",
+    "@pluralsight/ps-design-system-storybook-addon-theme": "^1.1.44",
+    "@storybook/addon-storyshots": "3.2.14",
+    "@storybook/addons": "3.2.14",
+    "@storybook/react": "3.2.14",
+    "babel-cli": "^6.24.1",
+    "babel-preset-env": "^1.4.0",
+    "babel-preset-react": "^6.24.1",
+    "babel-preset-stage-2": "^6.24.1",
+    "jest": "^21.2.1",
```
jaketrent: Dane has been upgrading these versions.  He'll take care of this.

packages/starrating/src/css/index.js
```diff
@@ -0,0 +1,17 @@
+export default {
+  button: {
```
jaketrent: We use BEM in our css files because we export stylesheets.  For an example, please see:

https://github.com/pluralsight/design-system/blob/master/packages/button/src/css/index.js#L16

packages/starrating/src/react/half-star-dark.js
```diff
@@ -0,0 +1,58 @@
+import React from 'react'
+
+const DarkHalfStar = () => (
+  <svg
+    width="24"
```
jaketrent: We usually try to avoid styling and static dimensions in svg.  Rely on viewBox for dimensions and css for the styling.

packages/starrating/src/react/half-star-dark.js
```diff
@@ -0,0 +1,58 @@
...
+    height="24"
+    viewBox="0 0 24 24"
+    fill="none"
+    xmlns="http://www.w3.org/2000/svg"
+  >
+    <mask
+      id="mask0"
+      mask-type="alpha"
```
jaketrent: react-compatible?

packages/starrating/src/react/half-star-light.js
```diff
@@ -0,0 +1,60 @@
+import React from 'react'
+
+const HalfLightStar = () => (
```
jaketrent: Could we remove the 2nd star if we moved styling into css?  That would allow us to color one shape and only need that one shape in svg.

packages/starrating/src/react/index.js
```diff
@@ -0,0 +1,179 @@
...
+    )
+  }
+}
+
+StarRating.propTypes = {
+  value: PropTypes.number,
+  onClick: PropTypes.func
+}
+
+StarRating.defaultProps = {
```
jaketrent: Not required if defaults are nulls.

packages/starrating/src/react/index.js
```diff
@@ -0,0 +1,179 @@
...
+    const { value } = this.props
+
+    const halfIntRoundedValue = Math.floor(value * 2) / 2
+
+    let fullStars = Math.floor(halfIntRoundedValue)
+    let halfStars = Math.floor(halfIntRoundedValue) !== halfIntRoundedValue
+
+    const isHovering = hoverIndex !== null
+
+    return new Array(TOTAL_STARS).fill(undefined).map((_, index) => {
```
jaketrent: Array.fill need a polyfill in IE?

packages/starrating/src/react/index.js
```diff
@@ -0,0 +1,179 @@
...
+      const ratingValue = index + 1
+      const ratingString = `Rate ${ratingValue} star${
+        ratingValue > 1 ? 's' : ''
+      }`
+      return (
+        <button
+          key={index}
+          tabIndex={-1}
+          style={css.button}
+          onClick={() => onClick(ratingValue)}
```
jaketrent: Usually prefer bound functions.  Perf boost.  Can test indep.

packages/starrating/stories/index.js
```diff
@@ -0,0 +1,26 @@
...
+const valueStory = storiesOf('value', module)
+;[null, 1, 1.5, 2, 2.5, 3, 3.5, 4, 4.5, 5].forEach(value =>
+  valueStory.add(`value ${value}`, _ => <StarRating value={value} />)
+)
+
+const onClickStory = storiesOf('onClick', module)
+onClickStory.add('onClick', _ => <StarRating value={4.5} onClick={alert} />)
+onClickStory.add('!onClick', _ => <StarRating value={4.5} />)
+
+const themeStory = storiesOf('theme', module)
```
jaketrent: There's a theme decorator that you've imported on line 7 that would be great for this.  (All stories can react to Theme that way.)

packages/starrating/package.json
```diff
@@ -0,0 +1,48 @@
...
+  "license": "Apache-2.0",
+  "repository": "pluralsight/design-system",
+  "peerDependencies": {
+    "@pluralsight/ps-design-system-normalize": "^3.0.24",
+    "glamor": "^2.20.0",
+    "react": ">=0.15.0 < 17.0.0"
+  },
+  "dependencies": {
+    "@pluralsight/ps-design-system-core": "^4.3.3",
+    "@pluralsight/ps-design-system-theme": "^1.3.2",
```
jaketrent: Missing `@pluralsight/ps-design-system-icon` package.

packages/starrating/src/react/index.js
```diff
@@ -0,0 +1,179 @@
...
+        <span style={css.screenReaderContent}>{`This is rated ${value}`}</span>
+        {StarIcons}
+      </div>
+    )
+  }
+}
+
+StarRating.propTypes = {
+  value: PropTypes.number,
+  onClick: PropTypes.func
```
jaketrent: Feel like `onChange` might be more appropriate because it returning the new value and its change over time from the component.

packages/starrating/src/react/index.js
```diff
@@ -0,0 +1,179 @@
...
+    StarIcons = onClick ? this.makeClickable(StarIcons) : StarIcons
+    return (
+      <div>
+        {onClick && (
+          <label>
+            <span style={css.screenReaderContent}>Rate</span>
+            <input
+              style={css.screenReaderContent}
+              type="range"
+              onChange={event => onClick(event.target.value)}
```
jaketrent: Please return the original event as the 2nd param in the callback.

dborodyansky:
@TravisBumgarner Excited to see this coming to life! Thank you.

I would like to note a clarification on the `interactive` hover/selected visual treatment. Please let me know if you may have questions on the following:
image

TravisBumgarner:
@dborodyansky What does the state of `not clicked` and `has rating when loaded` look like?

dborodyansky:
> @dborodyansky What does the state of `not clicked` and `has rating when loaded` look like?

@TravisBumgarner It would look same as `clicked`(same as read-only), although unlikely. For the most part the component will be read-only in context. We want to support interactivity but it wouldn't be a default behavior wherever the star ratings appear. Make sense?

packages/starrating/package.json
TravisBumgarner: @danethurber Does it make sense to just leave this like this or is there a version I should revert it to?

TravisBumgarner: I'm not sure I'll be able to update snapshots on my local machine if I revert this.

danethurber: all the other packages have been upgraded to jest@23. i've done quite a bit of work to the project and have sped up installs and bootstrapping so it should be faster. give me an hour or so and i'll send you a pr

packages/starrating/src/react/index.js
TravisBumgarner: Does this apply to these as well?
```          onClick={() => onChange(ratingValue)}
          onMouseOver={() => this.setState({ hoverIndex: index })}
          onMouseLeave={() => this.setState({ hoverIndex: null })}```

packages/starrating/src/css/index.js
TravisBumgarner: Would you happen to know of a class component that does this? I'm a bit confused by what's going on.

packages/starrating/package.json
danethurber: pr opened -> https://github.com/TravisBumgarner/design-system/pull/1

let's compare together :)

TravisBumgarner:
Thanks for the feedback! Unfortunately it looks like I'm having installation problems again with lerna and I can't work on this again next week. Most of the issues are out of my area of knowledge, if anyone thinks they'd be able to quickly make the changes, I'd appreciate the help. Otherwise, I can take a look next week.

jaketrent:
Hey Travis, I haven't gotten to this this week either.  I'll be happy to help with it in the coming week.  I know Dane is as well.

danethurber:
new pr tracked here -> https://github.com/pluralsight/design-system/pull/339

PR #325, opened 2018-11-05
fix: linting on commit

danethurber:


package.json
```diff
@@ -31,17 +31,18 @@
       "scripts/create-component/package-template/**/*"
     ],
     "linters": {
+      "**/*.css": [
+        "prettier-eslint --write",
```
danethurber: format css files

package.json
```diff
@@ -31,17 +31,18 @@
...
+        "git add"
+      ],
       "**/*.js": [
         "prettier-eslint --write",
+        "eslint",
```
danethurber: i was confused. `prettier-eslint` just formats the file. we can use `eslint` to do the linting

package.json
```diff
@@ -31,17 +31,18 @@
...
         "git add"
       ],
       "package.json": [
         "format-package -w",
         "git add"
-      ],
-      "packages/**/*.css": [
-        "node_modules/.bin/prettier --single-quote --no-semi --write",
```
jaketrent: Without the single quote and no semi flags, are those getting read from somewhere else that's committed to the project?  We should get a .prettierrc in here, but we don't have one yet.

jaketrent: I am blind. :)

PR #326, opened 2018-11-05
datepicker: update value when prop changes

### LINKED ISSUE #304: Date Picker: no visible update after changing value prop
### Expected behavior
Date picker should be updated after having changed the property "value" after the initial value had already been set up.

### Actual behavior
After calling the setState for value which is connected to the property "value". the render method is executed, the new value is provided, the attribute of date picker input is updated, but mm, dd, yyyy inputs and calendar popup value are not. 
Applying such code fixes the problem
```
componentDidUpdate(prevProps: Readonly<any>, prevState: Readonly<any>, snapshot?: any): void {
    if (prevState.value === this.state.value) {
      return;
    }

    if (this.reactRef && this.reactRef.current) { // React.createRef()
      this.reactRef.current.handleCalendarSelect(this.state.value);
    }
  }
```

danethurber:
### What You're Solving

- update/format datepicker package
- fixes #304


packages/datepicker/src/react/index.js
```diff
@@ -173,6 +173,12 @@ class DatePicker extends React.Component {
     this.handleCalendarSelect = this.handleCalendarSelect.bind(this)
     this.handleIconClick = this.handleIconClick.bind(this)
   }
+  componentWillReceiveProps(nextProps) {
```
danethurber: this is the bug fix

packages/datepicker/package.json
```diff
@@ -40,16 +41,19 @@
...
+    "@pluralsight/ps-design-system-storybook-addon-theme": "*",
+    "@pluralsight/ps-design-system-textinput": "*",
     "@storybook/addon-actions": "^4.0.0",
     "@storybook/addon-storyshots": "^4.0.0",
     "@storybook/react": "^4.0.0",
     "babel-cli": "^6.24.1",
     "babel-preset-env": "^1.6.0",
     "babel-preset-react": "^6.24.1",
     "babel-preset-stage-2": "^6.24.1",
+    "enzyme": "^3.2.0",
```
jaketrent: What would you think of us migrating our project to react-testing-lib?  I like it because: avoids ginormous enzyme api and encourages more blackbox testing and avoids having to do an adapter for every react version.

danethurber: i have no concerns if you want to try it. i haven't used it before - i will check it out

PR #327, opened 2018-11-05
refactor(dropdown): use halo for focus indicator

jaketrent:


packages/dropdown/src/react/index.js
```diff
@@ -1,6 +1,5 @@
-// TODO: maybe? back the visible ui with a native select that an be submitted with <form target=?? /> browser submission
-import ActionMenu from '@pluralsight/ps-design-system-actionmenu/react'
```
jaketrent: unused

packages/dropdown/src/react/index.js
```diff
@@ -135,11 +100,11 @@ class Dropdown extends React.Component {
     this.setState({ isOpen: !this.state.isOpen })
     if (typeof this.props.onClick === 'function') this.props.onClick(evt)
   }
-  handleFocus() {
+  handleFocus(evt) {
```
jaketrent: unrelated bug possibility

Issue #318, opened 2018-11-02
halo: focus blue should show on error'ed components


jaketrent:
### Expected behavior

Focus on a form control component that is in `error` state.  Should still show blue focus indicator.

### Actual behavior

Shows red halo with no focus indicator.

### Steps to repro

1. Go to https://design-system.pluralsight.com/components/textinput/#error
2. Click in TextInput
3. See red

PR #329, opened 2018-11-05
Halo adjustment and form-related usage

### LINKED ISSUE #289: form-related: focus ring not showing up

jaketrent:
Halo change to not require js focus state.  Refactor checkbox integration to use new Halo.  If this looks good, I'll do the other component adjustments.  If we're still not sure if the Halo change will support all other use cases, I can do the questionable ones first to make sure before any merge. 

Fixes #289, #318


jaketrent:
Hey @danethurber , this now includes all components refactored to use the adjusted Halo component.  There were a couple more tweaks to Halo as well.  Please take another look when you have a chance.  Totally fine to wait until your return.  Thanks!

PR #338, opened 2018-11-13
switch: update to add ps-blue color option

deltaepsilon:
### What You're Solving

We need ps blue instead of orange or green

### How to Verify

Build and open it up in Storybook


jaketrent:
Thanks for the add, @deltaepsilon

PR #339, opened 2018-11-13
StarRating

danethurber:
OUTSTANDING TASKS:

- [x] convert to BEM classnames
- [x] integrate glamor build and common glamor style pattern
- [x] refactor to use Icon w/props
- [x] style Icon using css
- [x] simplify svgs needed to do theming
- [x] refactor to use react conventions
- [x] break into smaller pure components
- [x] refactor to expose `onChange` prop
- [x] `onChange` should surface original event
- [x] interactive with no rating should show outlined star icon
- [x] fix half star mask color
- [x] keyboard support
- [x] remove `polished`
- [x] ~docs~ moved to https://github.com/pluralsight/design-system/issues/340

I'm opening a new branch in the main repo so I can do some refactoring.
Original PR from @TravisBumgarner  here -> #324

---

### What You're Solving

- Creating a component that didn't exist previously

### Design Decisions

- This problem was a bit difficult to solve because the logic was a bit difficult to grok. The logic being whether something is being hovered or not, whether it is dark or light, whether the value to display included full stars/empty stars/half star. 
- The `HalfStarIcon` was also difficult to incorporate due to being part of the `StarRating` and not an `Icon`.
- I've gotten feedback from a few people already and think it's about time to open this up for a more general review.


packages/starrating/src/css/index.js
```diff
@@ -0,0 +1,79 @@
...
+    ...resetButton,
+    lineHeight: 0
+  },
+  [`${BASE_CLASSNAME}__star--active`]: {
+    color: core.colors.yellow
+  },
+  [`${BASE_CLASSNAME}__star--interactive`]: {
+    cursor: 'pointer'
+  },
+  [`${BASE_CLASSNAME}__star--theme-dark`]: {
```
jaketrent: Thinking/writing about interpolation above makes me realize that we've used the vars in places like these elsewhere.  Do you feel like that has value?

Example: https://github.com/pluralsight/design-system/blob/master/packages/button/src/css/index.js#L43

My thought in the past has been that it ties our selectors to the code that drives those features, thus removing a magic string that can get out of sync if naming changes. 🤷‍♂️

packages/starrating/src/react/__stories__/index.story.js
```diff
@@ -0,0 +1,18 @@
...
+import { storiesOf } from '@storybook/react'
+
+import StarRating from '..'
+
+const VALUES = [0, 0.5, 1, 1.5, 2, 2.5, 3, 3.5, 4, 4.5, 5, null, undefined]
+
+const interactive = storiesOf('StarRating / interactive', module)
+const nonInteractive = storiesOf('StarRating / not interactive', module)
+
+VALUES.forEach(value => {
```
jaketrent: elegant story!

packages/starrating/src/react/index.js
```diff
@@ -0,0 +1,155 @@
...
+import PropTypes from 'prop-types'
+import React from 'react'
+
+import { withTheme } from '@pluralsight/ps-design-system-theme/react'
+
+import Star from './star'
+
+import css, { BASE_CLASSNAME } from '../css'
+
+const styles = {
```
jaketrent: so clean!

packages/starrating/src/react/index.js
```diff
@@ -0,0 +1,155 @@
...
+
+class StarRating extends React.PureComponent {
+  constructor(props) {
+    super(props)
+
+    this.handleStarClicked = this.handleStarClicked.bind(this)
+    this.handleStarEnter = this.handleStarEnter.bind(this)
+    this.handleStarLeave = this.handleStarLeave.bind(this)
+
+    this.state = { hoverIndex: null, interactive: !!this.props.onChange }
```
jaketrent: Does interactive need to be on state?  Doesn't seem to be something that would change much.  I'll keep reading to find out more...

If was not on state, would not need componentWillReceiveProps

packages/starrating/src/react/index.js
```diff
@@ -0,0 +1,155 @@
...
+    if (!this.state.interactive) return
+    this.setState(() => ({ hoverIndex: index }))
+  }
+
+  handleStarLeave(index, event) {
+    if (!this.state.interactive) return
+    this.setState(() => ({ hoverIndex: null }))
+  }
+
+  get stars() {
```
jaketrent: I don't use getters often.  Teach me, please, why do you reach for a getter in this case?

packages/starrating/src/react/index.js
```diff
@@ -0,0 +1,155 @@
...
+    const { hoverIndex, interactive } = this.state
+    const { starCount, value } = this.props
+
+    const hasValidRating = !Number.isNaN(parseFloat(value, 10))
```
jaketrent: parseFloat does not have radix arg.

packages/starrating/src/react/index.js
```diff
@@ -0,0 +1,155 @@
...
+
+    const valueRounded = Math.floor(value * 2) / 2
+    const fullStars = Math.floor(valueRounded)
+    const halfStars = Math.floor(valueRounded) !== valueRounded
+
+    return [...Array(starCount)].map((_, index) => {
+      let active = false
+      let appearance
+
+      if (isHovering) {
```
jaketrent: these are some complicated rules, but you've made them as clean as possible, it seems.  Great job!

packages/starrating/src/react/index.js
```diff
@@ -0,0 +1,155 @@
...
+
+        <ScreenReaderText>This is rated {value}</ScreenReaderText>
+      </div>
+    )
+  }
+}
+
+StarRating.propTypes = {
+  onChange: PropTypes.func,
+  starCount: PropTypes.number,
```
jaketrent: fancy!  not strictly a requirement.  Should we keep?

packages/starrating/src/react/star.js
```diff
@@ -0,0 +1,146 @@
...
+          <symbol id="star">
+            <path d="M7.71590761,13.8522772 L2.25298114,9.9052852 C2.0291483,9.74356485 1.97879619,9.43101223 2.14051654,9.2071794 C2.23452325,9.07706721 2.38528246,9 2.54580174,9 L9.32999992,9 L11.5259342,2.42042174 C11.6133563,2.15848284 11.8965694,2.01700903 12.1585083,2.10443114 C12.3076212,2.15419757 12.4246481,2.2711924 12.4744554,2.42029163 L14.6724396,9 L21.4541983,9 C21.7303406,9 21.9541983,9.22385763 21.9541983,9.5 C21.9541983,9.66051929 21.877131,9.81127849 21.7470189,9.9052852 L16.2840924,13.8522772 L18.5094075,20.5282225 C18.5967314,20.7901941 18.4551514,21.0733541 18.1931797,21.160678 C18.0405447,21.2115563 17.8727519,21.1858764 17.7423206,21.091676 L12.0021534,16.9459997 L6.25731585,21.0924971 C6.03340576,21.2541105 5.72087724,21.203609 5.55926386,20.9796989 C5.46513169,20.8492818 5.43948525,20.6815443 5.49034729,20.5289581 L7.71590761,13.8522772 Z" />
+          </symbol>
+
+          <mask id="maskHalf">
+            <rect width="50%" height="100%" fill="white" />
+          </mask>
+        </defs>
+
+        <use xlinkHref="#star" {...styles.halfStarSecondary(props)} />
```
jaketrent: Surprised this works.  Last time I tried to use xlink in svg in react, it didn't work.  Is this a recent development?

packages/starrating/src/react/star.js
```diff
@@ -0,0 +1,146 @@
...
+  return (
+    <Icon {...filteredProps}>
+      <svg
+        role="img"
+        aria-label="star half icon"
+        viewBox="0 0 24 24"
+        xmlns="http://www.w3.org/2000/svg"
+      >
+        <defs>
+          <symbol id="star">
```
jaketrent: Do these ids need namespaced to the DS and the component? Might be good to be a good citizen inside others' apps.

packages/starrating/src/react/star.js
```diff
@@ -0,0 +1,146 @@
...
+        {...filteredProps}
+        aria-label={this.label}
+        title={this.label}
+        onBlur={this.handleLeave}
+        onClick={this.handleClicked}
+        onFocus={this.handleEnter}
+        onMouseEnter={this.handleEnter}
+        onMouseLeave={this.handleLeave}
+      >
+        {appearance === 'full' && <Icon id={Icon.ids.starFill} size={size} />}
```
jaketrent: Opp to use the named vars instead of magic strings like 'full'.

packages/starrating/vars.js
```diff
@@ -0,0 +1 @@
+module.exports = require('./dist/vars')
```
jaketrent: I don't think there are any vars in this package.  I am looking through it wondering if there should be, but I'm not coming up with the traditional vars that are usually shared between react and css.  For instance, APPEARANCES is only used in react/star.js.  Do you know of any?

packages/theme/src/react/__specs__/with-default-theme.spec.js
```diff
@@ -1,11 +1,11 @@
 import { mount } from 'enzyme'
```
jaketrent: Might be reading the diff wrong, but might want to rename the file too.

packages/theme/src/react/index.js
```diff
@@ -3,6 +3,8 @@ import React from 'react'
 
 import * as vars from '../vars'
 
+export { default as withTheme } from './with-theme'
```
jaketrent: Does this require extra babel plugin?  seems like exporting from an import is not supported in es6 modules properly atm.  Seems like there was/is a proposal as well, so I might be wrong.  I guess you're using it, so we've proved it's ok....  Mind just gut-checking my thought here?

packages/starrating/src/css/index.js
```diff
@@ -0,0 +1,79 @@
...
+  '&::-moz-focus-inner': {
+    border: 0,
+    padding: 0
+  }
+}
+
+export default {
+  [BASE_CLASSNAME]: {},
+
+  [`${BASE_CLASSNAME}__star`]: {
```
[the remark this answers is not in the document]
danethurber: i actually have mixed feelings. I'm doing it because it seems like it could catch small(hard to find) typos...but i don't love the way it looks.

packages/starrating/src/css/index.js
danethurber: yeah i didn't think about that. i like using the the vars for this. i'll updated

packages/starrating/src/react/star.js
danethurber: let me look into it. svg defs might do something like shadow dom but i'm not sure. i think namespacing would be good

packages/starrating/src/react/star.js
danethurber: 👍

packages/theme/src/react/__specs__/with-default-theme.spec.js
danethurber: you're right!

packages/theme/src/react/index.js
danethurber: i think we're ok but i can look into the support

packages/starrating/src/react/index.js
danethurber: thanks. will update

packages/starrating/src/react/index.js
danethurber: the render method was getting messy and i wanted to encapsulate some rendering logic without creating a new component. i generally reach for a getter when i need this but i also get weird looks :)

packages/starrating/src/react/index.js
danethurber: i wouldn't document it as an option but I would define the propType is this case. the other idea would be to not make it prop and define elsewhere. this seemed easy enough but i could go either way

packages/starrating/src/react/star.js
danethurber: as far as i know with the def+symbol this is the way to do it

packages/starrating/src/react/index.js
danethurber: i could just do the check inline - or use a getter to encapsulate the logic. doing it here is faster...but not in any noticeable way

packages/starrating/src/react/star.js
danethurber: > Since SVG 2, the xlink:href attribute is deprecated in favor of simply href.

FYI

packages/starrating/vars.js
danethurber: no there are no vars to export

danethurber: if we add sizes(probably reexport icon sizes) there will be but right now there is only the one size

packages/theme/src/react/index.js
danethurber: if we wanted `default from` it's a new plugin but this seems to work 🤷‍♂️

danethurber: our current code works by compiling too

```
var _withTheme = require('./with-theme');

Object.defineProperty(exports, 'withTheme', {
  enumerable: true,
  get: function get() {
    return _interopRequireDefault(_withTheme).default;
  }
});
```

packages/starrating/src/react/index.js
jaketrent: I didn't remember to come back and update this comment, but I'd sure like to not use state if not required. 

I'm thinking of it this way: Prefer props, they're simpler, incoming/immutable data belongs based on react conventions, react state is for internal changing data, which this is not.

It reads well elsewhere (the `interactive` flag is easy to grok).  But why it's on `state` is not easy to understand and seems to be unconventional.  

I'd also be happy to do a refactor after merge.

packages/starrating/src/react/index.js
jaketrent: I'm fine for the moment.  Don't worry about it.  I've had enough experiences adding easy things that were required only to have problems and support effort required to maintain them.  There's also the need to work around these things in the future as we read and code in this component. It's also way easier to add things as they're needed vs retiring them. 
So I try to be careful not to over deliver on things not specifically needed so I can ensure more time for what is known to be needed.  But, you know, sometimes there are extra features we add that end up being unexpectedly used and successful, so it's hard to say "never".

packages/starrating/src/react/star.js
```diff
@@ -40,17 +40,23 @@ const HalfStarIcon = props => {
...
             <path d="M7.71590761,13.8522772 L2.25298114,9.9052852 C2.0291483,9.74356485 1.97879619,9.43101223 2.14051654,9.2071794 C2.23452325,9.07706721 2.38528246,9 2.54580174,9 L9.32999992,9 L11.5259342,2.42042174 C11.6133563,2.15848284 11.8965694,2.01700903 12.1585083,2.10443114 C12.3076212,2.15419757 12.4246481,2.2711924 12.4744554,2.42029163 L14.6724396,9 L21.4541983,9 C21.7303406,9 21.9541983,9.22385763 21.9541983,9.5 C21.9541983,9.66051929 21.877131,9.81127849 21.7470189,9.9052852 L16.2840924,13.8522772 L18.5094075,20.5282225 C18.5967314,20.7901941 18.4551514,21.0733541 18.1931797,21.160678 C18.0405447,21.2115563 17.8727519,21.1858764 17.7423206,21.091676 L12.0021534,16.9459997 L6.25731585,21.0924971 C6.03340576,21.2541105 5.72087724,21.203609 5.55926386,20.9796989 C5.46513169,20.8492818 5.43948525,20.6815443 5.49034729,20.5289581 L7.71590761,13.8522772 Z" />
           </symbol>
 
-          <mask id="maskHalf">
+          <mask id="psds-starrating__half-star__maskHalf">
             <rect width="50%" height="100%" fill="white" />
           </mask>
         </defs>
 
-        <use xlinkHref="#star" {...styles.halfStarSecondary(props)} />
```
jaketrent: What did you find out on this?  Because it was rendering before... was there a problem?

jaketrent:
Hey Dane, I had another look.  I left a few comments.  If you feel good, I feel good. Merge away.

packages/starrating/src/react/index.js
danethurber: i'll do a quick refactor into a getter and merge. we can sync up after and come up with a good approach

packages/starrating/src/react/star.js
danethurber: xlink-href is deprecated in favor of just href. i updated

PR #341, opened 2018-11-14
button: accept download prop when button is an anchor element

mavs0890:
Download is a valid attribute for anchor tags. Since the Button component uses anchor tags when an href attribute is present, it should allow the download attribute.


jaketrent:
Resolved with #342.

PR #345, opened 2018-11-16
StarRating tweaks

danethurber:
TODOS:

- [x] Default icon size should be 16px 
- [x] Update light theme colors (get from Dmitri)
- [x] Update docs on null/undefined usages in interactive mode
- [x] Update docs to have a space -> `Star Rating`

### DIFF
--- a/packages/site/pages/components/starrating.js
+++ b/packages/site/pages/components/starrating.js
@@ -77,21 +77,18 @@ export default withServerProps(_ => (
         themeToggle
         includes={{ StarRating }}
         codes={[
-          `<StarRating value={2.5} onChange={val => alert("The new value is: " + val)} />`
+          `<StarRating value={null} onChange={val => alert("The new value is: " + val)} />`
         ]}
       />
 
       <P>
-        A special empty state will be display if the{' '}
-        <Text.Code>onChange</Text.Code> prop is defined but your value is empty.
+        If a value exists, the component may still be interactive on mouse hover
       </P>
       <Example.React
         themeToggle
         includes={{ StarRating }}
-        orient="vertical"
         codes={[
-          `<StarRating value={undefined} onChange={() => null} />`,
-          `<StarRating value={null} onChange={() => null} />`
+          `<StarRating value={2.5} onChange={val => alert("The new value is: " + val)} />`
         ]}
       />
     </Content>
--- a/packages/site/pages/components/starrating.js
+++ b/packages/site/pages/components/starrating.js
@@ -17,8 +17,8 @@ import {
 
 export default withServerProps(_ => (
   <Chrome>
-    <Content title="StarRating">
-      <PageHeading packageName="starrating">StarRating</PageHeading>
+    <Content title="Star Rating">
+      <PageHeading packageName="starrating">Star Rating</PageHeading>
 
       <P>Install the component dependency:</P>
       <Code language="bash">
@@ -78,6 +78,20 @@ export default withServerProps(_ => (
         includes={{ StarRating }}
         codes={[
           `<StarRating value={2.5} onChange={val => alert("The new value is: " + val)} />`
+        ]}
+      />
+
+      <P>
+        A special empty state will be display if the{' '}
+        <Text.Code>onChange</Text.Code> prop is defined but your value is empty.
+      </P>
+      <Example.React
+        themeToggle
+        includes={{ StarRating }}
+        orient="vertical"
+        codes={[
+          `<StarRating value={undefined} onChange={() => null} />`,
+          `<StarRating value={null} onChange={() => null} />`
         ]}
       />
     </Content>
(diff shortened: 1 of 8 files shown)


packages/starrating/src/react/star.js
```diff
@@ -145,12 +146,13 @@ Star.propTypes = {
   onClick: PropTypes.func,
   onEnter: PropTypes.func,
   onLeave: PropTypes.func,
+  size: PropTypes.oneOf(Object.values(Icon.sizes)).isRequired,
```
jaketrent: So we have variable sizes of starrating now?  Is that a recent decision?

packages/starrating/src/react/index.js
```diff
@@ -85,6 +86,7 @@ class StarRating extends React.PureComponent {
       } else {
         if (this.isInteractive && !hasValidRating) {
           appearance = Star.appearances.empty
+          bright = true
```
jaketrent: So, the star is "bright" when it's clickable and hasn't been rated yet...  Lots of rules here.

packages/starrating/src/css/index.js
```diff
@@ -49,27 +49,32 @@ export default {
...
+    color: core.colors.gray04
   },
   [`${BASE_CLASSNAME}__star__half__secondary--theme-${themeNames.light}`]: {
     color: core.colors.gray01
   },
 
+  [`${BASE_CLASSNAME}__star--active`]: {
+    color: core.colors.yellow
+  },
+  [`${BASE_CLASSNAME}__star--bright`]: {
```
jaketrent: "Bright" may very well be the best name for this, I'm having a thought that might be something good to consider: traditionally we've tried to default these modifier names to match the props as closely as possible.  Two basic reasons I feel have been worthwhile: Keep from introducing new variable names for things (just call it the same idea throughout).  The other is to try to make the style rules more apparent in the css itself rather than having to look through the component internals.  

In this case, to get there, we could create a selector like:

```
.psds-starrating__star--interactive.psds-starrating__star--not-value {
 ...
}
```

It's not as short as "bright" but it doesn't intro new names and describes what state it's styling in the selector.

Anyway, something to consider.

jaketrent: If the `bright` prop was removed, this state could be calculated in the `Star` itself so you didn't have to pass the prop either.

packages/starrating/src/react/star.js
```diff
@@ -7,7 +7,7 @@ import { withTheme } from '@pluralsight/ps-design-system-theme/react'
 
 import css, { BASE_CLASSNAME } from '../css'
 
-const APPEARANCES = {
+export const APPEARANCES = {
```
jaketrent: I'm not seeing where these are used outside this module.

packages/site/pages/components/starrating.js
```diff
@@ -80,6 +80,20 @@ export default withServerProps(_ => (
           `<StarRating value={2.5} onChange={val => alert("The new value is: " + val)} />`
         ]}
       />
+
+      <P>
+        A special empty state will be display if the{' '}
```
jaketrent: Special *how* or *why* might be useful prose to write.  At least it seems so to me as one to whom the requirement reasoning is unknown.

packages/starrating/src/react/star.js
danethurber: oops - will remove

packages/starrating/src/react/index.js
danethurber: srsly. that's why i started writing unit tests

packages/starrating/src/react/star.js
danethurber: we don't. i added the proptype because it didn't catch the undefined error before. `Icon.sizes.xsmall ` didn't exist. i can move size out of a proptype to make this easier to understand

packages/site/pages/components/starrating.js
danethurber: tell me more. i'm not a great author :/

packages/starrating/src/css/index.js
danethurber: yeah this is a tricky one. there are a lot of variations and i'm trying to keep the individual Star component as simple as possible while doing all the logic in the StarRating. i don't have the perfect answer but i think keeping these props as presentational give a good about of flexibility...let me noodle on it

packages/site/pages/components/starrating.js
danethurber: i'm gonna ask dmitri why there is a special case

jaketrent: Hehe, yeah, I guess we both need more context if we're going to try to give context to other devs using the component.  You're a great author, btw!  This was just a potential enhancement to the "special style" section.

packages/starrating/src/css/index.js
jaketrent: Yes, straight-forward Star component also seems desirable.  Might be a competing priority.  I'm not sure which one has higher priority.

dborodyansky:
@danethurber I think the default size adjustment (16x16) is still outstanding.

danethurber:
@dborodyansky i needed to publish the changed. should be correct now

PR #335, opened 2018-11-16
Add a link to "Add Issue" that persists on the page in the Docs that links to this issue tracker

danethurber:
## Instructions
I had a difficult time finding this page for adding an issue to the design system


packages/site/src/ui/open-issue-prompt.js
```diff
@@ -0,0 +1,35 @@
...
+          Open an issue
+        </Button>
+
+        <style jsx>{`
+          .open-issue-prompt {
+            position: fixed;
+            bottom: 20px;
+            right: 20px;
+          }
+          .open-issue-prompt > :global(a) {
```
jaketrent: This feels like a weakness of styled-jsx.

danethurber: definitely - i tried to work around it but came up short

PR #333, opened 2018-11-19
card: add light theme

danethurber:
TODO:

- [x] remove glamourous
- [x] make a package for filter prop fn(fn not hoc or render prop)
- [x] add light theme styles


danethurber:
@jaketrent i'm going block this until after the break when we can talk about a few things. when removing glamorous i needed to find a way to filter props so that the update was non-breaking. i implemented a solution but i really want to get your input :)

packages/card/src/react/with-prop-filter.js
```diff
@@ -0,0 +1,87 @@
...
+
+  return (
+    isHTMLProp(tagName, propName) ||
+    isReactProp(propName) ||
+    (isCustomAttr(propName.toLowerCase()) &&
+      (!isCSSProp(propName) || isSVGTag(tagName)))
+  )
+})
+
+export default function withPropFilterFactory(_config) {
```
danethurber: i'm rethinking this and I think i'm going to refactor to use a render prop component. I think it will work better and i might be able to get ride of the need for whitelist/blacklist

danethurber:
@jaketrent when you get a chance please review. i also needed some guidance with updating some docs

PR #368, opened 2018-11-28
<StarRating /> forward allowed props

### LINKED ISSUE #362: StarRating: Add support for className
### Desired behavior

Support for className so that component can be styled with styled-components.

danethurber:
fixes #362

### DIFF
--- a/packages/starrating/src/react/index.js
+++ b/packages/starrating/src/react/index.js
@@ -2,6 +2,7 @@ import * as glamor from 'glamor'
 import PropTypes from 'prop-types'
 import React from 'react'
 
+import filterReactProps from '@pluralsight/ps-design-system-filter-react-props'
 import { names as themeNames } from '@pluralsight/ps-design-system-theme/vars'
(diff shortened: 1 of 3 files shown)


packages/starrating/src/react/index.js
```diff
@@ -117,8 +118,11 @@ class StarRating extends React.PureComponent {
   render() {
     const { starCount, value } = this.props
 
+    const filteredProps = filterReactProps(this.props)
+    delete filteredProps.onChange
```
jaketrent: Is this `delete` pattern going to be oft-required?  I imagine that you're saying here "I use onChange to mean something else in another place in the app. I don't want the user to be able to attache it to this div".  

This seems like one of the cognitive challenges of letting "all" props in... we have to anticipate if *any* of those that the user passes in will cause a problem with our code.  

Seems like we'll have to do this often.  What's your take?

danethurber: hopefully it's not used to often but I don't know enough to say for sure. I get how this is a side effect of allowing all props...something to continue to consider...

PR #369, opened 2018-11-29
update form widths to work with Halo

### LINKED ISSUE #357: form: width of inputs in form component
Form controls do not fill the full width inside the Form.VerticalLayout any more. They should.
image

image

danethurber:
### What You're Solving

fixes #357

### Design Decisions

- `<Halo />` is using flex to expand to fit the field's width
- on radio and checkboxs that need to be inlined i added a new prop to the halo that changes the flex rules
- I updated the dropdown styles so that the div inside the halo will expand to fill the width

### How to Verify

- check in storybook


packages/halo/src/react/index.js
```diff
@@ -29,43 +30,36 @@ const styles = {
...
-    shape,
-    visible,
-    visibleOnFocus,
-    ...filteredProps
-  } = props
-
-  return <div {...styles.halo(props)} {...filteredProps} />
-}
+const Halo = props => (
+  <div {...styles.halo(props)} {...filterReactProps(props)} />
```
jaketrent: Is the pass-through of react props needed here?  I can't see the need yet.  What's your current thinking for where we're using this.  Also, since Halo isn't really a "user-facing" component, is there less of a need?

danethurber: it's not needed but i thought it looked cleaner so i used it.

i'm inclined to use the `filterReactProps` function often. I am running with the assumption it's safe to forward those react supported props. i could be persuaded to hold back if it's worrisome or we could make some guard rails to use

danethurber:
@jaketrent thanks for catching the switch. i'll update and go through and double check the other components

Issue #359, opened 2018-11-26
badge: Badges on docs site have lost their outer spacing. 


dborodyansky:
Badges on docs site have lost their outer spacing. 
Unexpected `css` attribute on the dom element for Badge.

![image](https://user-images.githubusercontent.com/3674406/49042550-4f649d00-f17d-11e8-95b1-0e8db0bb5b2e.png)

![image](https://user-images.githubusercontent.com/3674406/49042561-54c1e780-f17d-11e8-9ed1-8e6126cc1b73.png)

PR #372, opened 2018-11-29
Patch/doc labels

danethurber:
### What You're Solving

fixes: #359 

### Design Decisions

- I broke the css prop forwarding on the `<Badge />` so I fixed that(along with more prop forwarding). 
- I refactored how spacing is done on the site to use another component instead of css overriding

### DIFF
--- a/packages/badge/src/react/index.js
+++ b/packages/badge/src/react/index.js
@@ -2,6 +2,7 @@ import * as glamor from 'glamor'
 import React from 'react'
 import PropTypes from 'prop-types'
 
+import filterReactProps from '@pluralsight/ps-design-system-filter-react-props'
 import { withTheme } from '@pluralsight/ps-design-system-theme/react'
 import { names as themeNames } from '@pluralsight/ps-design-system-theme/vars'
 
@@ -27,10 +28,9 @@ const styles = {
   }
 }
 
-const Badge = props => {
-  const { appearance, color, themeName, ...filteredProps } = props
-  return <div {...styles.badge(props)} {...filteredProps} />
-}
+const Badge = props => (
+  <div {...styles.badge(props)} {...filterReactProps(props)} />
+)
 
 Badge.appearances = vars.appearances
 Badge.colors = vars.colors
--- a/packages/site/src/ui/page-heading.js
+++ b/packages/site/src/ui/page-heading.js
@@ -1,20 +1,25 @@
+import React from 'react'
+
 import Badge from '@pluralsight/ps-design-system-badge/react'
 import core from '@pluralsight/ps-design-system-core'
 
 import ChangeLog from './change-log'
 import Heading from './heading'
 
-const renderBeta = ({ beta }) =>
-  beta
-    ? <Badge
-        css={{ marginLeft: core.layout.spacingLarge }}
-        color={Badge.colors.blue}
-      >
-        Beta
-      </Badge>
-    : null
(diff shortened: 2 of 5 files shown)


packages/badge/src/react/index.js
```diff
@@ -27,10 +28,9 @@ const styles = {
   }
 }
 
-const Badge = props => {
-  const { appearance, color, themeName, ...filteredProps } = props
-  return <div {...styles.badge(props)} {...filteredProps} />
-}
+const Badge = props => (
+  <div {...styles.badge(props)} {...filterReactProps(props)} />
```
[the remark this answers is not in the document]
jaketrent: Also, will Badge allow `css` prop through any more?  (I hope so... I'd like to make this maybe not as standard as className and style support, but I do think that it offers a valuable option for writing things like descendent selectors inside css-in-js.)

packages/site/src/ui/page-heading.js
```diff
@@ -24,19 +29,21 @@ const ChangeLogContainer = props =>
...
-      </ChangeLogContainer>
-    : null
+  packageName ? (
+    <ChangeLogContainer>
+      <ChangeLog packageName={packageName} />
+    </ChangeLogContainer>
+  ) : null
 
-export default props =>
+export default props => (
```
jaketrent: I like the good ol days where prettier didn't want (parens) around these.

packages/site/src/ui/prop-types.js
```diff
@@ -20,12 +24,10 @@ const PropTypesRow = props => (
...
       <code className="nameCode">{props.row.name}</code>
       {props.row.required ? (
-        <Badge
-          appearance={Badge.appearances.stroke}
-          css={{ marginLeft: core.layout.spacingXXSmall }}
-        >
-          Required
-        </Badge>
+        <React.Fragment>
+          <BadgeSpacer />
```
jaketrent: Why the spacer approach instead of just adding the margin?  (The added margin has a cleaner dom.) I'm fine with the spacer, btw.

packages/site/src/ui/page-heading.js
danethurber: i can look into that!

packages/badge/src/react/index.js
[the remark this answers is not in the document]
danethurber: `appearance`, `color`, `themeName` are rejected in the filter. in this case they are given to the style function but not passed to the div. 

uhh...i'll have to look into the `css` prop.  that's not a react thing but a glamor thing correct. do you have an example

packages/site/src/ui/prop-types.js
danethurber: I guess i just like the way it reads and I prefer to not override the styles on the badge if possible

packages/site/src/ui/page-heading.js
jaketrent: No, wasn't a thing to look into, just something that prettier changed in the earlier days. I don't think there's anything to do here.

packages/badge/src/react/index.js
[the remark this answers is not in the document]
jaketrent: Umm, I thought it was a glamor thing, but maybe it was just glamorous??? https://glamorous.rocks/basics#core-concepts

An example of where I do this manually (apparently I knew it wasn't a glamor thing in my mind somewhere already): https://github.com/pluralsight/design-system/blob/master/packages/textinput/src/react/index.js#L56

PR #373, opened 2018-11-29
Patch/avatar alt

### LINKED ISSUE #365: avatar: add alt tags [a11y]
a360 id 233 (line 49) 

Let dev give prop:
alt={props.dev}

If none given:
aria-hidden={true}

And update docs

danethurber:
fixes #365

@jaketrent please review. I followed the rules in the issue but I think we'll probably want to add a label or something when initials are displayed

### DIFF
--- a/packages/avatar/package.json
+++ b/packages/avatar/package.json
@@ -21,7 +21,8 @@
   },
   "jest": {
     "setupFiles": [
-      "raf/polyfill"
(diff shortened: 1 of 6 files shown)


danethurber:
yes I was thinking an aria-label with the name. i'll add
major version sounds good to me - i'll ammend

PR #374, opened 2018-11-29
one command to test them all

### LINKED ISSUE #365: avatar: add alt tags [a11y]

### LINKED ISSUE #366: Running test script
## Running test script

Hi, I was trying to run an `npm run test` from the root folder but it (obviously?) breaks on the package-template's package.json because of the templating. 

I think that script should run all the components' test, am I right?

Should I run it in another way? I'm not sure if my question could be considered as an issue.

danethurber:
#366 spurred me into action

- get tests for each package passing. there were some failing snapshot i updated
- clean up root level test dependency and config(artifacts from a bygone time?)
- update root leave test script to run tests for each package in serial

Future considerations

- now is the time to setup a ci build on prs - run tests and whatnot
- there are still some root level build deps and config we should explore removing
- some snapshots were broken because of updates in related packages
  - do we mock related package dependency? or do we update downstream package snapshots
- snapshots could also be broken because i last updated them with the wrong related package builds
  - i.e. old(or new) versions in the dist folder
  - can we automate this? postpull hook to bootstrap/build? that seems slow....other options?
- still lots of copy/paste test config. we should explore a package for test config(like we do with build)

resolves #366


danethurber:
oops. i meant #366 - updated!

Issue #358, opened 2018-11-26
site: guideline (DO / DON'T) badge color rethink


dborodyansky:
We should discuss how to maintain the theming and still get our guideline block badges to look as they did previously

![image](https://user-images.githubusercontent.com/3674406/49042395-eda43300-f17c-11e8-8ee9-9c94b807b4c2.png)

dborodyansky:
Proposing this alternative to iron our the support for badge styles across themes.

![image](https://user-images.githubusercontent.com/3674406/49047736-3bc13280-f18d-11e8-9c5e-c978df7355e0.png)

Pros:
- Less limitation when stronger/subtler badges are appropriate in context.
- Alleviates the discrepancy of 'stroke' appearance not being promoted (discouraged) in light theme docs.
- Removes the possibility to use 'Stroke/Neutral' badge in the light theme, which isn't contrast compliant.

Cons:
- Renaming 'stroke' appearance to 'subtle' problematic? "Subtle" is an appearance type we use in Forms.

jaketrent:
Fixed.

PR #375, opened 2018-11-30
badge: light theme refactor

jaketrent:
### What You're Solving

- Addresses: https://github.com/pluralsight/design-system/issues/358

### Design Decisions

- Followed the pattern for generating selectors with helper functions.  Tried to simplify and make more readable.  At the very least, had to detach the "always compounds theme" logic.
- Renamed `theme` to `themeName` in a lot of places.  This has been the conventional name for the var that stores `light|dark`
- Also added filter-react-props package since we've been doing this as we touch packages.


packages/badge/src/css/index.js
```diff
@@ -3,8 +3,8 @@ import { transparentize } from 'polished'
 import core from '@pluralsight/ps-design-system-core'
 import { names as themeNames } from '@pluralsight/ps-design-system-theme/vars'
 
-import { buildCompoundClass } from '../js'
-import { appearances, colors } from '../vars'
+import { defaultWithColor, subtleThemeWithColor } from '../js'
```
danethurber: 👍 this reads better.

obviously it would be nicer to just be able to add classes individually in the component and then use combos of css selectors to get all the variation. i don't yet have a path forward there...

packages/badge/src/react/index.js
```diff
@@ -4,28 +4,19 @@ import PropTypes from 'prop-types'
...
-
-    const baseClass = '.psds-badge'
-    const compoundClass = buildCompoundClass({ appearance, color, theme })
-
-    return glamor.css(css[baseClass], css[compoundClass])
-  }
+  badge: props =>
+    glamor.css(
+      css['.psds-badge'],
+      props.appearance === appearances.default
```
danethurber: 👍

packages/badge/src/react/index.js
```diff
@@ -36,7 +27,7 @@ Badge.appearances = vars.appearances
 Badge.colors = vars.colors
 
 Badge.propTypes = {
-  appearance: appearancePropChecker,
+  appearance: PropTypes.oneOf(Object.values(vars.appearances)),
```
danethurber: so much better

Issue #356, opened 2018-11-26
tab: does not support `href` attrubute


jhewlett:
### Desired behavior

It would be great if the Tab component had support for linking to items via `href`. The way that we use the tabs, each tab represents a distinct page with its own route.

### Today's behavior

I have to use an `onClick` with a redirect. I can't do things like open in a new tab. Probably not great for screen readers, either.

### The Value Add

Things that behave like links should be links for accessibility.

### The Tradeoffs

There now needs to be logic to render tab items as either a button or a link depending on props provided.

danethurber:
Triage: we think this is a good idea to expore

Issue #376, opened 2018-12-03
tab: Anchors inside Tab.ListItem don't work in Firefox


willsoto:
_This is kinda both a bug and a feature request, since I think implementing the feature will fix the bug._

### Expected behavior

Using an anchor for a tab works as expected.

### Actual behavior

Anchor tags inside of `button` don't work in Firefox (See [this SO post](https://stackoverflow.com/a/16280685))

[Source](https://github.com/ps-dev/Projects-UI/blob/c0ad605a1785717c4720c92eb8d1bfcf26f29650/src/detail/tabs/index.js#L112-L116)
[Link to GIF](https://drive.google.com/open?id=1_oUunjw5JpCzClYVAYDF6FNSl40zbVFD) (It's big, sorry). You can see an example of the correct behavior in Chrome (URL changes when tab is clicked so users don't lose their position)

### Steps to repro

1. Navigate to any Project detail page (eg [here](https://app.pluralsight.com/projects/add-features-with-forms-to-a-vue-js-reading-list-application). 
2. Click around on tabs in Chrome
3. Click around on tabs in Firefox

### Related packages

- button@10.8.3 (maybe? Would have to check implementation of Tab)

### Environment

- OS - MacOS 10.12
- Browser version - Chrome v71 (works) Firefox v63 (doesn't work)

### Desired behavior

Add ability to replace `Tab.ListItem` with custom markup

### Today's behavior

`Tab.ListItem` does not allow custom markup

### The Value Add

Related to #248. Similar value add

### The Tradeoffs

Slightly more complex implementation? But ultimately worth it in my opinion.

willsoto:
Just saw #356 - sorry for the dupe...it's kinda related I think, but it would be ideal if we could pass a custom component imo (since things like React Router use `Link` and not `href`)

jaketrent:
Funny, perhaps the same day this was filed here, almost same issue was filed on react-router site: https://github.com/ReactTraining/react-router/issues/6499

He created the same kind of markup that's in the source link in this issue above.

jaketrent:
With the completion of #356, `href` are now support on `Tab.ListItem`, making semantic markup possible for links from tabs.  

I've also verified that react-router has a story that satisfies its own routing needs while keeping the semantic markup.  This can be accomplished using its [`withRouter` HoC](https://reacttraining.com/react-router/web/api/withRouter), as in:

```js
import { withRouter } from "react-router-dom"
[...]

PR #388, opened 2018-12-07
tab: support href, amongst other modernization refactors

jaketrent:
Fixes: #356

### DIFF
--- a/packages/tab/src/react/__stories__/index.story.js
+++ b/packages/tab/src/react/__stories__/index.story.js
@@ -1,11 +1,9 @@
 import addons from '@storybook/addons'
-import core from '@pluralsight/ps-design-system-core'
 import React from 'react'
 import { storiesOf } from '@storybook/react'
-import Theme from '@pluralsight/ps-design-system-theme/react'
 import themeDecorator from '@pluralsight/ps-design-system-storybook-addon-theme'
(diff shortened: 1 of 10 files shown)


jaketrent:
Once we review and merge this, I will go on to ensure this use case is supported as well: https://github.com/pluralsight/design-system/issues/376

packages/tab/src/react/__stories__/index.story.js
```diff
@@ -1,11 +1,9 @@
 import addons from '@storybook/addons'
-import core from '@pluralsight/ps-design-system-core'
 import React from 'react'
 import { storiesOf } from '@storybook/react'
-import Theme from '@pluralsight/ps-design-system-theme/react'
 import themeDecorator from '@pluralsight/ps-design-system-storybook-addon-theme'
 
-import Tab from '../react'
+import Tab from '../index.js'
```
danethurber: you could just do `..` but it's all the same thing in the end

jaketrent: Yeah, we should chat about this at some point.  It's been a style thing for me for a long time to do exactly as you say (".."), but there are new developments about browser native modules and node modules that make me wonder if we should start doing "../index.js".

PR #389, opened 2018-12-10
row: refactor and  light theme

### LINKED ISSUE #334: row: add light theme

danethurber:
resolves #334 
fixes: #390 

Tasks:

- [x] remove glamorous
- [x] cleanup proptypes
- [x] add light theme support


packages/row/src/css/index.js
```diff
@@ -67,7 +55,12 @@ export default {
     background: transparentize(0.5, core.colors.black),
     transition: `opacity ${core.motion.speedNormal}`,
     pointerEvents: 'none',
-    opacity: 0
+    opacity: 0,
+    color: core.colors.white,
```
danethurber: this might be problematic. i'm using it to set the icon color - is this a breaking change?

jaketrent: Which icon?  

If it wasn't being set before, perhaps people are setting it themselves.  And if they are, where would they be doing that?  And once we determine that, would this override what they're doing?  And if so, is that bad, since maybe we should have been doing it all along (if we aren't already, still not sure)?

So, I'm still not sure. :)

danethurber: the playicon that was in the story. if people are setting it themselves i don't think this will break anything since that would most likely be more specific. and yes i think this way is better

packages/row/src/react/__stories__/index.story.js
```diff
@@ -1,69 +1,66 @@
 import { action } from '@storybook/addon-actions'
-import addons from '@storybook/addons'
+import { storiesOf } from '@storybook/react'
+import React from 'react'
+
```
jaketrent: What is the import convention that you're showing here?  Thusfar, my import convention in this project is generally:

named imports, alpha order
blank line
local imports, alpha order
blank line
code...

packages/row/src/react/__stories__/index.story.js
```diff
@@ -91,23 +96,19 @@ Object.keys(Row.sizes).forEach(size =>
...
-        <Row.Image src="http://lorempixel.com/output/technics-q-c-680-320-6.jpg" />
-      }
+      image={<Row.Image src={getImgSrc()} />}
     />
   ))
 )
 
 const progressStory = storiesOf('progress', module)
-;[0, 66.6667, 100].forEach(progress =>
+Array.from([0, 66.6667, 100]).forEach(progress =>
```
jaketrent: why copy the array here?

packages/row/index.js
```diff
@@ -1,3 +1,5 @@
-const React = require('./dist/react')
+const css = require('./css')
+const react = require('./react')
+const vars = require('./vars')
 
-module.exports = { react: React }
+module.exports = { css: css, react: react, vars: vars }
```
jaketrent: oooh, nice catch; I thought we got all these once upon a time.

packages/row/src/css/index.js
jaketrent: If it's the playicon, I'd leave out the style.  The Card spec doesn't define what should be in the overlay where the play icon often is.

packages/row/src/react/__specs__/element-of-type.spec.js
```diff
@@ -0,0 +1,61 @@
+import React from 'react'
```
jaketrent: Awesome!  Fancy!

packages/row/src/react/__specs__/element-of-type.spec.js
```diff
@@ -0,0 +1,61 @@
...
+    expect(result).toBeInstanceOf(Error)
+    expect(result.message).toMatchInlineSnapshot(
+      `"Invalid prop \`testProp\` with value \`\\"string - not an element\\"\` supplied to MockComponent, expected element of type \`Uno\`"`
+    )
+  })
+
+  it('should warn when expected element is of the wrong type', () => {
+    const fn = elementOfType(Uno)
+
+    const props = { [PROP_NAME]: <Dos /> }
```
jaketrent: MS-Dos

packages/row/src/react/element-of-type.js
```diff
@@ -0,0 +1,78 @@
+// NOTE: largely taken from `prop-types` and `react-element-proptypes`
```
jaketrent: What part is custom to us?  Or, would it be possible to just use react-element-proptypes without needing to maintain this custom code?

packages/row/src/react/index.js
```diff
@@ -331,33 +333,33 @@ const Row = ({ title, ...props }) => (
...
-  // fullOverlayVisible: PropTypes.bool,
-  // image: validateComponentType([Image, ImageLink]),
-  // metadata1: PropTypes.arrayOf(validateComponentType([Text, TextLink])),
-  // metadata2: PropTypes.arrayOf(validateComponentType([Text, TextLink])),
-  // progress: PropTypes.number,
-  // title: PropTypes.oneOfType([Text, RowComponent.TextLink]).isRequired,
-  // size: PropTypes.oneOf(Object.keys(vars.sizes))
-
-  actionBar: PropTypes.arrayOf(PropTypes.element),
+  actionBar: PropTypes.arrayOf(elementOfType(ActionBarAction)),
```
jaketrent: So cool!  Way to go on getting this to work!

packages/row/src/css/index.js
```diff
@@ -101,13 +104,18 @@ export default {
     lineHeight: 0,
     cursor: 'pointer',
     border: 'none',
-    color: core.colors.gray02,
     background: 'none',
-    transition: `all ${core.motion.speedNormal}`,
+    transition: `all ${core.motion.speedNormal}`
+  },
+  [`.psds-row__action-bar__button--theme-${themeNames.dark}`]: {
```
jaketrent: We have traditionally tried to refer to `defaultTheme` from the `theme` package instead of `themeNames.dark`.  I frankly can't think of the reasoning that went into this decision in the first place -- it's just something I've tried to be consistent on since.  

Are there any pros/cons that would push us one way or the other besides internal consistency that occur to you?

packages/row/src/css/index.js
```diff
@@ -101,13 +104,18 @@ export default {
...
+    color: core.colors.gray02,
 
-    '&:active,&:hover': {
-      color: core.colors.white
-    }
+    '&:active, &:hover': { color: core.colors.white }
+  },
+  [`.psds-row__action-bar__button--theme-${themeNames.light}`]: {
```
jaketrent: The theme modifier for selectors has traditionally been something more like:

```
`.psds-row__action-bar__button.psds-theme--${themeNames.light}`
```

Same for line 110 ^.

packages/row/src/css/index.js
```diff
@@ -189,6 +202,9 @@ export default {
     }
   },
 
+  // __text
+  '.psds-row__text': {},
```
jaketrent: Do we want/need this if it's empty?

packages/row/src/react/index.js
```diff
@@ -31,28 +31,29 @@ const styles = {
...
     }),
-  actionBarAction: ({ disabled }) =>
+  actionBarAction: ({ disabled, themeName }) =>
     glamor.css({
       ...css['.psds-row__action-bar__button'],
+      ...css[`.psds-row__action-bar__button--theme-${themeName}`],
       ...(disabled && css['.psds-row__action-bar__button--disabled'])
     }),
-  fullOverlay: ({ fullOverlayVisible: visible, isFocused }) =>
+  fullOverlay: ({ fullOverlayVisible: visible, isFocused, themeName }) =>
```
jaketrent: themeName appears unused here.

packages/row/src/react/index.js
```diff
@@ -91,16 +97,15 @@ const ActionBar = props => (
   <div {...styles.actionBar(props)} {...filterReactProps(props)} />
 )
 
-const ActionBarAction = props => {
+const ActionBarAction = withTheme(({ icon, ...props }) => {
```
jaketrent: Why destructure out `icon` out?  Seems like `filterReactProps` would have taken it out, no?

packages/row/.storybook/config.js
```diff
@@ -1,9 +1,16 @@
...
+import { addDecorator, configure } from '@storybook/react'
 
-import { configure } from '@storybook/react';
+import centerDecorator from '@pluralsight/ps-design-system-storybook-addon-center'
+import themeDecorator from '@pluralsight/ps-design-system-storybook-addon-theme'
 
-function loadStories() {
-  require('../stories');
+addDecorator(fn => <div style={{ maxWidth: 'calc(100vw - 40px)' }}>{fn()}</div>)
+addDecorator(centerDecorator)
```
jaketrent: I think it might be best not to use this decorator here.  It's interfering with the full-width nature of the row component.  I can't determine what styles are constraining it after a quick look, but this is what I see:

With center:

screen shot 2018-12-11 at 9 49 16 am

As before, without:

screen shot 2018-12-11 at 9 48 17 am

It made me concerned that the full-width flexibility of the Row may have been broken (which it was not, as I discovered when I disabled this decorator).  Perhaps it's better to go without.

danethurber: good point - i'll do something different

packages/row/src/css/index.js
danethurber: will do

packages/row/src/css/index.js
danethurber: well....i've probably broken that convention a couple times :/

I don't have a good sense either way. i'll update to use `defaultTheme` and match convention

packages/row/src/css/index.js
danethurber: good catch. i'll fix

packages/row/src/css/index.js
danethurber: no i'll just remove it. i think i was using it but didn't end up needing it

packages/row/src/react/element-of-type.js
danethurber: I opted to copy and modify because `react-element-proptypes` hasn't had any contribution for 2 years. also there are some things that looked to have changed in how react does things so I went with writing our own for now

packages/row/src/react/index.js
danethurber: i'll fix

packages/row/src/react/index.js
danethurber: can't remember. i'll double check

packages/row/src/react/__stories__/index.story.js
danethurber: so i guess i've been organizing DS package separately because they kind of feel like a type of local imports...not sure when i started doing that :/

packages/row/src/react/__stories__/index.story.js
danethurber: i didn't like the leading `;`...thoughts?

jaketrent: I love the leading semicolon.  Semicolons are not needed in most places, but they are here.  So the "love" comes from using it in the limited places it is needed.

packages/row/src/react/__stories__/index.story.js
jaketrent: I'm open to that.  I guess we need to decide what to do here.

packages/row/src/react/__stories__/index.story.js
danethurber: :) i'll revert. i mean when else are we going to get the chance to use a semi

packages/row/src/react/index.js
danethurber: turns out it is a html attr -> https://github.com/jackyho112/react-html-attributes/blob/master/src/react-html-attributes.json#L37

so i guess that's while i'm filtering it

danethurber:
@jaketrent will you give this another review please :)

PR #392, opened 2018-12-11
<Icon /> refactor and additional prop support

### LINKED ISSUE #256: site: add docs for "does it belong in the DS?"

danethurber:
- remove glamorous from Icon
- add mock for storybook theme decorator for smaller storyshots
- ~simplify svg > React transpiling~
- support prop overrides (resolves #256 )


packages/icon/src/react/__specs__/__snapshots__/storyshots.spec.js.snap
```diff
@@ -2,1891 +2,1514 @@
```
danethurber: these snapshots should be really close. i'm mocking the storybook theme decorator so you'll see that an extra div is removed

packages/icon/src/react/__specs__/index.spec.js
```diff
@@ -0,0 +1,26 @@
+import React from 'react'
```
danethurber: i added these tests...not sure they're that valuable. i wanted to make sure each icon still rendered without errors....

packages/site/pages/components/icon.js
```diff
@@ -79,7 +80,7 @@ export default withServerProps(_ => (
             <span>
               <code>Icon.ids</code> enum
             </span>,
-            true,
+            null,
```
danethurber: id isn't required...because you can pass in a custom svg

packages/site/pages/components/icon.js
```diff
@@ -91,6 +92,15 @@ export default withServerProps(_ => (
             null,
             null,
             'all flat Core colors'
+          ]),
```
danethurber: this was missing. i added it

packages/icon/src/react/index.js
```diff
@@ -1,41 +1,51 @@
...
-const rmNonHtmlProps = props => {
-  const { id, ...rest } = props
-  return rest
+import fileNames from './file-names'
+
+const IconNotFound = () => null
+
+export const loadIconByID = id => {
+  try {
+    return require(`./icons/${fileNames[id]}`).default
```
danethurber: i'm having second thoughts about this. i'm not sure that it is going to work...

packages/storybook-addon-theme/__mocks__/index.js
```diff
@@ -0,0 +1,6 @@
+var React = require('react')
+
+const MockTheme = props => React.createElement(React.Fragment, props)
```
jaketrent: what is your thinking on why it's desirable to mock the theme in stories?

packages/icon/src/react/index.js
```diff
@@ -1,40 +1,40 @@
...
+        ? css[`.psds-icon--color-${color} > svg`]
+        : css['.psds-icon > svg'],
 
-const Icon = props => {
-  return (
-    <IconContainer {...rmNonHtmlProps(props)}>
-      {props.children || icons[props.id](React)}
-    </IconContainer>
-  )
+      // NOTE: this is a holdover from the old glamorous api,
```
jaketrent: still a desire to remove?  Or did you find a usecase where it's still required to do what most users need to do with icon?

packages/icon/README.md
```diff
@@ -1,17 +1,16 @@
...
 
 ## To Add New Icon Svgs
 
-1. Name icons *.icon.svg
+1. Name icons \*.icon.svg
 1. Put in the `src/svg` directory
 1. Add `role="img" and aria-label="<iconId here> icon"` to the root `svg` element of each icon
 1. Manually refactor the svg to not use xlink/defs.
-1. Run `npm run build` in the `packages/icon` directory (new `*.svg.dist.js` files will generate new icon ids automatically)
+1. Run `npm run build` in the `packages/icon` directory (new `*.svg.js` files will generate new icon ids automatically)
```
jaketrent: Not a huge deal, so please don't revert, but I liked the `dist` inclusion in the file name so that it was clear it was from the build and not src.  What was your thinking on the name change?

packages/icon/src/react/index.js
```diff
@@ -4,11 +4,19 @@ import React from 'react'
...
 import filterReactProps from '@pluralsight/ps-design-system-filter-react-props'
 
-import icons from '../js/icons'
-
 import css from '../css'
 import * as vars from '../vars'
 
+import fileNames from './file-names'
+
+export const loadIconByID = id => {
```
jaketrent: plz2camelize "ID" -> "Id"

packages/icon/src/vars/index.js
```diff
@@ -1,8 +1,8 @@
 import core from '@pluralsight/ps-design-system-core'
 
-import icons from '../js/icons'
+import fileNames from '../react/file-names'
```
jaketrent: I feel like this is a good reason for file-names to remain in `/js`...  `vars` are not going to be framework agnostic, as they're meant to be (so can be reused across side-by-side js framework tech and the css, if they are pulling in react dependencies.

Perhaps what file-names includes now needs to split -- the general use bits for the file-name collection could go back into `js/` and the react component template could stay in the new `react/` location.

packages/icon/src/react/index.js
```diff
@@ -30,11 +32,11 @@ const style = {
 }
 
 const Icon = ({ id, ...props }) => {
-  const Comp = props.children || loadIconByID(id)
+  const Comp = loadIconByID(id)
```
jaketrent: by inlining line 35 after the `||` on line 39, you could make that load lazy

packages/icon/src/react/icon-transformer.js
```diff
@@ -1,5 +1,6 @@
 // @preval
 
+const babel = require('babel-core')
```
jaketrent: I like the rename of the file.

packages/icon/src/react/icon-transformer.js
```diff
@@ -12,6 +13,15 @@ const pascalize = str => {
   return trimmed.substr(0, 1).toUpperCase() + camelize(trimmed.substr(1))
 }
 
+const insertAfterChar = (str, char, content) => {
+  const charIndex = str.indexOf(char)
+
+  if (charIndex === -1) return str
+
+  const offset = charIndex + 1
+  return [str.slice(0, offset), content, str.slice(offset)].join('')
```
jaketrent: making an array to then concat strings could be simplified by just concating the strings.

packages/icon/src/react/icon-transformer.js
```diff
@@ -50,6 +67,9 @@ module.exports = fileNames
       buildComponentFile(filePath, { componentName })
     )
 
-    acc[id] = outputFileName
+    acc[id] = require(`./icons/${outputFileName}`)
+
     return acc
   }, {})
+
+module.exports = { ...temp }
```
jaketrent: what is the reasoning for exporting a shallow clone of temp vs just keeping what was on old line 37?

packages/storybook-addon-theme/__mocks__/index.js
danethurber: I like the snapshots to be smaller and more inline with what is actually being tested. having the theme components in each snapshot adds a lot of extra levels to sort through

packages/icon/src/react/index.js
danethurber: no i removed it

packages/icon/src/react/icon-transformer.js
danethurber: left over debugging. will remove

packages/icon/src/react/icon-transformer.js
danethurber: good catch. fixed

packages/icon/src/vars/index.js
danethurber: i see you point. i'll move it back to `js` for now. we can readdress a better approach in the future

packages/icon/README.md
danethurber: i actually like it better that way now that i think of it. i changed it back

PR #395, opened 2018-12-12
migrate to jest projects

danethurber:
jest has a feature to support monorepos called projects. i've setup the configuration in a way that we can run jest from the root to run all tests, run jest from the root for a single project or run the `test` script in an individual package to just run the package's tests. examples:

_from the project root we can do_

`jest` ->  runs all the tests
`jest --project package/avatar` -> runs just the avatar tests
`jest --project package/avatar package/badge` -> runs avatar and badge tests

_from the avatar package_

`npm test`  -> runs just the avatar tests

### Perceived Benefits

- one global configuration to maintain
- each package can still update their own config
- polyfills and enzyme setup are in one place
- can add custom jest matchers/plugins in one place. i'm planning to add `jest-axe` at some point
- i plan to add the enzyme snapshot serializer to get nice jest snapshots of react elements
- faster installs - jest(and related) are installed in the root
- faster runs(local and in ci)
- project wide coverage reports
- if everything is configured in a similar way there is no context switching between packages. ie...i know i'll have this matcher or that serializer installed in every test

### Possible Downsides

- everything uses the same versions and configs. this could make upgrading more painful.

### Outstanding tasks:

- [x] update the remaining packages to use the shared config
- [x] update the `component:create` script with new config
- [x] add jest typeahead for easy searching in the cli gui
- [x] upgrade root `npm test` to use single jest instance instead of `lerna run`


packages/actionmenu/package.json
```diff
@@ -15,15 +15,10 @@
     "build:watch": "npm run build:js -- --watch",
     "prepublish": "npm run build",
     "storybook": "start-storybook -p 6006",
-    "test": "jest",
+    "test": "../../node_modules/.bin/jest",
```
danethurber: we could simplify this if we used something like https://github.com/nicojs/node-link-parent-bin

jaketrent:
Questions that come to mind:

- I'd like to know how these two benefits co-exist: "one global configuration to maintain", "each package can still update their own config"
- Can a dev come in and dev a single component without knowing/caring about the other unrelated packages?  
- Can we have a totally different one-off test runner in the future for a single package?

package.json
```diff
@@ -75,6 +75,8 @@
     "babel-preset-stage-2": "^6.24.1",
```
jaketrent: I assume you'll want to get the jest commands in here to replace the lerna run test commands?

jaketrent: for some reason, typing `jest` in test, takes me to the `/jest` dir without running anything.  Maybe this is zsh being "helpful"...

danethurber:
this shaved ~2min off the build. not as much as I hope. bootstrapping is still the longest part

jaketrent:
oops, I merged.  old habits... hopefully we were ready for that.  :) anyway, pr approved!

PR #421, opened 2018-12-21
New thumb icons

jaketrent:
With some build updates - to allow fill-rule svg attr; to whitelist props based on svg tagName
And some icon package test updates - point to dist; preval done at build; required for functioning component

Hh, a quick request on this one: Dane, if you end up working on babel@7 upgrade, please try to merge this first since there was some trouble around making sure that babel/build/tests were working nicely here.  I'd like to get this one resolved on that baseline before that upgrade, if possible.


jaketrent:
Another issue related: Found out that the 9.0.2 version of icon, published for some other thing on 20 Dec, included these new thumb icons.  

How?  I believe it is because the publish is done locally and the *.dist.js were in the src directory and got babeled into dist.  I solved this with a hotfix to master (52028b4c) and rerelease of icon@9.0.3.

PR #258, opened 2018-12-28
Remove glamorous

danethurber:
TASKS:

- [x] actionmenu
- [x] breadcrumb
- [x] drawer
- [x] link
- [x] tab
- [x] text
- [x] viewtoggle
- [x] root level babel config
- [x] audit related snapshot changes
- [x] update docs


danethurber:
<table>
  <thead>
    <tr>
      <th width="50"></th>
      <th width="100%" data-danger-table="true">Warnings</th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>:exclamation: Please assign a teammate to review your PR.</td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against be97291693787b8174b9f603c1745cfce4f73da6
</p>

packages/viewtoggle/src/react/index.js
```diff
@@ -41,17 +57,28 @@ class Option extends React.Component {
...
 }
 
-const List = glamorous.div(css['.psds-viewtoggle'])
+const List = props => (
+  <div {...styles.list(props)} {...filterReactProps(props)} />
+)
 
-const ActivePillBg = glamorous.div(css['.psds-viewtoggle__option-bg'])
+const ActivePillBg = props => {
+  return <div {...styles.activePillBg(props)} {...filterReactProps(props)} />
```
jaketrent: Unless the component is separately exposed to the user, I don't think it makes sense to pass all props to it as in `{...filterReactProps(props)}`.  

In this case, I think that means that only `<ViewToggle />` and `<ViewToggle.Option />` should receive that ability.

Components that are not exposed publicly are internal details, and a user really doesn't know what to expect when sending in html-like props in those cases.

There's also the issue of duplicating props across multiple elements, as in passing `className` to internal components.

In this case, however, the main reason I wouldn't do it is that it's not needed.  The internal use of `<ActivePillBg />` and `<PillBgSpacer />` don't pass on the user-provided.  Instead, we're filtering what we as DS devs pass as props, which we should probably just trust (and verify at dev time).  It seems like unneeded work when looking at it at the moment.

.babelrc
```diff
@@ -1,9 +1,5 @@
 {
-  "plugins": [
-    "syntax-export-extensions",
-    "transform-export-extensions",
-    "glamorous-displayname"
-  ],
+  "plugins": ["syntax-export-extensions", "transform-export-extensions"],
```
jaketrent: This is the babel plugin that supports `export from` syntax. (from earlier convo) I would love to remove it.  The more I babel, the less I want to.

PR #433, opened 2019-01-11
Dropdown/ActionMenu: keyboard-specific focus style

### LINKED ISSUE #363: ActionMenu: Add border for accessibility
### Expected behavior

Each Button for an ActionMenu inside a DropDown should have the border (blue in Chrome) for accessibility.

### Actual behavior

There is a very faint gray and a slightly darker gray that distinguish the two now. 
screen shot 2018-11-27 at 3 56 21 pm

jaketrent:
Fixes #363

### DIFF
--- a/packages/actionmenu/src/css/index.js
+++ b/packages/actionmenu/src/css/index.js
@@ -1,6 +1,6 @@
 import core from '@pluralsight/ps-design-system-core'
 
-import * as vars from '../vars'
+import * as vars from '../vars/index.js'
 
 const menuOriginRight = {
   transform: `translateX(${core.layout.spacingXSmall})`
@@ -91,6 +91,11 @@ export default {
     outline: 'none',
     color: core.colors.gray05
   },
+  '.psds-actionmenu__item--focus-keyboard': {
+    background: core.colors.blue,
+    outline: 'none',
+    color: core.colors.white
+  },
 
   // __item--link
   '.psds-actionmenu__item--link': {
@@ -130,8 +135,15 @@ export default {
 
   // __item__arrow
   '.psds-actionmenu__item__arrow': {
+    color: '#888',
     marginLeft: 'auto',
     paddingLeft: core.layout.spacingXSmall
+  },
+  '.psds-actionmenu__item__arrow--focus-keyboard': {
+    color: core.colors.white
+  },
+  '.psds-actionmenu__item__arrow__svg': {
+    fill: 'currentColor'
   },
 
   // __overlay
--- a/packages/actionmenu/src/react/__stories__/index.story.js
+++ b/packages/actionmenu/src/react/__stories__/index.story.js
@@ -6,54 +6,7 @@ import React from 'react'
 import Icon from '@pluralsight/ps-design-system-icon/react'
 import themeDecorator from '@pluralsight/ps-design-system-storybook-addon-theme'
 
-import ActionMenu from '..'
-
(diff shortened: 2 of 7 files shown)


packages/actionmenu/src/css/index.js
```diff
@@ -1,6 +1,6 @@
 import core from '@pluralsight/ps-design-system-core'
 
-import * as vars from '../vars'
+import * as vars from '../vars/index.js'
```
jaketrent: This might be a bigger convo, not sure.  We can have any time. https://docs.google.com/document/d/1ek9NzuJdQTWBzAwgQwlTQRUdU4ZS-hCg2qQiGU4VSQk/edit

packages/actionmenu/src/react/__stories__/index.story.js
```diff
@@ -6,54 +6,7 @@ import React from 'react'
 import Icon from '@pluralsight/ps-design-system-icon/react'
 import themeDecorator from '@pluralsight/ps-design-system-storybook-addon-theme'
 
-import ActionMenu from '..'
-
-// TODO: snapshot test generator can't handle.
```
jaketrent: There are a couple TODO removals here.  Just removing the backlog that I'm doing nothing about.

packages/dropdown/src/react/index.js
```diff
@@ -232,27 +236,27 @@ class Dropdown extends React.Component {
...
-                      ? this.field.getBoundingClientRect().width
-                      : 'auto'
-                  }
-                })}
-              </div>
-            )}
+          {props.menu && state.isOpen && (
+            <div {...styles.menu(allProps)}>
+              {React.cloneElement(props.menu, {
+                isKeyboarding: state.isKeyboarding,
```
jaketrent: Mostly a prettier change; this is the important line.

packages/dropdown/src/react/index.js
```diff
@@ -109,6 +110,9 @@ class Dropdown extends React.Component {
     if (typeof this.props.onBlur === 'function') this.props.onBlur(evt)
   }
   handleKeyDown(evt) {
+    if (evt.key === 'ArrowDown' || evt.key === ' ' || evt.key === 'Enter') {
```
jaketrent: All 3 because this is where I want to discover "is a keyboarder!" but the space and enter keys also trigger onClick

packages/actionmenu/src/css/index.js
danethurber: interesting. thanks for bring it up

something like https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/extensions.md might help out

packages/dropdown/src/react/index.js
danethurber: what does `evt.key === ' '` check for? is it the space key and is there no `Space` string? asking for a friend

packages/actionmenu/src/css/index.js
jaketrent: Yes, that type of eslint plugin might help.  Maybe we leave it on the retro docket for later today.  Since you've approved the PR in total, I take that as ok to move forward with what's committed.  If we have a later convo and change our mind about this generally, I'm happy to change this.

packages/dropdown/src/react/index.js
jaketrent: Yes, it's for space.  In my experimenting there was no `Space` string, just 1 char space string.

Issue #446, opened 2019-01-15
site: polyfill needed for ie11


danethurber:
we get an error that `Object.values` can't be found. there might be others

![image_720](https://user-images.githubusercontent.com/226356/51207462-14066500-18c8-11e9-975d-43064126b905.png)

jaketrent:
or could: https://github.com/pluralsight/design-system/pull/450/files#diff-ea2925d9b7851c0826722207d9f3a68bR84

jaketrent:
timebox polyfill add to component

or don't use
or create util
or rely on BCs having polyfill

PR #452, opened 2019-01-17
fix(site): polyfill es6 and es7 features

danethurber:
fixes: #446

### DIFF
--- a/packages/site/src/polyfills.js
+++ b/packages/site/src/polyfills.js
@@ -1,1 +1,2 @@
-import 'core-js/fn/object/assign'
+import 'core-js/es6'
+import 'core-js/es7'
--- a/packages/site/pages/install.js
+++ b/packages/site/pages/install.js
@@ -27,6 +27,12 @@ npm install @pluralsight/ps-design-system-normalize
 
 Include this vanilla CSS in your application in a method appropriate for your project. For usage options, follow the [Core Usage docs](/core/usage) patterns.
 
+## JavaScript Environment Requirements
+
+Many components in the the Design System depend on features in more modern versions of ecmascript(es6 and es7). Older browsers and devices that we support do not provide these natively (e.g. IE 11) so you'll need to include a global polyfill in your bundled application, such as \`core-js\` or \`babel-polyfill\`.
+
+To explore features by version you can visit the [ECMAScript Compatibitily Table](https://kangax.github.io/compat-table/es6/).
+
 ## Use Core
 
 ### Step 1: Install Core


packages/site/src/polyfills.js
```diff
@@ -1 +1,2 @@
-import 'core-js/fn/object/assign'
+import 'core-js/es6'
+import 'core-js/es7'
```
danethurber: es7 has `Object.values` as well as `Array.includes` which we use

jaketrent: I like the ease of this., but I'm still concerned that it hides the issue from us a bit and from our users.

From us because we "deploy the DS to an app, the site, and the components work".  We think, "well, the next BC that imports these components will have it work for them as well."

From our users because they come to the site, see it working in IE, go import it in their site, it *may* not work, and they wonder what's up.

To expose this a bit more to ourselves and our users, I think we need to either need to: 

- Track what's polyfilled on the site (our only IE11, team-owned runtime) and document in the dev docs so users can make sure they polyfill one way or the other. // pro: if they have a mega polyfill, they're covered, if they're trying to play it light, they'll see the features that they need polyfilled and grab those as one-offs.
- Or build the polyfills into the components that need them for the IE11 (and other) runtimes that we support. // pro: users of our components don't have to worry, it's taken care of;  con: potential bloat from multiple copies of the same polyfill (no tree shaking without es modules)

What are your thoughts on this?  Maybe something to CPT?  Or maybe add this to a Dash search term (core-js/babel-polyfill)?

danethurber: yep, i opened this as a conversation starter. and I have the same concerns and have the same solutions you have outlined.

I think up until this point we've probably been assuming that consumers have something like `babel-polyfill` and that has gotten us this far. adding that expectation to integration docs is definitely something we could do.

we have bundled in other polyfills(resize-observer, focus-within) but I think that's because those are outside of what `core-js`/`babel` provide. would including `babel-polyfill` in our bundle cause large file sizes? could it get bundled twice?

I lean more towards making `babel-polyfill` a requirement by the consuming dev team. they can include the entire polyfill(probably pretty common) or setup webpack/babel `preset-env` to include only the necessary polyfills.

jaketrent: Yes, I too believe it's the general use of babel-polyfill/equivalent in projects that has probably allowed us to get this far w/o hearing about it much.  Given that good piece of likely evidence, explicitly requiring users to use a polyfill (via docs) seems reasonable.

Yes, including the whole babel-polyfill will probably make our components unreasonably large.  That said, I haven't attempted it.  We could test that expectation.  And yes, it could definitely get bundled twice (or MORE! -- `n + 1` DS  omponents + their own BC) because we lack an es module build.

We already use `babel-preset-env`, of course.  Babel is the transpiling, converting our new age syntax to browser-ready syntax.  But separate polyfills per browser-absent *features* are still required.  Babel does not add those, or at least preset-env does not.

danethurber:
TODO: update docs around polyfills. required consumers to have polyfills that support up to es7 features. link to docs about what es7 support is needed

danethurber:
@jaketrent please review

PR #454, opened 2019-01-17
Bug/tab focus safari

jaketrent:
Lots of style re-treatment to try to simplify reading it.  Turns out it wasn't required.  Still nice to have, probably.


packages/tab/src/react/list-item.js
```diff
@@ -99,16 +38,18 @@ const ListItem = (props, context) => {
       },
       { tagName }
     ),
-    <div {...styles.textWidth(props)}>
+    <div {...styles.textWidth(props)} tabIndex="-1">
```
danethurber: this is causing related snapshots to fail fyi

PR #456, opened 2019-01-22
Row.Action ref support

### LINKED ISSUE #451: ActionMenu in a Row.Action
https://pluralsight.slack.com/archives/C70HPSJPP/p1547752073034000

we need a way to have an `ActionMenu` shown when `Row.Action` is clicked. 

- the consumer could put a ref on the `Row.Action` and get the BoundingClientRec and position the menu themselves
  - error is currently thrown -> `Warning: Function components cannot be given refs. Attempts to access this ref will fail.`
  - could convert to a class component
  - could upgrade to React@16.3 and use `forwardRef`

- another option is adding support to rendering an `ActionMenu` as a child of the `Row.Action`

danethurber:
i updated the Row.Action to be a class component to resolve #451

### DIFF
--- a/packages/row/src/react/__specs__/__snapshots__/storyshots.spec.js.snap
+++ b/packages/row/src/react/__specs__/__snapshots__/storyshots.spec.js.snap
@@ -218,57 +218,6 @@ exports[`Storyshots actionBar one action 1`] = `
 </div>
 `;
 
-exports[`Storyshots actionBar with a ref 1`] = `
-<div
-  data-css-claeju=""
->
-  <div
-    style={
-      Object {
-        "margin": "24px auto",
-        "maxWidth": "calc(100vw - 64px)",
-      }
-    }
-  >
-    <div
-      data-css-vl7fes=""
-    >
-      <div
-        data-css-1tiyrld=""
-      >
-        <div
-          data-css-1jlgfae=""
-        >
-          I support refs
-        </div>
-      </div>
-      <div
-        data-css-okadky=""
-      >
-        <button
-          data-css-2qvjr2=""
-        >
-          <div
-            data-css-otejxm=""
-          >
-            <svg
-              aria-label="bookmark icon"
-              role="img"
-              viewBox="0 0 24 24"
-              xmlns="http://www.w3.org/2000/svg"
-            >
-              <path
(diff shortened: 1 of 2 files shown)


packages/halo/src/react/__specs__/index.spec.js
```diff
@@ -19,12 +19,12 @@ describe('Halo', () => {
   })
 
   it('renders', () => {
-    expect(() => shallow(<Halo />)).not.toThrow()
+    expect(() => mount(<Halo />)).not.toThrow()
```
jaketrent: why the change here?

I'm getting this from the docs:

> mount(<Component />) for Full DOM rendering is ideal for use cases where you have components that may interact with DOM apis, or may require the full lifecycle in order to fully test the component (ie, componentDidMount etc.)

vs.

> shallow(<Component />) for Shallow rendering is useful to constrain yourself to testing a component as a unit, and to ensure that your tests aren't indirectly asserting on behavior of child components.

packages/halo/src/react/__specs__/index.spec.js
```diff
@@ -1,5 +1,5 @@
 import React from 'react'
```
jaketrent: The commit msg is for theme, so I'm not sure if conventional commit will bump the version on the halo package or not.  But it probably doesn't matter anyway, given that this is a non-functional refactor.

packages/drawer/src/react/__specs__/__snapshots__/storyshots.spec.js.snap
```diff
@@ -21,9 +21,7 @@ exports[`Storyshots drawer controlled 1`] = `
         Toggle drawer
       </span>
     </button>
-    <div
-      data-css-1fq8nx6=""
-    >
+    <div>
```
jaketrent: What change would have removed the `data-css-` attr here?

packages/emptystate/src/react/index.js
```diff
@@ -113,34 +113,40 @@ Illustration.propTypes = {
...
-        >
-          {this.props.illustration}
-          {this.props.heading}
-          {this.props.caption}
-          {this.props.actions}
-        </div>
-      </Context.Provider>
+      <WithTheme>
+        {({ themeName }) => {
+          const ctx = { hasRenderedOnce, size, themeName }
```
jaketrent: yeah, this unexpected rewiring that wasn't required before is not my favorite...

packages/drawer/src/react/__specs__/__snapshots__/storyshots.spec.js.snap
```diff
@@ -21,7 +21,9 @@ exports[`Storyshots drawer controlled 1`] = `
         Toggle drawer
       </span>
     </button>
-    <div>
+    <div
```
jaketrent: oh, they're back!

PR #459, opened 2019-01-29
fix(datepicker): add keys to avoid errors

danethurber:
was pairing with @dannytannertantrum and tracked down(and fixed) a react key error.

### DIFF
--- a/packages/datepicker/src/react/index.js
+++ b/packages/datepicker/src/react/index.js
@@ -314,8 +314,8 @@ class DatePicker extends React.Component {
           </div>
         </Halo>
         {state.isOpen && [
-          <Overlay onClick={this.handleOverlayClick} />,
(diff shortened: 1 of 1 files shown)


packages/datepicker/src/react/index.js
```diff
@@ -314,8 +314,8 @@ class DatePicker extends React.Component {
           </div>
         </Halo>
         {state.isOpen && [
-          <Overlay onClick={this.handleOverlayClick} />,
-          <div {...styles.calendarContainer(allProps)}>
+          <Overlay onClick={this.handleOverlayClick} key="dialog-overlay" />,
```
danethurber: any better suggestion on key names?

jaketrent: no, I think it works for me.  These don't have to be globally unique, right?  Just w/in component.  In that case, probably not a big deal.

Issue #460, opened 2019-02-05
table: Cells are out of horizontal alignment when used with `Drawer`


dborodyansky:
![image](https://user-images.githubusercontent.com/3674406/52310109-1bfd7600-2957-11e9-9b4c-af9ee275be1f.png)

PR #464, opened 2019-02-07
table nested in drawer

danethurber:
fixes: #460

- style full tables that are nested inside a Drawer
- some refactoring of the table component to bring up to current standards

TODO:

- [x] docs

### DIFF
--- a/packages/table/src/css/index.js
+++ b/packages/table/src/css/index.js
@@ -7,15 +7,6 @@ import { vars as drawerVars } from '@pluralsight/ps-design-system-drawer'
 
 import * as vars from '../vars'
 
-const cellAHover = {
-  color: core.colors.white,
-  textDecoration: 'underline',
-  transition: `all ${core.motion.speedNormal}`
-}
-const cellAHoverLight = {
-  color: core.colors.black
-}
-
 export default {
   '.psds-table': {
     width: '100%',
@@ -34,10 +25,9 @@ export default {
   '.psds-table__row': {
     width: '100%',
     display: 'flex',
-    borderTop: `1px solid ${core.colors.gray04}`
-  },
-  '.psds-table__row:first-of-type': {
-    borderTop: 'none'
+    borderTop: `1px solid ${core.colors.gray04}`,
+
+    '&:first-of-type': { borderTop: 'none' }
   },
   [`.psds-table__row--${themeNames.light}`]: {
     borderTop: `1px solid ${core.colors.gray01}`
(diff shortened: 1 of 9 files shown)


packages/table/src/css/index.js
```diff
@@ -7,22 +7,17 @@ import { vars as drawerVars } from '@pluralsight/ps-design-system-drawer'
...
-}
-
 export default {
   '.psds-table': {
     width: '100%',
     fontWeight: core.type.fontWeightBook,
     fontSize: core.type.fontSizeSmall,
     lineHeight: core.type.lineHeightTight
   },
+  '.psds-table--in-drawer': {
```
danethurber: this is new

packages/table/src/react/__stories__/index.story.js
```diff
@@ -420,3 +414,36 @@ const drawerStory = storiesOf('drawer', module)
...
+        <Table.Cell>Parent Cell</Table.Cell>
+        <Table.Cell>Parent Cell</Table.Cell>
+        <Table.Cell>Parent Cell</Table.Cell>
+      </Table.Row>
+    )
+
+    return (
+      <Table>
+        <Drawer base={baseRow} startOpen>
+          <Table inDrawer>
```
danethurber: added a new `inDrawer` prop to add styling. alternative is Drawer could check if child is a table at add prop using cloneElement

jaketrent: A few thoughts:

- I want to minimize whatever any component knows about the other.  
- I want to minimize the little "mode" tweaks that devs have to add to props
- I want to minimize total props on components.

In this case, the first point bears the most weight for me in general.  

The Table already knows about Drawer (it horz pads its own rows if it has Drawer children, which have horz padd as well).  

I think the `inDrawer` prop is probably as good as the known alternate solutions.

So, ^ that's just me talking through it. :)

packages/table/src/react/index.js
```diff
@@ -1,71 +1,45 @@
-import { vars as drawerVars } from '@pluralsight/ps-design-system-drawer'
 import * as glamor from 'glamor'
-import Icon from '@pluralsight/ps-design-system-icon/react'
-import { defaultName as themeDefaultName } from '@pluralsight/ps-design-system-theme/react'
 import PropTypes from 'prop-types'
 import React from 'react'
 
+import { vars as drawerVars } from '@pluralsight/ps-design-system-drawer'
```
jaketrent: I would use `/vars` in the import instead in order to not get the whole Drawer code included here. (current untreeshakeable)

packages/table/src/react/index.js
```diff
@@ -128,17 +104,22 @@ class ColumnHeader extends React.Component {
     )
   }
 }
+
+ColumnHeader.displayName = 'Table.ColumnHeader'
+
+/* eslint-disable react/no-unused-prop-types */
```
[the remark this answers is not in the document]
danethurber: the linter can't handle it. the props are passed to a style function and the linter can't follow along

PR #465, opened 2019-02-07
fix(icon): aria-label should update svg attr

### LINKED ISSUE #462: Icon aria-label should be configurable
from @jim-cooper 

| We noticed that the `Icon` doesn't have a title property.  On the course table of contents where we show checkmark icons; for screen readers we'd love to have it say "Module Complete" instead of "Check Icon Image".  Is there a good way to do this currently?

### TODO

add prop to make this configurable

danethurber:
better screen reader support for custom icon labels

fixes #462

### DIFF
--- a/packages/icon/src/react/index.js
+++ b/packages/icon/src/react/index.js
@@ -23,23 +23,22 @@ const style = {
 const IconNotFound = () => null
 
 class Icon extends React.PureComponent {
-  getIconById(id) {
-    const Comp = icons[id] ? icons[id](React, filterReactProps) : IconNotFound
-    return <Comp />
-  }
-
   render() {
-    const { id, ...props } = this.props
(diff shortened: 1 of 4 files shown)


packages/icon/src/react/index.js
```diff
@@ -23,23 +23,22 @@ const style = {
...
 
 class Icon extends React.PureComponent {
-  getIconById(id) {
-    const Comp = icons[id] ? icons[id](React, filterReactProps) : IconNotFound
-    return <Comp />
-  }
-
   render() {
-    const { id, ...props } = this.props
+    const { id, 'aria-label': ariaLabel, ...props } = this.props
```
jaketrent: rename is needed, right?  (only to get a valid var name)

danethurber: yeah because `aria-label` isn't a valid var name

PR #476, opened 2019-02-27
Patch/theme refs

danethurber:
- updates the `withTheme` hoc to properly forward refs
- bumps react version to 16.3 to gain access to `React.forwardRef`

cc @willsoto

### DIFF
--- a/packages/actionmenu/package.json
+++ b/packages/actionmenu/package.json
@@ -30,7 +30,7 @@
   "peerDependencies": {
     "@pluralsight/ps-design-system-normalize": "^3.0.24",
     "glamor": "^2.20.0",
-    "react": ">=0.15.0 < 17.0.0"
+    "react": ">=0.16.3 < 17.0.0"
   },
   "devDependencies": {
     "@pluralsight/ps-design-system-build": "^1.9.2",
--- a/packages/row/src/react/index.js
+++ b/packages/row/src/react/index.js
@@ -96,22 +96,19 @@ const ActionBar = props => (
 )
 
 const ActionBarAction = withTheme(
-  class extends React.PureComponent {
-    render() {
-      // eslint-disable-next-line react/prop-types
-      const { icon, ...props } = this.props
-      const filteredProps = filterReactProps(props, {
-        tagName: 'button'
-      })
+  React.forwardRef((props, ref) => {
+    // eslint-disable-next-line react/prop-types
+    const { icon, ...rest } = props
+    const filteredProps = filterReactProps(rest, { tagName: 'button' })
 
-      return (
-        <button {...styles.actionBarAction(props)} {...filteredProps}>
-          {icon}
-        </button>
-      )
-    }
-  }
+    return (
+      <button {...styles.actionBarAction(props)} ref={ref} {...filteredProps}>
+        {icon}
+      </button>
+    )
+  })
 )
+
 ActionBarAction.displayName = 'Row.Action'
 ActionBarAction.propTypes = { icon: PropTypes.element.isRequired }
 
--- a/packages/theme/src/react/with-theme.js
+++ b/packages/theme/src/react/with-theme.js
@@ -12,8 +12,7 @@ const getDisplayName = Component => {
   return Component.displayName || Component.name || 'Component'
 }
 
-export default function withTheme(BaseComponent, options = {}) {
-  const { forwardRef = true } = options
+export default function withTheme(BaseComponent) {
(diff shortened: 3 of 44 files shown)


packages/theme/src/react/with-theme.js
```diff
@@ -12,17 +12,29 @@ const getDisplayName = Component => {
   return Component.displayName || Component.name || 'Component'
 }
 
-export default function withTheme(BaseComponent) {
+export default function withTheme(BaseComponent, options = {}) {
+  const { forwardRef = true } = options
```
danethurber: i added an escape hatch if we don't want to use the ref forwarding. this is similar in other hocs i've seen(redux#connect) but it's not actually used anywhere at the moment.

packages/actionmenu/package.json
```diff
@@ -30,7 +30,7 @@
   "peerDependencies": {
     "@pluralsight/ps-design-system-normalize": "^3.0.24",
     "glamor": "^2.20.0",
-    "react": ">=0.15.0 < 17.0.0"
+    "react": ">=0.16.3 < 17.0.0"
```
jaketrent: These are all the packages where Theme is used, correct?

It's funny that one of these HoCs that is used internally was one of the first to get the react@16 treatment.  I imagined more that react@16 would be set as a peerDep more on a per-package basis, when that package wanted to use one of those features.

I think it's no big deal to have a bunch of packages bump like this, because teams as essentially off react 15.  But for me, it does illuminate the coupling that we have between many packages in the list.

packages/theme/src/react/with-theme.js
jaketrent: Ok, cool.  In what case would one want to avoid ref forwarding?

packages/row/src/react/index.js
```diff
@@ -96,22 +96,19 @@ const ActionBar = props => (
...
-      )
-    }
-  }
+  React.forwardRef((props, ref) => {
+    // eslint-disable-next-line react/prop-types
+    const { icon, ...rest } = props
+    const filteredProps = filterReactProps(rest, { tagName: 'button' })
+
+    return (
+      <button {...styles.actionBarAction(props)} ref={ref} {...filteredProps}>
```
jaketrent: this was all so they could get the Action button's ref for positioning action menus, tooltips, or what-not, right?

packages/theme/src/react/with-theme.js
```diff
@@ -12,17 +12,29 @@ const getDisplayName = Component => {
...
+    return <BaseComponent {...props} ref={forwardedRef} themeName={themeName} />
   }
 
-  EnhancedComponent.BaseComponent = BaseComponent
   EnhancedComponent.contextTypes = { themeName: PropTypes.string }
-  EnhancedComponent.displayName = `withTheme(${name})`
+  EnhancedComponent.propTypes = { forwardedRef: PropTypes.any }
 
+  if (forwardRef) {
+    const Forwarded = React.forwardRef(function forwardThemeRef(props, ref) {
```
jaketrent: /me Thinking it's an interesting React API...

packages/row/src/react/index.js
```diff
@@ -96,22 +96,19 @@ const ActionBar = props => (
...
-
-      return (
-        <button {...styles.actionBarAction(props)} {...filteredProps}>
-          {icon}
-        </button>
-      )
-    }
-  }
+  React.forwardRef((props, ref) => {
+    // eslint-disable-next-line react/prop-types
```
jaketrent: /me thinking I'm pretty sure I dislike linters most of the time.

packages/actionmenu/package.json
danethurber: yeah the theme change basically bumped the version on most everything. i updated the root level react version that we use for storybook/testing as well

packages/row/src/react/index.js
danethurber: mostly so will could add a ref to the Row.Action...but yes we need greater ref support

packages/theme/src/react/with-theme.js
danethurber: i don't actually know. maybe i should just remove this and we can add in later if we need to. i'm now thinking it was a react 15 compat thing....

packages/theme/src/react/with-theme.js
danethurber: the main issue i hit is, unlike a functional component, i don't have access to context in the forwardRef so i had to get tricky

PR #486, opened 2019-03-21
card: fix action-bar z-index & pointer events

KaidenR:
### What You're Solving

- ActionBar needs a z-index > 0 because it may contain an ActionMenu that
overflows the card. The z-index will prevent the menu from going behind other
cards.
- Set pointer-events on ALL children of action-bar to be `all` so users
can interact with them. Allowed us to remove the pointer-events on
ActionBar buttons specifically.

#### Easy way: look at these screenshots:

Before: 
Screen Shot 2019-03-21 at 11 27 07 AM

After:
Screen Shot 2019-03-21 at 11 26 37 AM

#### Hard way:
Add some element to an ActionBar that overflows the card (ex: ActionMenu). Notice that the action-menu no longer goes behind the next card. It appears on top instead.

### DIFF
--- a/packages/card/src/css/index.js
+++ b/packages/card/src/css/index.js
@@ -140,6 +140,11 @@ export default {
     transition: `opacity ${core.motion.speedNormal}`,
     pointerEvents: 'none',
     opacity: 0,
+    zIndex: 10,
+
+    '> *': {
+      pointerEvents: 'all'
+    },
 
     '&:focus-within, &[focus-within]': {
       opacity: 1
@@ -154,7 +159,6 @@ export default {
 
   // __action-bar__button
   '.psds-card__action-bar__button': {
-    pointerEvents: 'all',
     fontSize: core.type.fontSizeXSmall,
     padding: 0,
     cursor: 'pointer',
--- a/packages/card/src/react/__specs__/__snapshots__/storyshots.spec.js.snap
+++ b/packages/card/src/react/__specs__/__snapshots__/storyshots.spec.js.snap
@@ -26,10 +26,10 @@ exports[`Storyshots actionBar locked visible 1`] = `
           }
         />
         <div
-          data-css-1di4zm0=""
+          data-css-1kb1o0e=""
         >
           <button
-            data-css-z6g59z=""
+            data-css-226xmr=""
             title="Bookmark"
           >
             <div
@@ -175,10 +175,10 @@ exports[`Storyshots actionBar multiple actions 1`] = `
           }
         />
         <div
-          data-css-1ifisgx=""
+          data-css-4ed7vx=""
         >
           <button
-            data-css-z6g59z=""
+            data-css-226xmr=""
             title="Bookmark"
           >
             <div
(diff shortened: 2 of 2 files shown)


packages/card/src/css/index.js
```diff
@@ -140,6 +140,11 @@ export default {
     transition: `opacity ${core.motion.speedNormal}`,
     pointerEvents: 'none',
     opacity: 0,
+    zIndex: 10,
```
jaketrent: I really like to avoid zIndex application where possible.  I feel like it's possible here.  One way to avoid it is to render the ActionMenu outside of the DOM of the card.  One could do that using React.createPortal.

Layering is just such a mess (or can be), I love to try avoid it.  

Do you feel like a portal (or other way to render the ActionMenu elsewhere in the DOM) is a good way to accomplish this?

packages/card/src/css/index.js
```diff
@@ -140,6 +140,11 @@ export default {
...
+
+    '> *': {
```
jaketrent: This is more flexible, for sure.  But it the Card `actionBar` is specifically meant to only allow as children `Card.Action`s.  

Secondarily, it's a much less-performant selector than using `.psds-card__action-bar__button` as a selector (ie, it's `*` vs a className and resolve right to left, so the first-child hierarchy doesn't make it more performant than normal `*`).  But again, this is secondary.  

I'm just wondering... why support this (or look like we support this; "this", being `actionBar` containing `*` kinds of things) if we don't.  

Is there a general design case to be made for supporting things other than `Card.Action` in the `actionBar`?

jaketrent:
Not sure what we should do with this issue atm.

packages/card/src/css/index.js
KaidenR: Hmm, we could probably figure out how to do it with portals but it seems like that would add a lot of complexity around positioning the menu in the right place. Not sure I'd want to take that on when a z-index works just as well. I definitely understand you wanting to avoid layering though.

packages/card/src/css/index.js
KaidenR: Maybe that's the core of the problems I'm having: card isn't built to support the common use case of including an action-menu in the action bar. This seemed like a more flexible way of doing that rather than asking you to build in support specifically for ActionMenus. If that's the way you'd prefer to do it than that works for me too.

jaketrent: It is a good point: that Card should support Card.Action's trigger ActionMenu appearance, because that's in the top 2 most common use cases.  

I guess it's down to a question of: Should Card support ActionMenu appearing *within* its own dom subtree?  Obviously, when making this, I thought no.  But, considering I've heard of people wanting to do it this way a couple times, maybe we go ahead and support it.  React.createPortal still feels cleaner to me, but there is the code of writing the portal rendering vs rendering inline with your Card.Action.

Thinking 2 more mins on this...

jaketrent:
I merged it.  Let's do this thing!  Thanks for your obvious patience, @KaidenR .

KaidenR:
Thanks for your time and input on this @jaketrent!

PR #503, opened 2019-04-23
Refactor/row responsive

### LINKED ISSUE #496: row: add default responsive adjustment
### Desired behavior

As a developer, my rows adjust responsively automatically. A Medium row, appears like a small row below the breakpoint (smaller font size, w/o thumbnail)

image

### Today's behavior

No responsive

### The Value Add

Responsive default = better

### The Tradeoffs

If we adjust based on media query, we may not account for all desired use cases.

danethurber:
resolves #496

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -16,11 +16,13 @@ module.exports = {
     'prettier/standard',
     'plugin:jest/recommended'
   ],
-  plugins: ['jest', 'react'],
+  plugins: ['jest', 'react', 'react-hooks'],
   env: {
     'jest/globals': true
   },
   rules: {
+    'react-hooks/exhaustive-deps': WARNING,
+    'react-hooks/rules-of-hooks': ERROR,
     'react/jsx-no-bind': [IGNORE]
   }
 }
--- a/packages/drawer/src/react/__stories__/index.story.js
+++ b/packages/drawer/src/react/__stories__/index.story.js
@@ -143,13 +143,14 @@ storiesOf('drawer', module)
     <Drawer
       base={
         <Row
-          image={<Row.Image src="https://cataas.com/cat" />}
-          title="Look at me! I'm a <Row />!"
-          metadata1={['Kitten McCatbuns', '23 hours of cuteness']}
-          actionBarVisible
           actionBar={[
             <Row.Action key="iHeartCats" icon={<Icon id="more" />} />
           ]}
+          actionBarVisible
+          image={<Row.Image src="https://cataas.com/cat" />}
+          metadata1={['Kitten McCatbuns', '23 hours of cuteness']}
+          size={Row.sizes.medium}
+          title="Look at me! I'm a <Row />!"
         />
       }
     >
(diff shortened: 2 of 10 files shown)


packages/row/src/react/index.js
```diff
@@ -374,22 +375,27 @@ renderProgress.propTypes = {
 
 // NOTE: the `title` prop clashes with a native html attr so we're exclude it
 //       from being mistakenly used in any child component
-const Row = ({ title, titleTruncated, ...props }) => (
-  <div {...styles.row(props)} {...filterReactProps(props)}>
-    {renderOverlays(props)}
+const Row = ({ title, titleTruncated, ...props }) => {
+  const desktop = useMatchMedia('(min-width: 769px)')
```
jaketrent: I at first mis-read this as just `matchMedia`.  `use-` indicates it's a hook.  I saw that later.  Even though it kinda bugs me in the native matchMedia api, but I wonder if changing the hook impl to `return { matches }` to match the native api might be more familiar?

jaketrent: perhaps that ^ or change the varName to something that doesn't seem like something you'd name the result of a native matchMedia call, like isDesktop?

packages/row/src/react/index.js
```diff
@@ -374,22 +375,27 @@ renderProgress.propTypes = {
...
+  if (!props.size) props.size = desktop ? vars.sizes.medium : vars.sizes.small
 
-    <Words image={props.image} size={props.size} actionBar={props.actionBar}>
-      <Title size={props.size} truncated={titleTruncated}>
-        {title}
-      </Title>
+  return (
```
jaketrent: what really changed here...?  hard to parse.

jaketrent: addition of the `<div {...styles.row(props)} {...filterReactProps(props)}>`?

packages/drawer/src/react/__stories__/index.story.js
```diff
@@ -143,13 +143,14 @@ storiesOf('drawer', module)
...
-          title="Look at me! I'm a <Row />!"
-          metadata1={['Kitten McCatbuns', '23 hours of cuteness']}
-          actionBarVisible
           actionBar={[
             <Row.Action key="iHeartCats" icon={<Icon id="more" />} />
           ]}
+          actionBarVisible
+          image={<Row.Image src="https://cataas.com/cat" />}
+          metadata1={['Kitten McCatbuns', '23 hours of cuteness']}
+          size={Row.sizes.medium}
```
jaketrent: just curious, why was this necessary?

.eslintrc.js
```diff
@@ -16,11 +16,13 @@ module.exports = {
     'prettier/standard',
     'plugin:jest/recommended'
   ],
-  plugins: ['jest', 'react'],
+  plugins: ['jest', 'react', 'react-hooks'],
```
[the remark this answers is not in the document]
danethurber: yeah. but these are helpful :)

packages/drawer/src/react/__stories__/index.story.js
danethurber: it keeps the snapshots from changing - since medium was the previous default behaviour

packages/row/src/react/index.js
danethurber: good call. i'll update

packages/row/src/react/index.js
danethurber: the only change is a new `props.size` based on the media query is added. it can be override if a size is provided

packages/row/src/react/use-match-media.js
```diff
@@ -0,0 +1,22 @@
+import { useEffect, useState } from 'react'
+
+export default function useMatchMedia(query) {
```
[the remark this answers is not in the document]
danethurber: srsly - loving it!

PR #507, opened 2019-04-29
feat(icon): code, trend up, and clock icons

### LINKED ISSUE #494: icon: time duration
### Desired behavior

Icon to replace the clock used here...
image

### Today's behavior

Using 3rd party library.

### The Value Add

Using icons in the design system instead of 3rd party.

danethurber:
BREAKING CHANGE: previous code icon is now called `code-branded`

resolves #494


jaketrent:
@dborodyansky , did you get looped into this?  Are these icons considered general enough for the common set?

PR #511, opened 2019-05-07
prop-types: workaround circular json error

josmardias:
### What You're Solving

When the validator receives a prop that happens to be a circular json, JSON.stringify will throw an error, and as stated on prop-types docs you can't throw inside a PropTypes.oneOf validator.

It happened to me while using `ActionMenu` with multiple `ActionMenu.Item`, the children was an array of elements and failed on stringify due to a circular json.
In my specific case, the `PropTypes.oneOf` will validate as ok if the stringify don't throw.

### Design Decisions

Probably the message `with value ${JSON.stringify(...)}` is not that pretty for this circular json situation (you can test it by passing multiple ActionMenu.Item and a non ActionMenu.Item to an ActionMenu).

Another approach is to try-catch the `JSON.stringify` but I don't know how the `with value xxx` message should look like in this case.

### How to Verify

Render an `ActionMenu` with more than one `ActionMenu.Item` on a page.

### DIFF
--- a/packages/prop-types/src/utils/index.js
+++ b/packages/prop-types/src/utils/index.js
@@ -39,3 +39,18 @@ export const getDisplayName = Component => {
 }
 
 export const isNil = value => value == null
+
+// NOTE: example from MDN docs
+//       https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Cyclic_object_value#Examples
+export const getCircularReplacer = () => {
+  const seen = new WeakSet()
+  return (key, value) => {
+    if (typeof value === 'object' && value !== null) {
+      if (seen.has(value)) {
+        return
+      }
+      seen.add(value)
+    }
+    return value
+  }
+}
--- a/packages/prop-types/src/__specs__/element-of-type.spec.js
+++ b/packages/prop-types/src/__specs__/element-of-type.spec.js
@@ -22,6 +22,19 @@ describe('elementOfType', () => {
     )
   })
 
+  it('should not throw error when expected element is a circular json', () => {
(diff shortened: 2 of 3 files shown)


danethurber:
<table>
  <thead>
    <tr>
      <th width="50"></th>
      <th width="100%" data-danger-table="true">Messages</th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:book:</td>
      <td>Thanks for contributing :tada:</td>
    </tr>
  </tbody>
</table>


<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 2ea410e6ed2f34d30ad096285eea76b660a2bb38
</p>

josmardias:
Another approuch is to try-catch the JSON.stringify but I don't know how the `with value xxx` message should in this case

packages/prop-types/src/utils/index.js
```diff
@@ -39,3 +39,18 @@ export const getDisplayName = Component => {
 }
 
 export const isNil = value => value == null
+
+// NOTE: example from MDN docs
+//       https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Cyclic_object_value#Examples
+export const getCircularReplacer = () => {
+  const seen = new WeakSet()
```
jaketrent: Verified that babel-built version of a package (ie, ActionMenu) falls back to Set if WeakSet not an available function.

Also found out that the element-of-type.js code is *included* in a webpack-bundled app in production mode.  This isn't great.  I would have thought the propTypes info would have been cleared right out.  This is not fault of this PR, but I think we should be stripping that bloat out in prod.

Issue #458, opened 2019-01-24
ps-design-system-radio: Radio button is squished with label in IE 11


jhewlett:
### Expected behavior
There is some space between the radio button and the label

### Actual behavior
The radio button is squished up against the label

![image](https://user-images.githubusercontent.com/1883526/51712599-82e06e00-1fec-11e9-9440-d5f1bd078011.png)

### Steps to repro

1. View documentation for radio

### Environment

- Browser version - IE 11

PR #519, opened 2019-05-13
fix(radio): [ie11] dont collapse margin on label

jaketrent:
Fixes: #458

Could not get to the bottom of why only IE did this.  It has something
to do with the Halo positioning.  Think it's related to IE's treatment
of splitting up the flex space.  Instead of flexing, just
display-inlined the circle and label.

### DIFF
--- a/packages/radio/src/css/index.js
+++ b/packages/radio/src/css/index.js
@@ -14,12 +14,19 @@ export default {
   },
 
   '.psds-radio-button': {
+    display: 'flex',
+    alignItems: 'center',
     padding: `${core.layout.spacingXXSmall} 0`,
     cursor: 'pointer'
   },
 
   '.psds-radio-button__halo': {
     transform: 'translateY(3px)'
+  },
+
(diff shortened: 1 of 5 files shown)


packages/radio/src/css/index.js
```diff
@@ -2,29 +2,26 @@ import core from '@pluralsight/ps-design-system-core'
...
-  // group
-
   '.psds-radio-group:focus': {
     outline: 'none'
   },
   '.psds-radio-group--disabled': {
     opacity: '0.5'
   },
 
-  // __button-container
```
danethurber: do we want to drop this convention?

jaketrent: yes, I think so.  Specifically, keep the newline, but remove the comment.

jaketrent:
@danethurber , I adjusted again in order to get better label wrapping.  Added extra div to help layout.  IE seems to be doing ok with it.

PR #523, opened 2019-05-13
 use naming conventions for theme classes

### LINKED ISSUE #432: Halo: Adjust selectors for theme
From this
https://github.com/pluralsight/design-system/blob/master/packages/halo/src/css/index.js#L49

To
.psds-theme--${themeNames.light} etc

danethurber:
resolves #432


packages/halo/src/css/index.js
```diff
@@ -46,31 +52,31 @@ export default {
     }
   },
 
-  [`${BASE_CLASSNAME}--gap-size-default${BASE_CLASSNAME}--theme-dark`]: {
+  [`${BASE_CLASSNAME}--gap-size-default${themeClasses[themeNames.dark]}`]: {
```
jaketrent: Since we have teams coding to these selector names, I believe this is a BREAKING CHANGE.

packages/halo/src/react/index.js
```diff
@@ -6,19 +6,19 @@ import React from 'react'
 import { withTheme } from '@pluralsight/ps-design-system-theme/react'
 import filterReactProps from '@pluralsight/ps-design-system-filter-react-props'
 
-import css, { BASE_CLASSNAME } from '../css'
+import css, { BASE_CLASSNAME, themeClasses } from '../css'
```
jaketrent: While you're editing this file/line, you might as well add the /index.js filename to the pathing.

Issue #524, opened 2019-05-15
emptystate protype error


danethurber:
https://pluralsight.slack.com/archives/C70HPSJPP/p1557851216017900

PR #525, opened 2019-05-15
EmptyState: resize observer state should not update after unmounting component

danethurber:
related #524 

there was an error being called in the resize-observer of emptystate. i have been unable to track it down and repro 😬 but it might be related to a few things i've done...shooting in the dark here

- There was an issue where state was being updated if the component was quickly mounted and unmounted. this was a race condition with the ResizeObserver callback
  - i have a solution that checks and only updates component state if the component is still mounted
  - it turned out easer to do with a hook so i went ahead and upgrade the component to use react 16.8
- i upgrade prop-types just because it might be related...but probably not

BREAKING CHANGE: react version upgrade

---

maybe fixes: #524


packages/emptystate/src/react/__specs__/storyshots.spec.js
```diff
@@ -6,8 +6,6 @@ import initStoryshots, {
 jest.mock('@pluralsight/ps-design-system-storybook-addon-center')
 jest.mock('@pluralsight/ps-design-system-storybook-addon-theme')
 
-jest.mock('../resize-observer')
```
jaketrent: why the mock removal for storyshots?

packages/emptystate/src/react/index.js
```diff
@@ -110,55 +108,35 @@ Illustration.propTypes = {
...
-          const ctx = { hasRenderedOnce, size, themeName }
-          return this.renderContent(ctx)
-        }}
-      </ResizeObserver>
-    )
-  }
-}
+const EmptyState = withTheme(function EmptyState(props) {
+  const [hasRenderedOnce, setHasRenderedOnce] = useState(false)
+  const [ref, { width }] = useResizeObserver()
```
jaketrent: kind of an odd api to return an array with an object that is then destructured.  Why not just return an object and then destructure ref out of it as well?  Seems it would simplify the data structure.

packages/emptystate/src/react/index.js
```diff
@@ -110,55 +108,35 @@ Illustration.propTypes = {
...
+
+  useEffect(() => {
+    setHasRenderedOnce(true)
+  }, [])
+
+  const hasPropSizeOverride = !!props.size
+  const observableSize = width <= 450 ? vars.sizes.small : vars.sizes.large
+
+  const size = hasPropSizeOverride ? props.size : observableSize
+  const ctx = { size, themeName: props.themeName }
```
jaketrent: Why pull theme from context just to add it to context again?  I guess it's a different context.  Seems odd at first glance.

packages/emptystate/src/react/index.js
```diff
@@ -110,55 +108,35 @@ Illustration.propTypes = {
...
+const EmptyState = withTheme(function EmptyState(props) {
+  const [hasRenderedOnce, setHasRenderedOnce] = useState(false)
+  const [ref, { width }] = useResizeObserver()
+
+  useEffect(() => {
+    setHasRenderedOnce(true)
+  }, [])
+
+  const hasPropSizeOverride = !!props.size
+  const observableSize = width <= 450 ? vars.sizes.small : vars.sizes.large
```
jaketrent: What is the 450 magic number?  Saving as a variable in order to label would be nice.

packages/emptystate/src/react/use-resize-observer.js
```diff
@@ -0,0 +1,50 @@
...
+      setHeight(contentRect.height)
+    }
+
+    const debouncedResize = debounce(150, handleResize)
+    const observer = new ResizeObserver(debouncedResize)
+
+    observer.observe(el)
+
+    return () => {
+      subscribed = false
```
jaketrent: Is this where we're setting "unmounted"?

packages/emptystate/src/react/use-resize-observer.js
```diff
@@ -0,0 +1,50 @@
...
+    function handleResize(entries) {
+      if (!subscribed || !isBrowser) return
+      const { contentRect } = entries[0]
+
+      setWidth(contentRect.width)
+      setHeight(contentRect.height)
+    }
+
+    const debouncedResize = debounce(150, handleResize)
+    const observer = new ResizeObserver(debouncedResize)
```
jaketrent: `useEffect` is potentially called multiple times.  Does creating new instance here create a problem?  Do we want to ever just have one?

packages/emptystate/src/react/use-resize-observer.js
```diff
@@ -0,0 +1,50 @@
...
+      subscribed = false
+      clearTimeout(debouncedResize)
+      observer.unobserve(el)
+    }
+  }, [])
+
+  return [ref, { width, height }]
+}
+
+function debounce(delay, fn) {
```
jaketrent: beaut

packages/emptystate/src/react/__specs__/storyshots.spec.js
danethurber: before it was an hoc and therefore muddied up my snapshots with elements. now that it's a hook the snapshot is simplier

packages/emptystate/src/react/index.js
danethurber: before it was called `renderSmallIfElementLessThan` 😛 

i'll go back to calling it that

packages/emptystate/src/react/index.js
danethurber: yeah i can see that. i'll update

packages/emptystate/src/react/use-resize-observer.js
danethurber: yes - not the prettiest thing in the world

packages/emptystate/src/react/use-resize-observer.js
danethurber: because the second arg to useEffect is an empty array it should(🤞) just run once on mount. i'm still fuzzy on the api so i'll verify that's how it works

PR #526, opened 2019-05-15
Feature/position

jaketrent:
- [x] js function
- [x] react component
- [x] fix the single react issue with site
- [x] update button to be forwardRef compat (all components that want to use Position component will need to support forwardRef)
- [x] update tooltip docs
- [x] portal rendering support
- [x] position docs

Some things to look for?:
- [ ] button still looks ok. I'm concerned because in the site, Button page has weird styles for "flat" appearance and light them.  Button storybook, however, looks fine.  Maybe site is messed?  
- [ ] tooltip and button conversions to forwardRef look good.  Is there something to do to test this more?

This got pretty big.  I am ready to walk through it or answer questions.


packages/position/src/react/__stories__/index.story.js
```diff
@@ -0,0 +1,131 @@
...
+  ))
+)
+
+reactStory.add('when=false', () => (
+  <Below when={false} show={<Tooltip>The tip</Tooltip>}>
+    <Box>Hidden</Box>
+  </Below>
+))
+
+reactStory.add('test.skip inNode', () => {
```
danethurber: `test.skip` - guessing that's a typo?

packages/button/package.json
```diff
@@ -21,9 +21,9 @@
   "keywords": [],
   "dependencies": {
     "@pluralsight/ps-design-system-core": "^4.4.2",
+    "@pluralsight/ps-design-system-filter-react-props": "^1.1.3",
     "@pluralsight/ps-design-system-icon": "^12.0.1",
     "@pluralsight/ps-design-system-theme": "^2.0.2",
-    "@pluralsight/ps-design-system-util": "^2.0.2",
```
danethurber: i think you still need this for the `transparentize` function

packages/position/src/react/__stories__/index.story.js
jaketrent: No, that's a pattern for storyshots to see and not test.

packages/button/package.json
jaketrent: I don't think I'm using transparentize any more.  Unless grep is failing me...

jaketrent:
hmmm, yeah, surprisingly the docs are messed for me too.  Not sure why it was working earlier. 👀 ...

packages/button/package.json
jaketrent: Wait, you're right, this is `button`...

packages/position/src/react/__stories__/index.story.js
danethurber: oh cool. i didn't think that was a possiblity

jaketrent:
Hey @danethurber , the docs were broken because I was of bad combo of forwardRef and useRef.  I don't know why it seemed to work when I deved on it.  

Please re-review when you have the time.

Issue #528, opened 2019-05-20
note/comment component


jaketrent:
Design (Web): https://www.figma.com/file/YsTklBecdddy9RZ3HXddIseD/PSDS-Web?node-id=2366%3A310 
Design (specced): https://www.figma.com/file/TbG4IGHd8th48OGQiQxYAwxb/DS-Note-Component

No nested comments

Todo:
- [x] Followup with devtective on their usage

jaketrent:
Triage: need design review

jaketrent:
For the design review:

```
## Note design review

- Called "note"?
- Avatar - just the one size?
- Clarify: Avatar and "Creator Name"/title always appear *together*?
- Message data type?
- Metadata:
  - data type?
  - limit on items?
- actionBar 
  - data type? 
  - Note.Action a la Card?
  - limit on actions?
- Edit state part of DS? Separate component?
  - Don't want to ship edit bits to read-only experiences
  - Perhaps separate EditNote component?

#### Proposed API

```
<Note 
  avatar={<Avatar name="Jake Trent" src="http://img.jpg" />}
  message={}
  metadata={[ , ]}
  actionBar={}
/>
```

- Avatar size overridden internally
- Avatar name drives the Note title
```

I plan on bringing this up in standup today.  I'm hoping this will speed up the conversation to fit in that mtg. cc @danethurber , @dborodyansky .

jaketrent:
Design review outcomes:
- Ask devtective if they need more than strings.
- Message *maybe* as children
- Metadata as in Card
- Metadata no limit in count
- actionBar as in Card, locked viz, etc
- 2 Actions max
- No edit case
- Explicit heading prop
- Semantics on heading={<h3>string only/name</h3>}, no link
- Build this for threaded convo that devtective is working on; mult teams in content tools plan to use
- Defer onClick on the whole region; no anchor/href support;

jaketrent:
- devtective - node, because multi-paragraphs, formatting over lines, workflow comments
- message={node} - Text.P x 1-n example
- margin from parent; first-child remove top margin; last-child remove bottom margin; rely on collapse to not double up spacing
- linkable - ask devtective, otherwise .x release when someone asks for it
- meta flex space handled like Card/Row, nowrap
- Divider in stack of notes follow css rules of Row, border is on top of each item (except for first)
- List semantics - 
<Note.List>
<Note />
<Note />
<Note />
</Note.List>
- TODO: Dmitry contact devtective

[...]

PR #543, opened 2019-06-04
Note

danethurber:
resolves: #528 

Outstanding tasks

- [x] metadata
- [x] actions/action menu
- [x] limit actions to 2
- [x] linkable avatar/name
- [x] linkable metadata
- [x] docs
- [x] override link style
- [x] metadata first item prominence
- [x] rethink selector patterns
- [x] update eslint to warn on missing extensions
- [x] add file extensions to imports
- [x] actionBarVisible prop and actions visible on hover


packages/note/src/react/index.js
```diff
@@ -0,0 +1,214 @@
+import * as glamor from 'glamor'
```
jaketrent: As we look forward to using ES modules in order to tree shake, I think we should start using named imports specifically where we are not "for sure going to use everything in this package".  This is because, as I understand it, if you `import * from`, nothing in that imported package can be tree-shaken.

I recently attempted this in datawell: https://github.com/pluralsight/design-system/blob/master/packages/datawell/src/react/index.js#L1

What do you think?

packages/prop-types/src/array-of-max-length.js
```diff
@@ -0,0 +1,45 @@
+import PropTypes from 'prop-types'
+
+import PropTypeError from './prop-type-error'
+import { createChainableTypeChecker, getCircularReplacer } from './utils'
+
+export default function arrayOfMaxLength(
```
jaketrent: This is the right solution.  But given the other issue that we've observed, where proptype code isn't being removed from builds (eg, elementOfType), this is a lot of code for validating proptypes, when we could just use `Array.slice` to limit what's displayed.  So, I'm not sure what's best at the moment, considering the bundle size tension here.  Thoughts?

packages/prop-types/src/array-of-max-length.js
```diff
@@ -0,0 +1,45 @@
+import PropTypes from 'prop-types'
+
+import PropTypeError from './prop-type-error'
+import { createChainableTypeChecker, getCircularReplacer } from './utils'
```
jaketrent: add file ext. here and a few other places in the pr.   ...not sure if I should mark them all.  is it noise or helpful.  What do you think?

packages/datawell/src/css/index.js
```diff
@@ -5,9 +5,7 @@ export default {
   '.psds-datawell': {
     borderLeft: `1px solid ${core.colors.gray03}`,
     minHeight: '104px',
-    padding: `0 ${core.layout.spacingMedium} ${core.layout.spacingMedium} ${
-      core.layout.spacingLarge
-    }`
+    padding: `0 ${core.layout.spacingMedium} ${core.layout.spacingMedium} ${core.layout.spacingLarge}`
```
jaketrent: I think that even small things like formatting should be kept out of a PR if it's for an unrelated thing.  It becomes noise.  And we shouldn't need to test the other thing with our eyes in a code review or by running code to make sure in the context of a PR.  Nbd, but I think it would be helpful.

packages/note/src/css/index.js
```diff
@@ -0,0 +1,129 @@
...
+  margin: 0,
+  overflow: 'visible',
+  padding: 0,
+  width: 'auto',
+
+  MozOsxFontSmoothing: 'inherit',
+  WebkitAppearance: 'none',
+  WebkitFontSmoothing: 'inherit',
+
+  '&::-moz-focus-inner': {
```
jaketrent: fancy!  Do we need all these special bits lines 15-21?  If so, how did you discover that we did?

packages/note/src/css/index.js
```diff
@@ -0,0 +1,129 @@
...
+  },
+
+  '.psds-note': {
+    alignItems: 'flex-start',
+    display: 'flex',
+    fontWeight: core.type.fontWeightMedium,
+    lineHeight: core.type.lineHeightStandard,
+    padding: `${core.layout.spacingLarge} ${core.layout.spacingMedium}`,
+
+    'li &': { borderTop: `1px solid ${core.colors.gray03}` },
```
jaketrent: I think this is a first for... goodness, what is this kind of selector called...  the "when I'm a child of a particular parent" selector :)... anyway, we have traditionally tried to be very limited on nesting that we were willing to do.  Is this new type of nesting one we want to put in-bounds generally?  Or in the particular case, do we want to make an exception?  

The alternative is something like a top-level `.psds-note__list-item .psds-note`, right?

packages/note/src/css/index.js
```diff
@@ -0,0 +1,129 @@
...
+  },
+  [`.psds-note.psds-theme--${themeNames.light}`]: {
+    borderColor: core.colors.gray01
+  },
+
+  '.psds-note__action-bar': {
+    color: core.colors.gray02,
+    display: 'flex',
+
+    'div + &': { marginLeft: core.layout.spacingMedium }
```
jaketrent: Part of why I wonder about this nesting type is that I have a hard timing diving its meaning.  

Does this mean "When an actionBar is sibling to a metadatum?"  It took me a running around in the react code to see html hierarchy and tagNames to figure that out, if it's even correct. :)

The alternative would be something like `.psds-note__metadata + .psds-note__action-bar`, right?

What are your thoughts on the readability?

packages/note/src/css/index.js
```diff
@@ -0,0 +1,129 @@
...
+  },
+
+  '.psds-note__action': {
+    ...resetButton,
+    cursor: 'pointer',
+    fontSize: core.type.fontSizeXSmall,
+    transition: `all ${core.motion.speedNormal}`,
+
+    '&:hover, &:active': { color: core.colors.white },
+    '& + &': { marginLeft: core.layout.spacingSmall }
```
jaketrent: For some reason, this is the easiest of all of these nested selectors to read because `&` occupies both the first and second position, making it easy to discern as a "2nd and onward sibling" selector.

packages/note/src/react/__specs__/index.spec.js
```diff
@@ -0,0 +1,31 @@
...
+    expect(getByTestId('the-message')).toBeInTheDocument()
+  })
+
+  it('renders metadata', () => {
+    const { container } = render(
+      <Note message={'message'} metadata={['meta1', 'meta2']} />
+    )
+
+    expect(container).toHaveTextContent('meta1')
+    expect(container).toHaveTextContent('meta2')
```
jaketrent: I like these general render tests.  Probably because in the latest Tab commits I disabled storyshots, leaving it with essentially no tests.

Either we need to get hooks working correctly and reliably in storybook or move to tests more like this to compliment the storybook viz dev env.

packages/note/src/react/__stories__/index.story.js
```diff
@@ -0,0 +1,185 @@
...
+    <NoteWithDefaults
+      actionBar={[
+        <Note.Action icon={<Icon id={Icon.ids.bookmark} />} title="Bookmark" />,
+        <Note.Action icon={<Icon id={Icon.ids.more} />} title="More" />
+      ]}
+    />
+  ))
+  .add('with long heading', _ => (
+    <NoteWithDefaults
+      heading="This is probably the greatest thing that's ever happened in my life"
```
jaketrent: What happens when the heading gets longer than 2 lines?

packages/note/src/react/__stories__/index.story.js
```diff
@@ -0,0 +1,185 @@
...
+          </Text.P>
+        </React.Fragment>
+      }
+    />
+  ))
+
+storiesOf('Note/metadata', module)
+  .addDecorator(fn => <ConstrainWidth>{fn()}</ConstrainWidth>)
+  .add('with long strings', _ => (
+    <NoteWithDefaults metadata={longStringsMetaData} />
```
jaketrent: The Design Review guidance is very terse:

```
Meta flex space handled like Card/Row
```

We did mention in verbally in the review, but the 1st meta data should be given twice the flex space as it's likely the most important.

packages/note/src/react/index.js
```diff
@@ -0,0 +1,214 @@
...
+  </button>
+))
+
+Action.displayName = 'Note.Action'
+Action.propTypes = {
+  icon: elementOfType(Icon).isRequired,
+  title: PropTypes.string.isRequired
+}
+
+function AvatarLink(props) {
```
jaketrent: What was the design thought around making a specialized `Note.AvatarLink` but not an equivalent `Note.TextLink` for the header or meta data (a la Card/Row), instead relying on them using the `<Link/>` package tag?  

Or did I misread what the design even is? :)

packages/note/src/react/index.js
```diff
@@ -0,0 +1,214 @@
...
+AvatarLink.propTypes = {
+  children: PropTypes.shape({
+    type: PropTypes.oneOf(['a']),
+    props: PropTypes.shape({
+      children: elementOfType(Avatar).isRequired
+    }).isRequired
+  }).isRequired
+}
+
+function Aside(props) {
```
jaketrent: Loving the consistent function style.

packages/note/src/react/index.js
```diff
@@ -0,0 +1,214 @@
...
+
+import css from '../css/index.js'
+
+const styles = {
+  note: (themeName, props) =>
+    glamor.compose(
+      glamor.css(css['.psds-note']),
+      glamor.css(css[`.psds-note.psds-theme--${themeName}`])
+    ),
+
```
jaketrent: Kind of an interesting order, with note up top separated by a newline.  It leaves me curious as to the meaning of the ordering... 🤔

packages/note/src/react/index.js
```diff
@@ -0,0 +1,214 @@
...
+  actionBar: themeName =>
```
jaketrent: A bit of a change here, with the styles functions not taking "props", instead just `themeName` as a single string.  What was your thought there?

Oh, I'm just seeing the `note: (themeName, props) =>` signature above, also different.

packages/site/pages/components/note.js
```diff
@@ -0,0 +1,240 @@
...
+    message={(
+      <p>Clothes make the man. Naked people have little or no influence on society.</p>
+    )}
+  />
+</Note.List>
+`
+        ]}
+      />
+    </Content>
+  </Chrome>
```
jaketrent: Getting a few errors in-browser on this page:

```
warning.js?3119d7c:33 Warning: Failed prop type: Invalid prop `avatar` supplied to `Note`.
    in Note
    in ReactExample (at note.js?entry:103)
  


warning.js?3119d7c:33 Warning: Failed prop type: Invalid prop `children.props.children` of element type `Avatar` supplied to AvatarLink, expected element of type `Avatar`
    in AvatarLink
    in ReactExample (at note.js?entry:121)
   


warning.js?3119d7c:33 Warning: Failed prop type: Invalid prop `icon` of element type `Icon` supplied to Note.Action, expected element of type `Icon`
    in Note.Action
    in ReactExample (at note.js?entry:159)
```

packages/site/pages/components/note.js
```diff
@@ -0,0 +1,240 @@
...
+))
```
jaketrent: Even more mysterious, when I refresh the `/components/note` page, I get nextjs not rendering, but giving this:

```
(0 , _psDesignSystemPropTypes.arrayOfMaxLength) is not a function
TypeError: (0 , _psDesignSystemPropTypes.arrayOfMaxLength) is not a function
    at Object.<anonymous> (/Users/jaketrent/dev/design-system/packages/note/dist/react/index.js:261:60)
    at Module._compile (internal/modules/cjs/loader.js:723:30)
    at Module._compile (/Users/jaketrent/dev/design-system/node_modules/source-map-support/source-map-support.js:492:25)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:734:10)
    at Module.load (internal/modules/cjs/loader.js:620:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:560:12)
    at Function.Module._load (internal/modules/cjs/loader.js:552:3)
    at Module.require (internal/modules/cjs/loader.js:659:17)
    at require (internal/modules/cjs/helpers.js:22:18)
    at Object.<anonymous> (/Users/jaketrent/dev/design-system/packages/note/react.js:1:18)
```

packages/datawell/src/css/index.js
danethurber: yeah i agree. i'm unsure why this got changed...maybe my new editor setup isn't working as expected - will investigate

packages/note/src/css/index.js
danethurber: i pulled these styles over from prism...which probably got it from an article somewhere 😄 

i'll look at see what the current state of webkit/moz is and if there are no longer needed

packages/note/src/css/index.js
```diff
@@ -0,0 +1,129 @@
...
+export default {
+  '.psds-note__list': {
+    listStyle: 'none',
+    marginLeft: 0
+  },
+
+  '.psds-note': {
+    alignItems: 'flex-start',
+    display: 'flex',
+    fontWeight: core.type.fontWeightMedium,
```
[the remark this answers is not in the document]
danethurber: good catch - i'll fix

packages/note/src/css/index.js
danethurber: yes i do think i could make these selectors more readable - i'll see what i can come up with

packages/note/src/react/__specs__/index.spec.js
danethurber: i haven't had issues with storybook yet...but i know that enzyme doesn't support hooks yet which could be an issue. hopefully we can just continue to use storyshots 🤞

packages/note/src/react/__stories__/index.story.js
```diff
@@ -0,0 +1,185 @@
...
+            <Avatar name="Bob Ross" src="//picsum.photos/128" />
+          </a>
+        </Note.AvatarLink>
+      }
+    />
+  ))
+  .add('with linked heading', _ => (
+    <NoteWithDefaults
+      heading={
+        <p>
```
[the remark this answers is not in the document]
danethurber: i felt that the `a` should be wrap in a `p` tag. it seemed more semantic

packages/note/src/react/__stories__/index.story.js
danethurber: ok cool - i missed that but will update

packages/note/src/react/__stories__/index.story.js
danethurber: currently it just wraps

packages/note/src/react/index.js
danethurber: i'm ok with that going forward 👍

packages/note/src/react/index.js
danethurber: i don't think there was a reasoning behind it. just an oversite.

packages/note/src/react/index.js
danethurber: now that themeName isn't coming from props but from `useContent` is seemed better to pass it to the styles separate from `props`. i opted for doing `themeName` as the first arg and `props` as the second

packages/note/src/react/index.js
danethurber: `AvatarLink` was necessary for setting the avatar size...where a `TextLink` component wouldn't do anything differently than just the normal `Link`

packages/prop-types/src/array-of-max-length.js
danethurber: i'll go through and search/destroy

danethurber: i need to see if we can update eslint to warn me about this...cause i never remember

packages/prop-types/src/array-of-max-length.js
danethurber: oh i wasn't aware that the code isn't be eliminated. i thought about using slice but the consumer of the component would just see something different than expected(ie only two items)...i feel like a warning is more useful/actionable

packages/site/pages/components/note.js
```diff
@@ -0,0 +1,240 @@
...
+            ]),
+            PropTypes.row([
+              'avatar',
+              <span>
+                <code>Avatar</code>
+                {' | '}
+                <code>Note.AvatarLink</code>
+              </span>,
+              null,
+              null
```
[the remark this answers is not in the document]
danethurber: oh yeah - i planned to do that and totally spaced 😬

packages/site/pages/components/note.js
danethurber: **fixed**. i'm seeing that the react team doesn't like this type of prop checker - it implies inheritance and is less flexible. however since we use cloneElement and the children have to be of a specific type i don't know what the alternative is

.eslintrc.js
```diff
@@ -21,6 +21,7 @@ module.exports = {
     'jest/globals': true
   },
   rules: {
+    'import/extensions': [WARNING, 'always', { ignorePackages: true }],
```
danethurber: i thought we could start with warnings on missing file extensions...but maybe we should go all the way to error? thoughts?

jaketrent: Yeah, I like warnings.

danethurber:
thanks @jaketrent. i get what you're saying now..and yes those where simple changes 👍

packages/note/src/react/index.js
```diff
@@ -0,0 +1,224 @@
...
+    </span>
+  )
+}
+
+Note.Action = Action
+Note.AvatarLink = AvatarLink
+Note.List = NoteList
+
+Note.propTypes = {
+  actionBar: arrayOfMaxLength(2, PropTypes.node),
```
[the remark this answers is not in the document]
dborodyansky: Yes, visible or on hover, like the card/row

packages/note/src/react/use-hover.js
```diff
@@ -0,0 +1,28 @@
+import { useEffect, useRef, useState } from 'react'
+
+export default function useHover() {
+  const [isHovered, setIsHovered] = useState(false)
+  const ref = useRef(null)
```
jaketrent: nit: null is not required here

packages/note/src/react/use-hover.js
```diff
@@ -0,0 +1,28 @@
...
+  const handleMouseOver = () => setIsHovered(true)
+  const handleMouseOut = () => setIsHovered(false)
+
+  useEffect(
+    function setup() {
+      const node = ref.current
+
+      if (!node) return
+
+      node.addEventListener('mouseover', handleMouseOver)
```
jaketrent: This is an interesting approach.  Seems like the more react approach would be to avoid refs entirely (since they're just an escape hatch to the dom) and instead use normal event handlers as props (ie, `onMouseOver`, etc). 

What do you think?

packages/note/src/react/__stories__/index.story.js
```diff
@@ -20,6 +20,7 @@ const ConstrainWidth = props => <div {...props} style={{ maxWidth: '420px' }} />
 
 const NoteWithDefaults = props => <Note {...props} />
 NoteWithDefaults.defaultProps = {
+  actionBarVisible: true,
```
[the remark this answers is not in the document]
danethurber: it was me. this is for the stories not the actual implementation. confusing enough?

packages/note/src/css/index.js
```diff
@@ -47,11 +47,16 @@ export default {
   '.psds-note__action-bar': {
     color: core.colors.gray02,
     display: 'flex',
-    marginLeft: 'auto'
+    marginLeft: 'auto',
+    opacity: 0,
+    transition: `opacity ${core.motion.speedNormal}`
```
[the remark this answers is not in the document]
danethurber: will do. i'll update card as well

packages/note/src/react/use-hover.js
danethurber: yeah that makes sense. this hook could be shared but that seems overly complicated and not actually what we need. i'm in a hook all the things mode...i need to step back and think sometimes 😉

packages/note/src/react/__stories__/index.story.js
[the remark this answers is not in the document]
jaketrent: oh, lol, I see it now.  Thanks

packages/note/src/react/__specs__/index.spec.js
```diff
@@ -28,4 +28,30 @@ describe('Note', () => {
     expect(container).toHaveTextContent('meta1')
     expect(container).toHaveTextContent('meta2')
   })
+
+  it('supports onMouseOver prop', async () => {
```
jaketrent: nice tests!

Issue #554, opened 2019-06-20
actionmenu: support new ref api for positioning


jaketrent:
### Desired behavior

Make ActionMenu support Position.

Related: https://pluralsight.slack.com/archives/C70HPSJPP/p1561062442019300

### Today's behavior

Uses old ref api, unsupported.

### The Value Add

The convenience of Position on more components.

### The Tradeoffs

Takes time to migrate.  Locks in ref usage to only new React ref API.  But this is the future, so we need to get there.

PR #563, opened 2019-06-26
Refactor/actionmenu refs

danethurber:
- refactor ActionMenu to the latested and greatest
- Position has 4 new options 💥 

fixes https://github.com/pluralsight/design-system/issues/554

### DIFF
--- a/packages/actionmenu/src/react/__specs__/__snapshots__/storyshots.spec.js.snap
+++ b/packages/actionmenu/src/react/__specs__/__snapshots__/storyshots.spec.js.snap
@@ -1384,3 +1384,107 @@ exports[`Storyshots shouldFocusOnMount false 1`] = `
   </div>
 </div>
 `;
+
+exports[`Storyshots with Position example 1`] = `
+<div
+  data-css-claeju=""
+>
+  <div
+    style={
+      Object {
+        "margin": "16px",
+      }
+    }
+  >
+    <button
+      data-css-fu22rr=""
+      disabled={false}
+      onClick={[Function]}
+      style={Object {}}
+    >
+      <div
+        data-css-y80vay=""
+      >
+        <div
+          data-css-otejxm=""
+        >
+          <svg
+            aria-label="more icon"
+            role="img"
+            viewBox="0 0 24 24"
+            xmlns="http://www.w3.org/2000/svg"
+          >
+            <path
+              d="M6 14.5c-1.1046 0-2-.8954-2-2s.8954-2 2-2 2 .8954 2 2-.8954 2-2 2zm12 0c-1.1046 0-2-.8954-2-2s.8954-2 2-2 2 .8954 2 2-.8954 2-2 2zm-6 0c-1.1046 0-2-.8954-2-2s.8954-2 2-2 2 .8954 2 2-.8954 2-2 2z"
+            />
+          </svg>
+        </div>
+      </div>
+      <span
+        data-css-15b13by=""
+      />
+    </button>
+    <div
+      style={
+        Object {
+          "left": "0px",
+          "position": "absolute",
+          "top": "12px",
+        }
+      }
+    >
+      <div
+        data-css-1pg0ko5=""
+        onKeyDown={[Function]}
+        role="menu"
+      >
+        <div
+          data-css-13azwyo=""
+        >
+          <button
+            aria-haspopup={false}
+            data-css-1stw8mg=""
+            onFocus={[Function]}
+            onKeyDown={[Function]}
+            onMouseOver={[Function]}
+            role="menuitem"
+            tabIndex="0"
+          >
+            One item
+          </button>
+        </div>
+        <div
+          data-css-13azwyo=""
+        >
+          <button
+            aria-haspopup={false}
+            data-css-1stw8mg=""
+            onFocus={[Function]}
+            onKeyDown={[Function]}
+            onMouseOver={[Function]}
+            role="menuitem"
+            tabIndex="0"
+          >
+            Two item
+          </button>
+        </div>
+        <div
+          data-css-13azwyo=""
+        >
+          <button
+            aria-haspopup={false}
+            data-css-1stw8mg=""
+            onFocus={[Function]}
+            onKeyDown={[Function]}
+            onMouseOver={[Function]}
+            role="menuitem"
+            tabIndex="0"
+          >
+            Three item
+          </button>
+        </div>
+      </div>
+    </div>
+  </div>
+</div>
+`;
--- a/packages/actionmenu/src/react/__specs__/__snapshots__/storyshots.spec.js.snap
+++ b/packages/actionmenu/src/react/__specs__/__snapshots__/storyshots.spec.js.snap
@@ -1385,7 +1385,7 @@ exports[`Storyshots shouldFocusOnMount false 1`] = `
 </div>
 `;
 
-exports[`Storyshots with Position example 1`] = `
+exports[`Storyshots with Position BelowLeft 1`] = `
 <div
   data-css-claeju=""
 >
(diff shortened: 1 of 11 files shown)


packages/actionmenu/src/react/index.js
```diff
@@ -193,104 +196,106 @@ const calcNestedMenuPosition = (menuWidth, origin) =>
...
+  React.useEffect(() => {
+    ref.current = value
+  }, [value])
+
+  return ref.current
+}
+
+const Item = props => {
+  const { icon, isActive, ...rest } = props
+  const Tag = props.href ? 'a' : 'button'
```
jaketrent: "Tag" was a confusing name for me.  For instance, we have a DS "Tag".  This was "tagName" in the previous solution, which I understand, but that looks odd: `<TagName />`.  Maybe semantically what it is? Which is what?  Root/Container/etc.??  What do you think?

packages/actionmenu/src/react/index.js
```diff
@@ -193,104 +196,106 @@ const calcNestedMenuPosition = (menuWidth, origin) =>
...
-  componentDidUpdate(prevProps) {
-    if (
-      !prevProps.isActive &&
-      this.props.isActive &&
-      !this.state.isNestedRendered
-    )
-      this.item.focus()
+  }, [isActive, isNestedRendered, prevIsActive])
+
+  const handleFocus = evt => {
```
jaketrent: Do we want/care to use the `function handleFocus` convention?

packages/actionmenu/src/react/__specs__/__snapshots__/storyshots.spec.js.snap
```diff
@@ -1384,3 +1384,107 @@ exports[`Storyshots shouldFocusOnMount false 1`] = `
   </div>
 </div>
 `;
+
```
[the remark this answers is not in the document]
danethurber: interesting. i'll have to figure out what's happening. great catch

packages/actionmenu/src/react/__stories__/index.story.js
```diff
@@ -306,3 +310,40 @@ storiesOf('disabled', module)
       </ActionMenu.Item>
     </ActionMenu>
   ))
+
+storiesOf('with Position', module)
```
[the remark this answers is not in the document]
danethurber: mostly i'm just validating that it works with the Position component. i see what you're saying about the alignment. there isn't anything custom but i can dig into what's happening and see what can be done to make the two components work better together #teamwork

packages/actionmenu/src/react/index.js
danethurber: yeah good call on not using `Tag`. `TagName` could work, i've also seen `Element` in the wild

packages/actionmenu/src/react/index.js
```diff
@@ -193,104 +196,106 @@ const calcNestedMenuPosition = (menuWidth, origin) =>
...
-class Item extends React.Component {
-  constructor(props) {
-    super(props)
-    this.state = {
-      isNestedRendered: false
+function usePrevious(value) {
+  const ref = React.useRef()
+
+  React.useEffect(() => {
+    ref.current = value
```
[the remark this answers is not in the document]
danethurber: so yeah....weird huh. this is how i'm maintaining a reference to a prevProp value so that i can compare it later later with the new prop. the ref is used to store a mutable value across renders. 

here is the react docs on the `usePrevious` hook -> https://reactjs.org/docs/hooks-faq.html#how-to-get-the-previous-props-or-state

and docs on using refs for more that dom references -> https://reactjs.org/docs/hooks-faq.html#is-there-something-like-instance-variables

🤷‍♂ thoughts?

packages/actionmenu/src/react/index.js
```diff
@@ -193,104 +187,106 @@ const calcNestedMenuPosition = (menuWidth, origin) =>
...
+
+  const itemEl = React.useRef()
+  const [isNestedRendered, setIsNestedRendered] = React.useState(false)
+
+  React.useEffect(() => {
+    if (isActive && props.shouldFocusOnMount) itemEl.current.focus()
+  }, [isActive, props.shouldFocusOnMount])
+
+  React.useEffect(() => {
+    if (!prevIsActive && isActive && !isNestedRendered) {
```
danethurber: comparing old vs new prop values

packages/actionmenu/src/react/index.js
danethurber: i can convert to non-arrow functions. i'm a creature or habit 😬

packages/position/src/js/__specs__/index.spec.js
```diff
@@ -1,11 +1,13 @@
 import { cleanup, render, waitForElement } from 'react-testing-library'
 import React from 'react'
 
-import { above, below, leftOf, rightOf } from '../index.js'
+import * as positionFns from '../index.js'
```
danethurber: there was starting to be a lot of imports here so i collapsed it down. you'll see this is a couple places

packages/position/src/react/__stories__/index.story.js
```diff
@@ -34,36 +34,44 @@ Box.propTypes = {
...
+
   componentDidMount() {
-    const { props } = this
-    if (this.boxEl && this.tooltipRef.current)
-      this.setState({
-        styles: position[props.styleFunction](this.boxEl).styleFor(
-          this.tooltipRef.current
-        )
-      })
+    if (!this.boxEl || !this.tooltipRef.current) return
```
danethurber: trying to make things for readable

packages/position/src/react/__stories__/index.story.js
```diff
@@ -72,22 +80,25 @@ class JsStory extends React.Component {
...
+  positionType: PropTypes.oneOf(Object.keys(positionFns)).isRequired
+}
+
 const jsStory = storiesOf('js', module)
-;['above', 'rightOf', 'below', 'leftOf'].forEach(styleFunction =>
-  jsStory.add(styleFunction, _ => {
-    return <JsStory styleFunction={styleFunction} />
-  })
+Object.keys(positionFns).forEach(pos =>
+  jsStory.add(pos, _ => <JsStory positionType={pos} />)
```
danethurber: changed a prop name. it's just the string that is passed an not an actual fn

packages/position/src/react/__stories__/index.story.js
```diff
@@ -99,18 +110,15 @@ reactStory.add('test.skip inNode', () => {
...
+    React.useEffect(() => {
+      setNode(portal.current)
+    }, [portal])
 
     return (
       <React.Fragment>
         <div
           style={{
-            position: 'relative',
+            positionFns: 'relative',
```
danethurber: oops. will fix

packages/position/src/react/index.js
```diff
@@ -30,13 +50,10 @@ function Position(props) {
...
   const [style, setStyle] = React.useState({ position: 'absolute' })
 
-  React.useLayoutEffect(
-    () => {
-      if (props.when && targetRef.current && elRef.current)
-        setStyle(props.position(targetRef.current).styleFor(elRef.current))
-    },
-    [props.when, targetRef, elRef, props]
-  )
+  React.useLayoutEffect(() => {
```
danethurber: prettier 🤷‍♂

packages/position/src/react/__stories__/index.story.js
```diff
@@ -99,18 +110,15 @@ reactStory.add('test.skip inNode', () => {
     const portal = React.useRef()
     const [node, setNode] = React.useState(portal.current)
 
-    React.useEffect(
-      () => {
-        setNode(portal.current)
-      },
-      [portal]
-    )
+    React.useEffect(() => {
```
danethurber: prettier strikes again

PR #576, opened 2019-07-09
banner: react 16

jaketrent:
description yo

### DIFF
--- a/packages/banner/src/react/index.js
+++ b/packages/banner/src/react/index.js
@@ -1,10 +1,10 @@
 import * as glamor from 'glamor'
-import Icon from '@pluralsight/ps-design-system-icon/react'
+import Icon from '@pluralsight/ps-design-system-icon/react.js'
 import PropTypes from 'prop-types'
 import React from 'react'
 
-import css from '../css'
-import * as vars from '../vars'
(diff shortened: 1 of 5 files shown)


packages/banner/src/react/index.js
```diff
@@ -1,44 +1,36 @@
...
-      <button {...styles.dismiss(props)} onClick={props.onClick}>
-        <Icon id={Icon.ids.close} />
-      </button>
-    )}
-  </div>
-)
+const Banner = React.forwardRef((props, ref) => {
+  const { onClick, ...rest } = props
+  return (
+    <div {...styles.banner(props)} {...filterReactProps(rest)} ref={ref}>
```
jaketrent: what should the ref go to?

root el?

main interactive element?

jaketrent: Had the thought when I wrote this ^^ and now when I'm in Breadcrumb, which was sending its ...props spread not to the root el like is usual but to the child Button

danethurber: i think we should pass the props and ref to the same dom element...usually the root one - but some components might be different 🤷‍♂ 

some components we might need to whip into shape to build this new world order

PR #577, opened 2019-07-09
breadcrumb: react 16

jaketrent:
desc yo

### DIFF
--- a/packages/breadcrumb/src/react/index.js
+++ b/packages/breadcrumb/src/react/index.js
@@ -1,5 +1,6 @@
 import * as glamor from 'glamor'
 import Button from '@pluralsight/ps-design-system-button/react.js'
+import filterReactProps from '@pluralsight/ps-design-system-filter-react-props'
 import Icon from '@pluralsight/ps-design-system-icon/react.js'
 import PropTypes from 'prop-types'
 import React from 'react'
@@ -11,14 +12,9 @@ const styles = {
 }
 
 const Breadcrumb = React.forwardRef((props, ref) => {
-  const breadcrumbProps = {
-    ...styles.breadcrumb(props),
-    ...(props.style ? { style: props.style } : null),
-    ...(props.className ? { className: props.className } : null)
-  }
-
+  const { onClick, ...rest } = props
   return (
-    <div {...breadcrumbProps}>
+    <div {...filterReactProps(rest)} {...styles.breadcrumb(props)}>
       <Button
         {...props}
         ref={ref}
@@ -33,6 +29,7 @@ const Breadcrumb = React.forwardRef((props, ref) => {
 Breadcrumb.propTypes = {
   className: PropTypes.string,
   disabled: PropTypes.bool,
+  onClick: PropTypes.func,
   style: PropTypes.object
 }
 Breadcrumb.defaultProps = {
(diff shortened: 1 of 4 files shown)


danethurber:
<table>
  <thead>
    <tr>
      <th width="50"></th>
      <th width="100%" data-danger-table="true">Warnings</th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>:pencil2: Please add a description.</td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 275788707ce97465189916810c793d988652749e
</p>

packages/breadcrumb/src/react/index.js
```diff
@@ -1,24 +1,20 @@
...
-  const breadcrumbProps = {
-    ...styles.breadcrumb(props),
-    ...(props.style ? { style: props.style } : null),
-    ...(props.className ? { className: props.className } : null)
-  }
-
+  const { onClick, ...rest } = props
   return (
-    <div {...breadcrumbProps}>
+    <div {...filterReactProps(rest)} {...styles.breadcrumb(props)}>
```
jaketrent: line 17 and 19 in combo concern me.  

"all components shall receive any html-legal props" is an easy-to-say rule.  But I'm having a hard time doing it in a useful way.  

Before, we had className and style going to root.  Then all other props going to the child button.  

But what props will be sent?

I think className and style should always go to root.  Because that's the most flexible (still terrible) way to override styles on a component with multiple elements.

So maybe I should have put filter on the button??

danethurber: hmm...getting tricky. i think props going to the root sounds like the best approach. events would hopefully bubble up from the button. the ref i would probably say should be on the root as well...but that's going to break what we have currently so i would leave it for now. 

it would be interesting to see why users would need to add attrs to the button or get the reference - and if becomes something we want to support I think we could do it better with a compound component

packages/breadcrumb/src/react/index.js
```diff
@@ -1,24 +1,20 @@
...
       <Button
         {...props}
```
danethurber: you've got props on two elements

danethurber: you're spreading props here and on the root component. i think you want to remove this line

packages/breadcrumb/src/react/index.js
```diff
@@ -1,24 +1,20 @@
...
 }
 
 const Breadcrumb = React.forwardRef((props, ref) => {
-  const breadcrumbProps = {
-    ...styles.breadcrumb(props),
-    ...(props.style ? { style: props.style } : null),
-    ...(props.className ? { className: props.className } : null)
-  }
-
+  const { onClick, ...rest } = props
```
danethurber: i don't think you want to do this...it's removing the click handler from being used

packages/breadcrumb/src/react/index.js
```diff
@@ -1,38 +1,43 @@
...
+  const { disabled, href, loading, onClick, ...rest } = props
   return (
-    <div {...breadcrumbProps}>
+    <div {...filterReactProps(rest)} {...styles.breadcrumb(props)}>
       <Button
-        {...props}
-        ref={ref}
-        icon={<Icon id={Icon.ids.caretLeft} />}
         appearance={Button.appearances.flat}
+        href={href}
```
jaketrent: Pushing Button-bound props through to button explicitly.  Spreading everything else onto root.

Issue #590, opened 2019-07-19
dropdown: support ability to reset to initial value


jaketrent:
#Enhancement Request

### Desired behavior

See initial value
Make a selection
Press clear button
Set back to initial value
As expressed here: https://pluralsight.slack.com/archives/C70HPSJPP/p1563467706088500?thread_ts=1563378818.076600&cid=C70HPSJPP

### Today's behavior

Can't reset internal selection

### The Value Add

Support the reset/clear use case

### The Tradeoffs

Sweet, sweet time.

jaketrent:
Shipped.

PR #591, opened 2019-07-19
actionmenu: add onSelect and value props

jaketrent:
Related: #590 

To support Dropdown.  Also, I think people will end up preferring a single handler like this in the API even when just using ActionMenu.

### DIFF
--- a/packages/actionmenu/src/react/index.js
+++ b/packages/actionmenu/src/react/index.js
@@ -132,6 +132,10 @@ const ActionMenu = React.forwardRef((props, forwardedRef) => {
     else navigate(evt, direction)
   }
 
+  function handleSelect(evt, value, label) {
+    if (typeof props.onSelect === 'function') props.onSelect(evt, value, label)
+  }
+
   return (
     <Menu onKeyDown={handleKeyDown} ref={ref} role="menu" {...props}>
       {React.Children.map(props.children, (child, i) =>
@@ -144,6 +148,7 @@ const ActionMenu = React.forwardRef((props, forwardedRef) => {
           _onDividerFocus: handleDividerFocus,
           _onItemFocus: focusItemAtIndex,
           _onMouseOver: focusItemAtIndexWithMouse,
+          _onSelect: handleSelect,
           _origin: props.origin
         })
       )}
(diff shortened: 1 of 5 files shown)


packages/actionmenu/src/react/index.js
```diff
@@ -144,6 +148,7 @@ const ActionMenu = React.forwardRef((props, forwardedRef) => {
           _onDividerFocus: handleDividerFocus,
           _onItemFocus: focusItemAtIndex,
           _onMouseOver: focusItemAtIndexWithMouse,
+          _onSelect: handleSelect,
```
danethurber: the native react `select` uses `onChange`. is that something we want to mimic?

packages/site/pages/components/actionmenu.js
```diff
@@ -60,17 +68,19 @@ function InAppExample(props) {
             when={isOpen}
             show={
               <div>
-                <ActionMenu origin={ActionMenu.origins.topLeft}>
+                <ActionMenu
+                  origin={ActionMenu.origins.topLeft}
+                  onSelect={handleSelect}
+                  whoa="adsf"
```
danethurber: did you mean to leave this in here

packages/actionmenu/src/react/index.js
danethurber: on wait. this is the actionmenu and not the dropdown. i think onSelect makes sense 👍

packages/site/pages/components/actionmenu.js
jaketrent: lol, no

packages/actionmenu/src/react/index.js
jaketrent: Yeah, it's a great point.  

I'm tempted to match onChange instead of introduce new verbiage (onSelect, as written thus far)

But then, as you point out, this isn't a dropdown.  We're not really "selecting" anything.  There's no concept of "a selection" to the user.  In the same way, there's no "change" to a new selection.

So, I guess both are flawed.  

Dropdown is going to utilize onSelect/onChange, whatever we choose.  So, there is a relationship there that I'll build.  So, if they matched, that might make a little sense.

Any further preferences or thoughts on your end?

Issue #593, opened 2019-07-26
Carousel: Context menus cut off by carousel container


nttibbetts:
### Expected behavior
When opening an action menu on a card in a carousel, the menu should not be hidden by the carousel container

### Actual behavior
The action menu is cut off at the edges of the carousel container

![Screen Shot 2019-07-26 at 10 16 09 AM](https://user-images.githubusercontent.com/1699510/61969090-ae4a1280-afa7-11e9-922f-dec2d1865d13.png)

We were just experimenting with replacing our carousel with the design system carousel but didn't commit the changes. If necessary we could commit this to a branch, but it'd likely be easier to mock it up as it's all design system components.

### Steps to repro

1. Add a card component with an action menu in its action bar with about 5 items to a carousel
2. Show the action menu

### Related packages

- carousel@1.1.1
- card@9.0.7
- actionmenu@4.1.7

### Environment

- OS - OSX 10.14.5
- Browser version - Chrome v75

jaketrent:
Triage: We can attempt the solution below.  React.createPortal is also a great solution for this.

Possible solution: thinking that overflow:hidden needed for transition time only.

nttibbetts:
Doing `overflow: hidden` only during transition was the solution we landed on for our custom built carousel component. The only downside was it was a touch tricky to nail down the positioning logic during transitioning vs after transitioning.

This was after considering using React Portals but this created the problem of needing to include a significant amount more of ReactDOM and nearly tripled our bundle size (we have a fairly basic experience so not much code/dependencies to start with). There was also the problem of needing to do a bit of CSS math to position it correctly, but @danethurber has pointed out the Position component that might solve that problem if you went with a portal based solution.

popthestack:
We had the same problem with menus getting cut off. I tried `React.createPortal` and was able to get the menu working, but it broke accessibility. For example, using the tab key to navigate worked until needing to dismiss the menu. Once dismissed the focus was at the bottom of the page where the portal element had been. I didn't look into any solutions, but instead switched to removing `overflow: hidden` when a menu is activated and putting it back when it is unmounted.

[...]

PR #606, opened 2019-08-05
carousel: only hide overflow during transition

danethurber:
resolves: #593


packages/carousel/src/react/index.js
```diff
@@ -17,13 +17,19 @@ import useResizeObserver from './use-resize-observer.js'
 import useSwipe from './use-swipe.js'
 import useUniqueId from './use-unique-id.js'
 
+const TRANSITION_DURATION_MS = 500
```
jaketrent: Can use core.motion.speedXSlow, like you do in the css.

packages/carousel/src/react/index.js
```diff
@@ -17,13 +17,19 @@ import useResizeObserver from './use-resize-observer.js'
...
+
 const styles = {
-  carousel: ({ ready }) =>
+  carousel: (props, { ready }) =>
```
jaketrent: why add props here?  (not used)

packages/carousel/src/react/index.js
```diff
@@ -126,6 +132,24 @@ function Item(props) {
...
+
+  React.useEffect(() => {
+    let timer
+    const starting =
+      prevActivePage !== undefined && prevActivePage !== context.activePage
+
+    if (starting) {
+      setTransitioning(true)
+    } else {
+      timer = setTimeout(() => {
```
jaketrent: ah, there it is.  bunch o these in my Tab impl. :)

packages/carousel/src/react/index.js
```diff
@@ -234,3 +258,13 @@ function usePager(pageCount, additionalSideEffectTriggers = []) {
     ref
   }
 }
+
+function usePrevious(value) {
```
jaketrent: Such a weird React 16 pattern.  Feels like a shoe-horned ability of the ref feature.

danethurber: yeah. it's because there is no hook method for componentWillReceiveNewProps so this is the way to store the prev prop(or state/context/whatever)

packages/carousel/src/react/index.js
danethurber: props as the first arg(or second if there is theme support 😬 ) seems to be the loose convention we're following so i thought i'd add the props as the first arg

PR #609, opened 2019-08-07
theme as peer dep

### LINKED ISSUE #604: Theme dep management
In dev only, two theme component versions, problem with light theme.

Maybe move peerDep.

Resolved by bundler in stage/prod.

Theory: react 16 context creates multiple instances if multiple theme packages exist.  Inner component sees the other context that is not rendered and defaults (falsely) to dark.

danethurber:
### What You're Solving

- resolves #604 

### Design Decisions

- packages that need theming now have the theme package as a peerDependency
- added theme devDep to packages for easier to find version requirements from lerna versioning
- updated CRA example to test package version publishing and resolution


packages/site/pages/install.js
```diff
@@ -25,7 +25,13 @@ Use of the Core or Components requires the existence of the Design System normal
 npm install @pluralsight/ps-design-system-normalize
 \`\`\`
 
-Include this vanilla CSS in your application in a method appropriate for your project. For usage options, follow the [Core Usage docs](/core/usage) patterns.
```
jaketrent: Why remove this?  Still seems valid (ie, all components still require normalize)

scripts/create-component/package-template/package.json
```diff
@@ -22,6 +22,7 @@
   "dependencies": {{dependencies}},
   "peerDependencies": {
     "@pluralsight/ps-design-system-normalize": "*",
+    "@pluralsight/ps-design-system-theme": "*",
```
jaketrent: Do these peerDeps w/ versions of "*" get published that way, or does it publish with a concrete (eg, "3.0.4") version?

packages/site/pages/install.js
```diff
@@ -25,7 +25,13 @@ Use of the Core or Components requires the existence of the Design System normal
...
+## Theming
+
+Most components support multiple themes and have a dependency on theme package. Install with:
```
jaketrent: I would modify this.  We've tried hard to make sure that *all* components support theming.  Whether there's an actual diff between themes is less important.  It's the perception problem that I don't want to reawaken.

Maybe:

```
All components support theming and have a peer dependency on the Theme package. Install with:
```

I would also *move* this down to the "Use Components" section.

packages/site/pages/install.js
danethurber: oops

scripts/create-component/package-template/package.json
danethurber: doesn't look like lerna does the versioning for us 😢 

i'll go through and update

packages/viewtoggle/package.json
```diff
@@ -26,7 +26,7 @@
     "prop-types": "^15.7.2"
   },
   "peerDependencies": {
-    "@pluralsight/ps-design-system-normalize": "*",
+    "@pluralsight/ps-design-system-normalize": "v3.1.4",
```
jaketrent: I'm pretty sure these numbers don't need and can't have the `v` prefix.  Do you know otherwise?

danethurber: oops. it should be a ^

scripts/create-component/package-template/package.json
```diff
@@ -21,7 +21,8 @@
   },
   "dependencies": {{dependencies}},
   "peerDependencies": {
-    "@pluralsight/ps-design-system-normalize": "*",
+    "@pluralsight/ps-design-system-normalize": "v3.1.4",
+    "@pluralsight/ps-design-system-theme": "v3.0.2",
```
jaketrent: rm v

packages/theme/package.json
```diff
@@ -20,7 +20,7 @@
     "prop-types": "^15.7.2"
   },
   "peerDependencies": {
-    "@pluralsight/ps-design-system-normalize": "*",
+    "@pluralsight/ps-design-system-normalize": "v3.1.4",
```
jaketrent: rm v

PR #615, opened 2019-08-13
site: upgrade to next9 and react16 and style conventions

jaketrent:
asdf


packages/link/src/react/index.js
```diff
@@ -1,31 +1,39 @@
...
-    ':hover': css['.psds-link:hover'],
-    ...(({ appearance }) => css[`.psds-link--appearance-${appearance}`])(props)
+  css(
+    {
+      ':hover': stylesheet['.psds-link:hover']
+    },
+    stylesheet[`.psds-link--appearance-${props.appearance}`]
+  )
+
+const Link = React.forwardRef((props, forwardedRef) => {
```
danethurber: does this count as our react16 upgrade for Link? let's mark it off the list !

packages/site/package.json
```diff
@@ -4,77 +4,66 @@
...
+    "@pluralsight/ps-design-system-textarea": "*",
+    "@pluralsight/ps-design-system-textinput": "*",
+    "@pluralsight/ps-design-system-theme": "*",
+    "@pluralsight/ps-design-system-tooltip": "*",
+    "@pluralsight/ps-design-system-util": "*",
+    "@pluralsight/ps-design-system-viewtoggle": "*",
+    "prop-types": "^15.7.2"
+  },
+  "dependencies": {
     "babel-preset-env": "^1.6.0",
```
danethurber: do we still need these babel presets? we have removed our `.babelrc` file from the package root so i don't think that we'd be doing any custom babel config.

there is also a `next/babel` preset that is suggested in the docs but not sure what it does.

packages/site/pages/components/position-portal-example.js
```diff
@@ -1,20 +1,17 @@
...
   const [node, setNode] = React.useState(portal.current)
-  React.useEffect(
-    () => {
-      if (portal.current) setNode(portal.current)
-    },
-    [portal]
-  )
+  React.useEffect(() => {
+    if (portal.current) setNode(portal.current)
+  }, [portal])
```
danethurber: should this be `portal.current`. asking for a friend

packages/site/pages/components/table.js
```diff
@@ -1,11 +1,11 @@
-import Avatar from '@pluralsight/ps-design-system-avatar/react'
+import Avatar from '@pluralsight/ps-design-system-avatar/react.js'
 import core from '@pluralsight/ps-design-system-core'
-import Drawer from '@pluralsight/ps-design-system-drawer/react'
+import Drawer from '@pluralsight/ps-design-system-drawer/react.js'
 import PT from 'prop-types'
```
danethurber: just a comment. maybe we should change our `PropTypes` component to be `PropTypesTable` or something. i just don't love the `PT` name. an issue for another day

packages/site/src/ui/index.js
```diff
@@ -1,25 +1,49 @@
-export ChangeLog from './change-log'
```
danethurber: should we config babel to allow this type of export?

packages/link/src/react/index.js
jaketrent: yes, I'll mark it.

packages/site/package.json
jaketrent: babel-standalone uses plugins these inside the eval for code inside the live code component.  It's not for next per se.  It's used in the browser.

packages/site/pages/components/position-portal-example.js
jaketrent: I actually have no idea.  I just chose one and hoped for the best.  That's yeehaa-driven-development.

packages/site/pages/components/table.js
jaketrent: Yeah, there's an odd naming clash there.  I don't love `PropTypes` either.  It's not what you anticipate it to be in the `site` package.

packages/site/src/ui/index.js
jaketrent: I chose not to just for the sake of writing more "javascript today"-style code.  It also meant less configuration.

PR #616, opened 2019-08-14
fix(viewtoggle): error that happens when the active option is removed

danethurber:
### What You're Solving

- https://pluralsight.slack.com/archives/C70HPSJPP/p1565630361034500

### DIFF
--- a/packages/viewtoggle/src/react/index.js
+++ b/packages/viewtoggle/src/react/index.js
@@ -40,6 +40,7 @@ const ViewToggle = React.forwardRef(({ onSelect, ...props }, forwardedRef) => {
 
   function renderActivePill() {
     const activeEl = React.Children.toArray(props.children)[activeIndex]
+    if (!activeEl) return null
 
     let activePillStyle = {}
     if (hasRenderedOnce && ref.current) {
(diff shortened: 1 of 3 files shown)


packages/viewtoggle/src/react/index.js
```diff
@@ -40,6 +40,7 @@ const ViewToggle = React.forwardRef(({ onSelect, ...props }, forwardedRef) => {
 
   function renderActivePill() {
     const activeEl = React.Children.toArray(props.children)[activeIndex]
+    if (!activeEl) return null
```
jaketrent: Why doesn't the removal of an option (change in `props.children`) retrigger line 33 then 31 above?

If it did, `findActiveIndex` will default to index 0, where the active pill will sit.  

But in this solution, we simply don't get to see a pill.  Would that ui state even make sense?

danethurber: it does trigger an update but the rendering happens before the useEffect.

jaketrent: Just a followup to the 2nd question here: 

> But in this solution, we simply don't get to see a pill. Would that ui state even make sense?

First of all, is my observation accurate?  If so, that doesn't seem right to me.  Pulling code to try it...

jaketrent: Looks like when active el is removed, index 0 is reset, active pill bg still shown.  👍

PR #621, opened 2019-08-15
dialog/focusmanager: update to react16 and hooks

danethurber:


.gitignore
```diff
@@ -43,3 +43,6 @@ jspm_packages
 .env*
 
 .cache
+
+# workspace config
+.vim
```
jaketrent: I would prefer personal dev environment ignores to happen at a system level, outside the project.  See  https://jaketrent.com/post/gitignore-system-files/

Does that seem reasonable?

packages/dialog/src/react/index.js
```diff
@@ -1,174 +1,151 @@
...
+  children: PropTypes.node,
   disableCloseButton: PropTypes.bool,
   disableCloseOnEscape: PropTypes.bool,
+  disableCloseOnOverlayClick: PropTypes.bool,
   disableFocusOnMount: PropTypes.bool,
   modal: PropTypes.bool,
   onClose: PropTypes.func,
-  style: PropTypes.object,
-  tailPosition: PropTypes.oneOf(Object.keys(vars.tailPositions))
+  tailPosition: PropTypes.oneOf(Object.keys(Dialog.tailPositions))
```
jaketrent: You want values instead of keys, right?

.gitignore
danethurber: good idea. i'll move it

packages/dialog/src/react/index.js
danethurber: yes....will update

PR #623, opened 2019-08-20
refactor(site): add guidelines for error messages to voice & tone page

sweeting:
Added guidelines and examples for writing error messages to the Voice & Tone page.


jaketrent:
Maybe one way to break it up and to put the dos/donts closers to the description would be to interleave the 1 set of dos/donts under each of the numbered paragraphs.

PR #627, opened 2019-08-22
fix(avatar): edgecase with single char name

danethurber:
### What You're Solving

- edgecase with single char name

fixes bug reported here -> https://pluralsight.slack.com/archives/C70HPSJPP/p1566424721012800?thread_ts=1566417670.011700&cid=C70HPSJPP

### DIFF
--- a/packages/avatar/src/js/index.js
+++ b/packages/avatar/src/js/index.js
@@ -8,7 +8,9 @@ export const getInitials = fullname => {
   const names = fullname.trim().split(' ')
 
   const first = names[0][0]
-  const second = names.length === 1 ? names[0][1] : names[names.length - 1][0]
+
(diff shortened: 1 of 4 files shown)


packages/avatar/src/js/index.js
```diff
@@ -8,7 +8,9 @@ export const getInitials = fullname => {
   const names = fullname.trim().split(' ')
 
   const first = names[0][0]
-  const second = names.length === 1 ? names[0][1] : names[names.length - 1][0]
+
+  let second = names.length === 1 ? names[0][1] : names[names.length - 1][0]
+  if (!second) second = ''
```
jaketrent: If you wanted to put it in one expression, you could:

```
let second = names.length === 1 
  ? names[0][1] 
  : names.length > 1 
    ? names[names.length - 1][0]
    : ''
```

PR #628, opened 2019-08-22
Feature/avatar error callback

danethurber:
### What You're Solving

- provide prop/function when image state changes

cc @anicholson-gitprime

### DIFF
--- a/packages/avatar/src/react/index.js
+++ b/packages/avatar/src/react/index.js
@@ -19,13 +19,23 @@ const styles = {
 
 const Avatar = React.forwardRef((props, ref) => {
   const [imageState, setImageState] = React.useState('loading')
+  const prevImageState = usePrevious(imageState)
+
+  React.useEffect(() => {
+    if (!props.onImageStateChange) return
+
+    const isNewImageState = prevImageState !== imageState
+    if (!isNewImageState) return
+
+    props.onImageStateChange(imageState, prevImageState)
+  }, [prevImageState, imageState, props])
 
   function handleImageLoadSuccess(evt) {
     const isFallbackPixel = evt.target.naturalWidth === 1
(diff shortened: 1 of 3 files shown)


packages/site/pages/components/avatar.js
```diff
@@ -36,6 +36,13 @@ export default _ => (
             null,
             'used to generate intials as a fallback image'
           ]),
+          PropTypes.row([
```
jaketrent: What does having this (exposing this) solve for users?

packages/avatar/src/react/index.js
```diff
@@ -19,13 +19,23 @@ const styles = {
 
 const Avatar = React.forwardRef((props, ref) => {
   const [imageState, setImageState] = React.useState('loading')
```
jaketrent: Now that these states for "loading", "error", and other? are being exposed, I suggest that we export them as a named enum.  This is something that would usually go in vars.  

/me. Now that we do js/ a lot, I wonder if we don't drop "vars/" for "js/" these days.

packages/site/pages/components/avatar.js
anicholson-gitprime: If you pass an avatar url that does not resolve to an avatar, and no name, the empty state is an empty circle. For my use case, a more sensible empty state would be to not render the avatar component at all. However, I don't know whether the Avatar will resolve the avatar url passed in. The solution I suggested was to expose a callback so that the client can know when the avatar does not resolve. Another option could be to take a render prop which Avatar calls to render the empty state. Yet another option would be to accept a boolean which indicates whether we want to render an `ident` gravatar in case the gravatar hash does not resolve to a real gravatar.

I put in a stop-gap solution to pass in a string that won't be undefined for the `name` prop for my use case, so this isn't critical for me at this time.

PR #640, opened 2019-08-28
refactor(radio): upgrade to react16 patterns

danethurber:
BREAKING CHANGE: using forwardRef. innerRef is no longer supported


danethurber:
@jaketrent good change. it has been fixed

PR #642, opened 2019-08-28
site: restore http-server needed for runtime

jaketrent:
was the reason for octopus breaking


jaketrent:
removed dane because he's on vacation.

PR #646, opened 2019-08-30
textarea: alternate width flex solution

jaketrent:
After discovering diff textarea requirements than I had in mind the other day.

### DIFF
--- a/packages/form/src/react/vertical-layout.js
+++ b/packages/form/src/react/vertical-layout.js
@@ -14,7 +14,7 @@ const VerticalLayout = (props, context) => (
     {React.Children.map(props.children, (child, i) => (
       <div {...styles.child(props)}>
         {React.cloneElement(child, {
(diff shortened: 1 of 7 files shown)


packages/form/src/react/vertical-layout.js
```diff
@@ -14,7 +14,7 @@ const VerticalLayout = (props, context) => (
     {React.Children.map(props.children, (child, i) => (
       <div {...styles.child(props)}>
         {React.cloneElement(child, {
-          style: { ...child.props.style, width: '100%' }
+          style: { ...child.props.style, width: '100%', maxWidth: 'none' }
```
jaketrent: Note that this addition to Form.VerticalLayout is required now.  Makes sense to have been there before.  Now it's required to support this use case.

EdwardIrby:
Looks like snapshots will need to be updated.

jaketrent:
Was merged manually.

PR #647, opened 2019-08-30
[WIP] carousel: re-add overflow hidden

jaketrent:
Overflow hidden is back.  Added a story in Carousel to support it.  Added forwardRef to Card to support that.  

*Problem*: Now I'm in site, adding docs for how one might do this IRL.  But, it's slow!  I'd love your thoughts on this, @EdwardIrby and @danethurber .  A couple things to look at:

1. The carousel paging seems slower than the others.  I wonder why that would be.
2. The "more" context menu click takes a long! time.  It also scrolls me to the top of the page on first click.

I'll keep looking at this later, but I thought I might make this available in case you had ideas.  Thanks!

### DIFF
--- a/packages/carousel/src/react/__stories__/index.story.js
+++ b/packages/carousel/src/react/__stories__/index.story.js
@@ -1,11 +1,13 @@
 import { storiesOf } from '@storybook/react'
 
-import * as glamor from 'glamor'
+import { css } from 'glamor'
 import React, { Fragment } from 'react'
 import PropTypes from 'prop-types'
 
-import ActionMenu from '@pluralsight/ps-design-system-actionmenu/react'
-import Card from '@pluralsight/ps-design-system-card/react'
+import ActionMenu from '@pluralsight/ps-design-system-actionmenu/react.js'
+import Card from '@pluralsight/ps-design-system-card/react.js'
+import Icon from '@pluralsight/ps-design-system-icon/react.js'
+import { BelowRight } from '@pluralsight/ps-design-system-position/react.js'
 
 import Carousel from '../index.js'
 
@@ -41,7 +43,7 @@ const longStringsMetaData = [
 
 const MockItem = props => (
   <div
-    {...glamor.css({
+    {...css({
       alignItems: 'center',
       background: 'pink',
       display: 'flex',
@@ -89,7 +91,7 @@ storiesOf('Carousel/items', module)
           </Carousel>
 
           <div
-            {...glamor.css({
+            {...css({
               alignItems: 'center',
               display: 'flex',
               justifyContent: 'center',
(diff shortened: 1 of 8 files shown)


packages/carousel/src/react/__stories__/index.story.js
```diff
@@ -181,20 +183,77 @@ Object.values(Carousel.sizes).forEach(size => {
...
+                  show={
+                    <ActionMenu>
+                      {new Array(8).fill(null).map((_, index) => (
+                        <ActionMenu.Item key={index}>
+                          item: {index}
+                        </ActionMenu.Item>
+                      ))}
+                    </ActionMenu>
+                  }
+                  key="a"
```
danethurber: this should be unique

packages/carousel/src/react/index.js
```diff
@@ -179,24 +175,6 @@ function Instructions(props) {
 
 const Pages = React.forwardRef((props, ref) => {
   const context = React.useContext(CarouselContext)
-  const prevActivePage = usePrevious(context.activePage)
-  const [transitioning, setTransitioning] = React.useState(false)
```
danethurber: we shouldn't need the transition group anymore. we could remove that dependency

packages/carousel/src/react/__stories__/index.story.js
```diff
@@ -181,20 +183,77 @@ Object.values(Carousel.sizes).forEach(size => {
...
+                >
+                  <Card.Action
+                    title="asdf"
+                    icon={<Icon id={Icon.ids.more} />}
+                    key="asdf"
```
danethurber: also needs to be a unique key

packages/carousel/src/react/__stories__/index.story.js
jaketrent: unique w/in the list, right?  And it is, I think.

packages/carousel/src/react/index.js
jaketrent: Hmm, wait, I still see `Transition` used in `carousel/src/react/index.js`, line 217.  Isn't that still needed to support the page advance slide animation?

packages/carousel/src/react/__stories__/index.story.js
jaketrent: actually, reading this again, I think we can remove `key` here, since it's not an array.

packages/carousel/src/react/index.js
danethurber: i added the Transition so i could hide items after the transition. with the overflow we shouldn't need to use Transition.

packages/position/src/react/index.js
```diff
@@ -55,13 +55,15 @@ function Position(props) {
...
 
-  const show = React.cloneElement(props.show, {
-    ref: elRef,
-    style: {
-      ...child.props.style,
-      ...style
-    }
-  })
+  const show =
+    props.when &&
```
jaketrent: This was the key for initial render performance.

jaketrent:
2 perf cases fixed.  1 remaining, with an oddity:

in site, click card action, is:

1) slow
2) scrolling to top

fix for storybook slowness wasn't the only thing required here.  Scrolling is totally left field.  I have no idea where that's coming from.

jaketrent:
A number of things I've found and fixed:

- Site rendering of Carousel after ActionMenu click was because top-level state was changing and needed to re-render the whole Carousel.  This was solved by storing ActionMenu toggle state per element.  Now vdom updates happen lower (inside the single array element).  As a side-effect of storing state lower, the upper component now doesn't control menu toggling.  Thus, the feature in the example of closing all ActionMenus when paging the Carousel needed to be re-implemented.  Looking at the ActionMenu, turns out we had an Overlay, but it wasn't used anywhere internally (but also exported).  I change ActionMenu to render Overlay when an onClose prop is provided.  The prop is a good conditional, because otherwise a "close the action menu" trigger from the ActionMenu can't notify any external entity controlling the actual rendering (because it's not controlled internally).
- Scrolling to the top of the page on ActionMenu toggling was happening because of a combo of 2 things: Position paints the shown element on the screen, then has refs, then measures it, then positions it.  ActionMenu focuses itself when rendered.  Thus rendering initially at 0,0, the focus would scroll to the top.  Now, focus on ActionMenu happens on nextTick.  It doesn't impact UX, afaict, and it supports this Position combo use case.

jaketrent:
@danethurber and/or @EdwardIrby This is ready for a re-review.

Who knew such a "small" change would have such cascading implications.  :) Not the first time, right?

PR #650, opened 2019-09-03
refactor(row): upgrade to functional components

EdwardIrby:


.gitignore
```diff
@@ -43,3 +43,6 @@ jspm_packages
 .env*
 
 .cache
+
+# ide
+.vscode
```
danethurber: i suggest you place this in you system level `~/.gitgnore` so you don't have to do it for every project

EdwardIrby: 👍🏿

PR #551, opened 2019-06-16
feat(tag): adds tag to small card for issue #549

Martskin:
## Instructions

- card: add tag to small card
- label: enhancement for issue #549 

### What You're Solving

- issue #549
- Removes conditional for tag display on card

### Design Decisions

I wasn't able to access the Figma link referenced in issue #549
https://www.figma.com/file/F9uJHgm4AOCWe5RouPFyeRSp/Small-Cards-and-Tags?node-id=0%3A1

### How to Verify

1. Launch storybook in `/packages/card` and view "sizes with tags" story
http://localhost:6006

Screen Shot 2019-06-16 at 8 36 22 AM

2. Run Design System doc site and view "Card" page
http://localhost:1337/components/card#tag

Screen Shot 2019-06-16 at 8 36 57 AM

PR #644, opened 2019-08-30
Feature/add tags small card

EdwardIrby:
## Instructions
See #551

PR #652, opened 2019-09-04
fix(card): adds tag to small card for issue

### LINKED ISSUE #549: Add tags to small card component
### Related packages
- care@10.0.1

## Enhancement Request
We would like to have the small card have the ability to display the tag on them similar to how the medium sized cards display the tag.

[https://www.figma.com/file/F9uJHgm4AOCWe5RouPFyeRSp/Small-Cards-and-Tags?node-id=0%3A1](url)

### The Value Add

Continued engagement on the homepage has time and time again shown that have a compact homepage leads to a higher retention rates.  Because of this we migrated the homepage away from the medium sized cards to the small sized card. In doing so we lost the added value of being able to differentiate content types like Interactive Courses and Projects. We feel identifying these content types if value to a users understanding that they are different from our regular course. 

### The Tradeoffs

We will definitely lose some of the image real estate, however user have let us know that those images have proved to be not helpful while selecting a piece of content to consume.

EdwardIrby:
fix #549 close #551

### DIFF
--- a/packages/card/src/react/__specs__/__snapshots__/storyshots.spec.js.snap
+++ b/packages/card/src/react/__specs__/__snapshots__/storyshots.spec.js.snap
@@ -3765,6 +3765,456 @@ exports[`Storyshots size small 1`] = `
 </div>
 `;
 
+exports[`Storyshots sizes with tags large 1`] = `
+<div
+  data-css-claeju=""
+>
+  <div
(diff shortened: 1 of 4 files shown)


EdwardIrby:
Feel confident this one won't break the build like #644 

close #644

Issue #424, opened 2019-01-03
typeahead/multiselect component


carolinapowers:
### Desired behavior

Projects content-tools need the ability to use the PS DS dropdown component to allow authors to enter tags that belong to a specific project. The author should be able to pick a tag name from the list and see a tag added to the input field as a result of the selection. As the author picks new tags, the new tags should be added to the input field. See example below:

![image](https://user-images.githubusercontent.com/10137641/50652451-8a17dd00-0f54-11e9-9702-5a2707909ce5.png)

You can see a live example of the desired behavior at the QA page "Ask a question" modal: https://app.pluralsight.com/questions-and-answers

![image](https://user-images.githubusercontent.com/10137641/50652878-d44d8e00-0f55-11e9-8260-1cd6796362bd.png)

The dropdown used above is a custom package called `'@pluralsight/mentors-searchable-dropdown'`

The Projects team currently uses a Vue.js module to accomplish such behavior, but we would like to start using the design system dropdown component.

Here is what our current content-tools application looks like:

![image](https://user-images.githubusercontent.com/10137641/50653546-dfa1b900-0f57-11e9-893b-9dc1e234f79f.png)

### Today's behavior

Today's behavior only allow one **text** to be shown at a time. 

When trying to implement this feature, there should be known that the tags need the ability to be clickable in order to delete tags added. The way that the dropdown is currently designed, any click on it will trigger the action menu to show or hide.  

### The Value Add

As right now we can already identify two teams that need such functionality, and we believe that other teams would need it as well.

jaredAnders:
Just came here to request this after seeing your team's demo today. Can confirm, we built something very similar for the Role IQ Manager experience and used Channels' implementation as a guideline. I can already think of another place where we're planning to use this. Would love to see something like this in the design system.

dborodyansky:
Design proposal for consideration:

https://www.figma.com/file/JImDLmaRbGWo6iWWMN1DL6nB/Select-extras?node-id=0%3A1

dborodyansky:
Updated design for reference:
https://www.figma.com/file/JImDLmaRbGWo6iWWMN1DL6nB/Select-extras?node-id=0%3A1

jaketrent:
Triage: design review fresh design

jaketrent:
Decide: 
- [ ] what name?  typeahead or other? 2 components?

jaketrent:
Triage: We will do this 2H 2019.  

Next steps:
[...]

PR #653, opened 2019-09-04
Typeahead

danethurber:
ref #424 

This is just the first pass at a typeahead. it is basic and doesn't support async loading of options - that will come later.

i had to update a few other components to make this work. an in-person pair review will probably help to show what has changed


danethurber:
i'm rethinking the api. initially i was going to create a single component to handle select/multiselect and typeahead. i think it's probably better to separate those and this is the new api i propose:

```jsx
<Typeahead
  appearance={}
  disabled
  error
  label=""
  loading
  name=""
  maxSuggestions={20}
  onChange={}
  onSearch={}
  placeholder=""
  subLabel=""
  value={}
>
  <Typeahead.Suggestion>first</Typeahead.Suggestion>
  <Typeahead.Suggestion>second</Typeahead.Suggestion>
  <Typeahead.Suggestion>third</Typeahead.Suggestion>
</Typeahead>
```

this well in essence be a TextInput with suggestions. we'll need to support basic string filter/fuzzy search as well as a async callback for filtering

packages/position/src/react/__stories__/index.story.js
```diff
@@ -1,139 +1,211 @@
...
-      above: Tooltip.tailPositions.bottomCenter,
-      aboveLeft: Tooltip.tailPositions.bottomLeft,
-      aboveRight: Tooltip.tailPositions.bottomRight,
-      rightOf: null,
-      below: Tooltip.tailPositions.topCenter,
-      belowLeft: Tooltip.tailPositions.topLeft,
-      belowRight: Tooltip.tailPositions.topRight,
-      leftOf: null
+storiesOf('Components | Position / portals', module).add(
+  'test.skip inNode',
```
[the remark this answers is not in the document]
jaketrent: Perhaps a similar scenario in the portal example in the docs site:  
Screen Shot 2019-09-23 at 1 17 35 PM

packages/textinput/src/react/index.js
```diff
@@ -1,55 +1,57 @@
...
+import Halo from '@pluralsight/ps-design-system-halo/react.js'
+import Icon from '@pluralsight/ps-design-system-icon/react.js'
+import { useTheme } from '@pluralsight/ps-design-system-theme/react.js'
+
+import stylesheet from '../css/index.js'
 import * as vars from '../vars/index.js'
 
 const styles = {
-  error: _ => glamor.css(css['.psds-text-input__error']),
+  error: _ => css(stylesheet['.psds-text-input__error']),
```
jaketrent: Nice catch!  I'm thinking we shouldn't have crossed these all off our list so quickly.  Seems like there were still some to style update.

packages/textinput/src/react/index.js
```diff
@@ -122,3 +126,34 @@ TextInput.iconAligns = vars.iconAligns
...
+}
+
+function isPlainObject(obj) {
+  return obj && !Array.isArray(obj) && typeof obj === 'object'
+}
+
+function useMultipleRefs(refs) {
+  const hasMultiple = isRefs(refs)
+
+  const prevFieldRef = hasMultiple ? refs['field'] : React.createRef()
```
jaketrent: `prev` as in "previous"?  I don't understand what "previous" means in this context.

packages/site/pages/components/textinput.js
```diff
@@ -72,13 +72,6 @@ export default _ => (
               horizontal icon placement (from <code>TextInput.iconAligns</code>)
             </span>
           ]),
```
jaketrent: Should the dual ref ability be documented?

packages/typeahead/src/react/__specs__/utils.spec.js
```diff
@@ -0,0 +1,106 @@
+import * as utils from '../utils.js'
+
+describe('utils', () => {
```
jaketrent: Starting to feel like there's a tipping point in the close future for putting these utils that have been dup'ed in several packages now into something reusable.

packages/typeahead/src/react/index.js
```diff
@@ -0,0 +1,257 @@
...
+  const inputRef = React.useRef()
+  React.useImperativeHandle(forwardedRef, () => inputRef.current)
+
+  const [controlled] = React.useState(typeof value !== 'undefined')
+  const [open, setOpen] = React.useState(false)
+
+  const [innerValue, setInnerValue] = React.useState(value || '')
+  const [searchTerm, setSearchTerm] = React.useState('')
+
+  React.useEffect(() => {
```
jaketrent: Please name the hook callbacks.

packages/typeahead/src/react/index.js
```diff
@@ -0,0 +1,257 @@
...
+
+    setOpen(false)
+  })
+
+  const suggestions = React.useMemo(() => {
+    const childArray = Children.toArray(children)
+    return childArray.map((child, index) => ({
+      comp: child,
+      index,
+      value: getSuggestionValue(child)
```
jaketrent: As we discussed, I think there's an opportunity to:

1) Support entity selection
2) Be consistent with the Dropdown API

By adding `value` as a prop to `Typeahead.Suggestion`.

packages/typeahead/src/react/index.js
```diff
@@ -0,0 +1,257 @@
...
+    PropTypes.arrayOf(elementOfType(Suggestion))
+  ]),
+  filterFn: PropTypes.func,
+  loading: PropTypes.bool,
+  onBlur: PropTypes.func,
+  onChange: PropTypes.func,
+  onFocus: PropTypes.func,
+  value: PropTypes.string,
+
+  // props used by underlying TextInput
```
jaketrent: Seems like these could be intermingled with all the rest of the proptypes. 🤷‍♂

packages/typeahead/src/react/index.js
```diff
@@ -0,0 +1,257 @@
...
+    const { current } = ref
+    document.body.appendChild(current)
+
+    return () => {
+      document.body.removeChild(current)
+    }
+  }, [])
+
+  return ref
+}
```
jaketrent: This file is potentially more readable if it's broken up into other files, like done with ActionMenu recently.  

That file was probably bigger.  But, navigation and structure discovery was helped by breaking it up.  But once I started it, I had a hard time stopping breaking things out. :)

packages/position/src/react/index.js
```diff
@@ -4,84 +4,152 @@ import ReactDOM from 'react-dom'
...
-        ...child.props.style,
-        ...style
-      }
-    })
+  const ref = React.useRef()
+  React.useImperativeHandle(forwardedRef, () => ref.current)
+
+  const child = React.Children.only(props.children)
+  const tetheredRef = React.useRef()
+  const tetheredEl = React.cloneElement(props.show, {
```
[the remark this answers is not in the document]
danethurber: naming things is hard :(

i changed it because the word `show` seemed like it should be boolean and was tripping me up. `tether` comes from similar libraries and code i've written in the past. i'm open to changing it to something else...

packages/site/pages/components/textinput.js
danethurber: yeah i should add that

packages/textinput/src/react/index.js
danethurber: yeah....i'm planning on doing a final audit/fix. soon hopefully 🤞

packages/textinput/src/react/index.js
danethurber: yeah that's confusing. it's the external/forwardeded ref. i'll change

packages/typeahead/src/react/__specs__/index.spec.js
```diff
@@ -0,0 +1,192 @@
...
+  })
+
+  it('forwards refs', () => {
+    const ref = React.createRef()
+    render(<Typeahead ref={ref} />)
+
+    expect(ref.current).not.toBeNull()
+  })
+
+  it('calls onFocus prop when input is focused', () => {
```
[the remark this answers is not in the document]
danethurber: good question. i'll have to ask. i though this was the right approach but you bring up a good point

packages/typeahead/src/react/__specs__/utils.spec.js
danethurber: 100%....once we get an esmodule build that's my hope

packages/typeahead/src/react/index.js
danethurber: agreed...i'll refactor

packages/position/src/react/index.js
```diff
@@ -4,84 +4,152 @@ import ReactDOM from 'react-dom'
...
+
+  const updateStyle = React.useCallback(() => {
+    const targetNode = props.target ? props.target.current : ref.current
+    if (props.when && targetNode && tetheredRef.current) {
+      const nextStyle = props.position(targetNode).styleFor(tetheredRef.current)
+      setStyle(nextStyle)
+    }
+  }, [props])
+
+  React.useEffect(() => {
```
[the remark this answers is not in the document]
danethurber: the function is meant to be called on mount and when certain props change. it's in an effect so i can handle those cases

packages/typeahead/src/react/__specs__/index.spec.js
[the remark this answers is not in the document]
danethurber: i do think the current approach is the right one because we're using this component like a Select...

packages/position/src/react/index.js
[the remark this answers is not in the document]
jaketrent: Yeah, the similarity to other libs makes sense.  Now there's just 2 names for the same thing in this lib.  `show` is not a boolean in this lib.  Probably not a big deal.  I love to use the same name for the same/related concept everywhere when possible.  Nbd at this point.

To clarify the api design, it was meant to read something like "Position below left of this thing (child), and *show* this other thing *when* something toggles true." fwiw

packages/actionmenu/src/css/index.js
```diff
@@ -32,6 +32,8 @@ export default {
     padding: `${vars.style.menuPaddingVert} 0`,
     minWidth: '160px',
     maxWidth: '320px',
+    maxHeight: '277px',
+    overflowY: 'auto',
```
jaketrent: For some reason, this or something else in this branch is affecting more than just Y overflow.  The X scrollbar is now always there, whether needed or not.  

For example:
Screen Shot 2019-10-01 at 11 20 01 AM

packages/position/src/react/__stories__/index.story.js
```diff
@@ -1,139 +1,220 @@
...
-    this.state = { style: {} }
-    this.tooltipRef = React.createRef()
+  basicStories.add(displayName, () => (
+    <Comp show={<MockToolip />}>
+      <Box>{name}</Box>
+    </Comp>
+  ))
+})
+
+const portalStories = storiesOf('Components | Position / portals', module)
```
jaketrent: The portal story is still looking messed up to me (not centered where it should be).  Does it look this way for you?

Screen Shot 2019-10-01 at 11 21 21 AM

jaketrent: The site portal example is also messed up:

Screen Shot 2019-10-01 at 11 20 30 AM

packages/position/src/react/index.js
```diff
@@ -4,84 +4,152 @@ import ReactDOM from 'react-dom'
...
+  React.useImperativeHandle(forwardedRef, () => ref.current)
+
+  const child = React.Children.only(props.children)
+  const tetheredRef = React.useRef()
+  const tetheredEl = React.cloneElement(props.show, {
+    ref: tetheredRef,
+    style: { ...child.props.style, ...style }
+  })
+
+  const updateStyle = React.useCallback(() => {
```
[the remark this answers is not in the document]
jaketrent: Looks like these got missed.  Will you please name the Position component effect callbacks?

packages/position/src/react/index.js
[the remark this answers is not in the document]
jaketrent: Thanks for the explanatory notes.  

So, when the updateStyle function changes, call it.  And I guess that changes when useCallback changes above, based on the inputs there.  Cool, thanks.

packages/position/src/react/use-debounce-callback.js
```diff
@@ -0,0 +1,32 @@
+import React from 'react'
+
+export default function useDebounceCallback(fn, delay = 100) {
+  const timeout = React.useRef()
+
+  const debounced = React.useCallback(
+    function() {
```
jaketrent: please name callback.

packages/position/src/react/use-debounce-callback.js
```diff
@@ -0,0 +1,32 @@
...
+      }
+
+      clearTimeout(timeout.current)
+      timeout.current = setTimeout(later, delay)
+    },
+    [fn, delay]
+  )
+
+  React.useEffect(
+    () => () => {
```
jaketrent: please name callback.

packages/typeahead/src/react/index.js
```diff
@@ -0,0 +1,250 @@
...
+  return (
+    <div
+      {...filterReactProps(omit(props, TEXT_INPUT_PROPS))}
+      ref={containerRef}
+    >
+      <BelowLeft
+        inNode={portal.current}
+        when={open}
+        show={
+          <SuggestionsMenu
```
jaketrent: No idea what might be causing this, but seeing weird scroll issues.  Happens in Chrome and FF.  Happens with scroll wheel on mouse but not on the Mac trackpad.

Would you mind seeing if you or another person have the same mouse wheel problem?  If it's just me, maybe something unrelated to the code.

FF:
ff

Chrome:
chrome

packages/position/src/react/__stories__/index.story.js
danethurber: weird. i'm not seeing the issue. i guess can you bootstrap and double check. i've noticed on the docs site it can take a sec to get the correct position - i assumed that was a nextjs in dev mode issue 🤷‍♂

packages/position/src/react/index.js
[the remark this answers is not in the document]
danethurber: ok i updated hooks but this one. because this is memoizing a function named `updateStyle` it seems redundant to also name the function. the name would probably just be `updateStyle` 🤷‍♂

packages/position/src/react/__stories__/index.story.js
jaketrent: POSITION ISSUE IN STORYBOOK:
Just updated the branch, clean, bootstrap, checked storybook for Position in ff, chrome, safari, brave.  They'll still look like the storybook screenshot above.  Does yours? 

(To clarify, it sounds like you were saying the *site* didn't look like the screenshot I posted above.  How about your storybook?  Does it match the problem screenshot?)

POSITION ISSUE IN SITE:
I also double-checked the site portal render in these browsers.  They all show the same as the problem screenshot above, except for Safari, which is a bit diff:

Screen Shot 2019-10-04 at 1 55 44 PM

And then I noticed the the "js example" below it is also messed up (supposed to be `rightOf`):

Screen Shot 2019-10-04 at 1 56 41 PM

ISSUE WITH SSR TYPEAHEAD:
Finally, I thought I'd build the site so we could test the hypothesis about nextjs dev server being part of the problem.  When I built, I got:

```
Error occurred prerendering /components/typeahead: ReferenceError: document is not defined
    at usePortal (/Users/jaketrent/dev/design-system/packages/typeahead/dist/react/index.js:277:35)
```

So I think there probably needs to be some conditional around the use of `document` in Typehaead for SSR support.

packages/actionmenu/src/css/index.js
danethurber: constraining the height on the ActionMenu is causing lots of issues because of the nested menus. i'll remove that from the ActionMenu

jest/setup/react-dom.js
```diff
@@ -0,0 +1,9 @@
+import ReactDOM from 'react-dom'
+
+beforeAll(() => {
+  ReactDOM.createPortal = jest.fn(el => el)
```
jaketrent: why do this?

packages/position/src/react/index.js
```diff
@@ -49,48 +49,52 @@ LeftOf.displayName = 'LeftOf'
...
 
   return (
     <React.Fragment>
       {target ? child : React.cloneElement(child, { ref })}
 
-      {when &&
-        (inPortal
-          ? ReactDOM.createPortal(tetheredEl, props.inNode)
-          : tetheredEl)}
+      {createUniversalPortal(
```
jaketrent: Why always render in portal?

What tradeoffs does this represent?  For instance, is Position any less useful generally?

packages/typeahead/src/react/menu.js
```diff
@@ -0,0 +1,76 @@
...
+  menu: () => css(stylesheet['.psds-typeahead__menu']),
+  menuItem: () => css(stylesheet['.psds-typeahead__menu__item']),
+  menuItemIcon: ({ selected }) =>
+    compose(
+      css(stylesheet['.psds-typeahead__menu__item__icon']),
+      selected && css(stylesheet['.psds-typeahead__menu__item__icon--selected'])
+    )
+}
+
+const SuggestionsMenu = React.forwardRef((props, forwardedRef) => {
```
jaketrent: Very nice and concise!  

Reuse is overrated.

packages/position/src/react/index.js
danethurber: using a portal helped simplify and fix some issues i was having. also i think a portal is the write call and we know that other's have been using it this way. i don't believe it makes it less useful but let me know if you have some scenarios i didn't think of

jest/setup/react-dom.js
danethurber: so i didn't have to skip the storyshots. jsdom(or whatever jest is using) doesn't work with portals but this way i found we can just render the portal's content back and still get the snapshot

danethurber:
uh yes....i thought i fixed the docs but looks like it didn't work....hmmm....so i can ssr it because i can't calculate the position. what would be the be way to say don't try to render until you're in the browser?

jaketrent:
As far as React APIs go, there's nothing to say "don't render this until browser land".  But perhaps a document onload listener that sets state and triggers (re) render of children?

PR #655, opened 2019-09-05
Refactor/test library

danethurber:
- remove enzyme in favor of testing-library
- remove some tests with little value
- update test deps
- update broken snapshots from master after merge


packages/actionmenu/src/react/__specs__/index.spec.js
```diff
@@ -57,5 +57,3 @@ describe('ActionMenu', () => {
     fireEvent.keyDown(opt, { key: 'Enter', code: 13 })
   })
 })
-
-afterEach(cleanup)
```
jaketrent: Why is this no longer required?

packages/breadcrumb/src/react/__specs__/index.spec.js
```diff
@@ -1,39 +1,41 @@
...
-})
+    expect(ref.current).not.toBeNull()
+  })
+
+  describe('when disabled', () => {
+    const handleClick = jest.fn()
+    let container, button
+
+    beforeEach(() => {
+      ;({ container } = render(
```
jaketrent: super weird syntax.  I'd recommend not destructuring here in order to avoid it.

packages/button/src/react/__specs__/index.spec.js
```diff
@@ -1,48 +1,52 @@
...
-  const ref = React.createRef()
-  mount(<Button ref={ref} />)
-  expect(ref.current).not.toBeNull()
-})
+  describe('when disabled', () => {
+    const handleClick = jest.fn()
+    let container
+
+    beforeEach(() => {
+      ;({ container } = render(
```
jaketrent: ditto

packages/card/src/react/__specs__/index.spec.js
```diff
@@ -1,14 +0,0 @@
-import { shallow } from 'enzyme'
-import React from 'react'
-
-import Card from '../index.js'
-
-test('click on Card.Title triggered once', () => {
```
jaketrent: Why remove this w/o replacement?  Still seems like a valuable test to me.

packages/datawell/src/react/__specs__/index.spec.js
```diff
@@ -1,11 +0,0 @@
-import React from 'react'
-import { shallow } from 'enzyme'
-
-import DataWell from '../index.js'
-
-describe('DataWell', () => {
```
jaketrent: Why the removal?  "Renders" tests have always been a little hokey, but they are at least some kind of smoke test.

packages/icon/src/react/__specs__/index.spec.js
```diff
@@ -18,14 +18,20 @@ describe('Icon', () => {
     it('should be exposed', () => expect(Icon.ids).toEqual(vars.ids))
   })
 
+  it.todo('forwards refs')
```
jaketrent: Did you want to address this in this commit?

packages/starrating/src/react/__specs__/index.spec.js
```diff
@@ -1,80 +1,57 @@
+import { render } from '@testing-library/react'
 import React from 'react'
-import { mount } from 'enzyme'
 
-import StarRating from '..'
+import StarRating from '../index.js'
```
jaketrent: Apparently we've all been missing style updates in quite a few spec files...

packages/starrating/src/react/__specs__/index.spec.js
```diff
@@ -1,80 +1,57 @@
...
     beforeEach(() => {
-      wrapper = mount(<StarRating />)
-      stars = wrapper.find('Star')
-
-      expect(stars).toHaveLength(5)
-    })
-
-    it('renders stars as NOT interactive', () => {
-      stars.forEach(star => expect(star.prop('interactive')).toBe(false))
+      ;({ container } = render(<StarRating />))
```
jaketrent: ditto

packages/starrating/src/react/__specs__/index.spec.js
```diff
@@ -1,80 +1,57 @@
...
-      wrapper = mount(<StarRating onChange={jest.fn()} />)
-      stars = wrapper.find('Star')
-
-      expect(stars).toHaveLength(5)
-    })
-
-    it('renders stars with empty appearance', () => {
-      stars.forEach(star => expect(star.prop('appearance')).toBe('empty'))
+      const handleChange = jest.fn()
+      ;({ container } = render(<StarRating onChange={handleChange} />))
```
jaketrent: Ditto.  And now that I'm looking at this, this could all be helped by just making a single `it` block instead of needing to share closure scope vars with a `beforeEach` block.

packages/switch/src/react/__specs__/index.spec.js
```diff
@@ -1,28 +1,73 @@
...
+    )
+
+    const label = container.querySelector('label')
+    fireEvent.click(label)
+
+    expect(handleClick).toHaveBeenCalled()
+  })
 })
+
+// test('onClick calls handler', () => {
```
jaketrent: I believe these commented tests can be removed.

packages/table/src/react/__specs__/__snapshots__/storyshots.spec.js.snap
```diff
@@ -1627,6 +1631,7 @@ exports[`Storyshots drawer single 1`] = `
       data-css-1i2htk9=""
     >
       <div
+        aria-hidden={true}
```
jaketrent: These are the snapshot updates (broken in master) that I've been asking Edward about because of his Drawer work.  I'm not 100% sure that the addition of `aria-hidden` here is correct.  Are you?

packages/theme/src/react/__specs__/index.spec.js
```diff
@@ -1,5 +1,82 @@
...
+    expect(child).not.toBeNull()
+  })
+
+  it('adds theme to context', () => {
+    const { container } = render(
+      <Theme name={Theme.names.light}>
+        <MockComponent />
+      </Theme>
+    )
+    expect(container).toHaveTextContent(Theme.names.light)
```
jaketrent: Clever!

jaketrent: I think it's funny how testing-library's "test what's in the DOM because that's what your user sees" is bent here.  It's in the DOM, but testing React.Context. :P

packages/theme/src/react/__specs__/with-theme.spec.js
```diff
@@ -1,54 +0,0 @@
-import { mount } from 'enzyme'
```
jaketrent: Is this gone because we don't plan to support `withTheme`?  If so, shouldn't we keep it around until that moment we actually remove the `withTheme` src?

packages/actionmenu/src/react/__specs__/index.spec.js
danethurber: breaking change in testing-library. it's no longer needed or available

packages/datawell/src/react/__specs__/index.spec.js
danethurber: idk...seemed pretty basic and there are storyshots that do a better job. I do usually write a test like this(it's in the generator) but it's not consistent for all components. we could decided not to do a "render" test or we could make sure all component have a render test. thoughts?

packages/starrating/src/react/__specs__/index.spec.js
danethurber: yeah. i think we're pretty close to the fated "update all the lint errors" commit. no rush though

packages/theme/src/react/__specs__/with-theme.spec.js
danethurber: yeah i can refactor to use testing-library for now and then we'll remove it when we remove `withTheme`

PR #656, opened 2019-09-05
build: use plopjs to replace home grown generator

danethurber:
### What You're Solving

- simplify generator
- more advanced generator templates
- less dependencies

### How to Verify

- `npm run generate`


plop-templates/component/src/react/__specs__/index.spec.js.hbs
```diff
@@ -0,0 +1,24 @@
...
+describe('{{pascalCase name}}', () => {
+  it('renders', () => {
+    const { getByTestId } = render(
+      <{{pascalCase name}} data-testid="undertest" />
+    )
+
+    expect(getByTestId('undertest')).toBeInTheDocument()
+  })
+
+  it('forwards refs', () => {
```
jaketrent: yissssss, free tests.

package.json
```diff
@@ -18,7 +18,8 @@
   "scripts": {
     "bootstrap": "lerna bootstrap --hoist && link-parent-bin",
     "clean": "lerna clean",
-    "component:create": "node ./scripts/create-component/index.js",
+    "generate": "plop",
```
jaketrent: are there any docs that need updated because the script name changed?

PR #659, opened 2019-09-06
fix(radio): stops onSelect firing twice every time we click a radio

### LINKED ISSUE #657: radio: onSelect firing twice every time we click a radio button
image (4)
From Travis Bumgarner
onSelect is called both when an item is deselected and when a new one is selected.

EdwardIrby:
fix #657

### DIFF
--- a/packages/radio/src/react/__specs__/index.spec.js
+++ b/packages/radio/src/react/__specs__/index.spec.js
@@ -1,5 +1,5 @@
 import React from 'react'
-import { render } from '@testing-library/react'
+import { render, fireEvent } from '@testing-library/react'
 
 import Radio from '../index.js'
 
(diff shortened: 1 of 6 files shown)


EdwardIrby:
> Were you able to repro this? It'd be nice to have a failing test that this PR would then address. What do you think?

I was able to reproduce this. I'll write a test to show it's not happening.

packages/radio/src/react/__specs__/index.spec.js
```diff
@@ -19,4 +19,19 @@ describe('Radio', () => {
     expect(group.current).not.toBeNull()
     expect(button.current).not.toBeNull()
   })
+  it('fires event once', () => {
+    const group = React.createRef()
```
danethurber: these refs aren't needed for this test

packages/radio/src/react/__specs__/index.spec.js
```diff
@@ -19,4 +19,19 @@ describe('Radio', () => {
...
+    const button = React.createRef()
+    const spy = jest.fn()
+    const { container } = render(
+      <Radio.Group ref={group} onSelect={spy}>
+        <Radio.Button value="red" label="Red" />
+        <Radio.Button ref={button} value="green" label="Green" />
+        <Radio.Button value="blue" label="Blue" />
+      </Radio.Group>
+    )
+    const rdio = container.querySelector('label')
```
danethurber: typo. should be `radio`

packages/radio/src/react/__specs__/index.spec.js
EdwardIrby: fixed

packages/radio/src/react/__specs__/index.spec.js
EdwardIrby: fixed

PR #663, opened 2019-09-16
fix(tab): focus borderline jump

EdwardIrby:


packages/tab/src/css/index.js
```diff
@@ -44,31 +44,21 @@ export default {
...
+      boxShadow: '0 0 0 2px rgba(0, 123, 255, .5)'
     }
   },
-  [`.psds-tab__list.psds-theme--${themeNames.light}`]: {
-    borderBottom: `1px solid ${core.colors.gray02}`
-  },
   [`.psds-tab__list.psds-theme--${themeDefaultName}`]: {
-    borderBottom: `1px solid ${core.colors.gray04}`
+    borderBottom: `1px solid ${core.colors.gray04}`,
+    '&:focus': {
```
danethurber: now that the focus state is the same for all themes you could define these styles once on the `.psds-tab__list` selector

PR #664, opened 2019-09-20
Feature/esm build button

EdwardIrby:
This PR is proof of concept of modifying a DS CJS package, Button,  into a ESM ready package with backwards compatibility. The purpose of which is to enable tree-shaking so that apps making use of our packages generate a smaller bundle improving performance.

It includes a build script addition to the package.json of Button in addition to a a new `process..env.ESM` variable, type boolean, for modifying our default babel configuration. Also in order to preserve SSR functionality for those using Next.js it necessitated modifying the current CJS module require usage patterns.

**CJS Usage**
```js
const Button = require('@pluralsight/ps-design-system-button')
const css = require('@pluralsight/ps-design-system-button/css.js')
const vars = require('@pluralsight/ps-design-system-button/vars.js')
```

**ESM Usage**
```js
import Button, {vars, css } from '@pluralsight/ps-design-system-button'
```


babel.config.js
```diff
@@ -12,17 +12,17 @@ module.exports = api => {
 
     ignore,
 
-    presets: ['@babel/preset-react', '@babel/preset-env'],
+    presets: ['@babel/preset-react', !esm && '@babel/preset-env'].filter(Boolean),
```
danethurber: i'm looking at this line and it looks like it's always including preset-env....am i reading it wrong?

packages/button/index.js
```diff
@@ -1,5 +1,7 @@
-const css = require('./css')
-const React = require('./react')
-const vars = require('./vars')
+import Button from './esm/react/index.js'
+import * as vars from './esm/vars/index.js'
+import css from './esm/css/index.js'
 
-module.exports = { css: css, react: React, vars: vars }
+export default Button
+export { vars }
```
danethurber: can the vars and css exports be on the same line?

EdwardIrby: Probably I've never tried that  I'll give it a shot

babel.config.js
EdwardIrby: preset-env is only included when esm is not true. When the esm build runs esm is true though I should probably make sure it's false when the CJS build runs.

danethurber: nevermind - i can't read

EdwardIrby:
> have you looked into publishing beta packages yet

Nope that's next up pretty sure it's in the Lerna docs

EdwardIrby:
> > have you looked into publishing beta packages yet
> 
> Nope that's next up pretty sure it's in the Lerna docs
https://github.com/lerna/lerna/tree/master/commands/publish#--preid
`lerna publish --canary --preid next`

EdwardIrby:
@jaketrent maded the mods we taslked about but decided to let **.css** & **.scss** files build to `/dist`. Otheriwse I would have had to add a windows fix for fs.mkdir on `dist/css` or update the required node from `"node": ">= 7.9",` to `"node": ">= 10.16",`

babel.config.js
```diff
@@ -12,7 +12,9 @@ module.exports = api => {
 
     ignore,
 
-    presets: ['@babel/preset-react', '@babel/preset-env'],
+    presets: ['@babel/preset-react', !esm && '@babel/preset-env'].filter(
```
jaketrent: If @babel/preset-env, I'm concerned that we cut it out when doing esm builds.  The problem is the same with cutting out the object rest spread plugin: we make it so consumers must transpile with babel to get the code browser ready.  We instead want it all browser-ready, minus the esm format, and webpack will take care of removing the module format and making it browser ready.

I think this will mean finding out what plugins @babel/preset-env is pulling in for our current env (probably driven by our browserlist?) and then adding those babel plugins specifically here.  

What do you think, Edward?

packages/build/bin/build-css.js
```diff
@@ -1,18 +1,17 @@
...
-const defaultOutputDirectoryPath = path.join(process.cwd(), 'dist', 'css')
+const defaultInputFilePath = 'dist/css/index.js'
+const defaultOutputDirectoryPath = 'dist/'
 
 const prefix = '[build-css] '
 const log = str => console.log(chalk.dim(prefix + str))
 const logError = str => console.log(chalk.red(prefix + str))
 const logSuccess = str => console.log(chalk.green(prefix + str))
-
+console.log(defaultInputFilePath)
```
jaketrent: rm

packages/build/bin/build-css.js
```diff
@@ -1,18 +1,17 @@
...
-
 const chalk = require('chalk')
 const program = require('commander')
 const path = require('path')
 
 const { css } = require('../index')
-const defaultInputFilePath = path.join(process.cwd(), 'dist', 'css', 'index.js')
-const defaultOutputDirectoryPath = path.join(process.cwd(), 'dist', 'css')
+const defaultInputFilePath = 'dist/css/index.js'
+const defaultOutputDirectoryPath = 'dist/'
```
jaketrent: Why rm the smarter path.join?  It worked, and it was guaranteed to be cross-platform happy.

packages/build/bin/build-css.js
```diff
@@ -28,8 +27,8 @@ program
   .option('-g, --useGlamor', 'Compile css using glamor', false)
   .parse(process.argv)
 
-const inputFilePath = program.input
-const outputDirectoryPath = program.output
+const inputFilePath = path.join(process.cwd(), program.input)
```
jaketrent: Was this required to get the input cli flag to work?

packages/site/pages/components/button.js
```diff
@@ -1,5 +1,5 @@
 import core from '@pluralsight/ps-design-system-core'
-import Button from '@pluralsight/ps-design-system-button/react.js'
+import Button from '@pluralsight/ps-design-system-button'
```
jaketrent: This is a good usage to replace.  Seems like maybe we want to replace (all) others?  ...Make this the default usage.  Where else might that be?  Storybooks, probably...

packages/button/package.json
```diff
@@ -5,11 +5,13 @@
   "license": "Apache-2.0",
   "repository": "pluralsight/design-system",
   "author": "jaketrent",
-  "main": "index.js",
+  "main": "react.js",
```
jaketrent: This definitely makes this a breaking change, but I don't think your commit specifies that yet.  Please follow this spec for that:

https://www.conventionalcommits.org/en/v1.0.0-beta.2/#summary

packages/build/bin/build-css.js
EdwardIrby: Yep

packages/build/bin/build-css.js
EdwardIrby: turns out the --input flag did not work unless I did this

packages/site/pages/components/button.js
EdwardIrby: Cool I'll add that. Wasn't sure how a big a change we wanted to do on what's a prototype export pattern right now.

babel.config.js
EdwardIrby: Fair I'll add a conditional on this 
```
{
  "targets": {
    "esmodules": true
  }
}
```

jaketrent:
Ran a little project experiment:
[treeShakeTest.zip](https://github.com/pluralsight/design-system/files/3673197/treeShakeTest.zip)

Numbers were diff on output, suggesting treeshaking working:

```
esm, dev
3.7M bundle.js

esm, prod
257K bundle.js

cjs, dev
3.7M bundle.js

cjs, prod
260K bundle.js
```

But, as you can see, the numbers are not thrilling.  And it's hard for me to really tell what's diff.  Perhaps I should look at the src files and anticipate what I'd *think* should be shaken out.

The incrementally better measurement suggests that there's probably more to do to make these packages reactive to treeshaking for greater perf gains.

packages/build/bin/build-css.js
```diff
@@ -1,12 +1,11 @@
 #!/usr/bin/env node
-
 const chalk = require('chalk')
 const program = require('commander')
 const path = require('path')
 
 const { css } = require('../index')
-const defaultInputFilePath = path.join(process.cwd(), 'dist', 'css', 'index.js')
-const defaultOutputDirectoryPath = path.join(process.cwd(), 'dist', 'css')
+const defaultInputFilePath = path.join('dist', 'css', 'index.js')
```
jaketrent: I put the cross-plat separators back in, kept cwd out, tested and it worked.

EdwardIrby: Cool

PR #667, opened 2019-09-23
feat(banner): add button component

### LINKED ISSUE #665: a Button in a Banner should inherit the text color
the button's text and border color should match the text color

danethurber:
### What You're Solving

- resolves #665

### Design Decisions

`Banner` now includes `Banner.Button`

I tried out a couple different approach but was hitting issues with css specificity in regards to the button styles. Our glamor setup was making it really difficult to fix the css order and specificity. I opted to create a new subcomponent(`Banner.Button`) which I feel is the best way to ensure the correct styles are applied. I also prefer this approach to encapsulate and ensure the right type of button is used.

### DIFF
--- a/packages/banner/package.json
+++ b/packages/banner/package.json
@@ -32,6 +32,7 @@
   },
   "devDependencies": {
     "@pluralsight/ps-design-system-build": "^1.10.1",
+    "@pluralsight/ps-design-system-button": "*",
     "@pluralsight/ps-design-system-storybook-addon-theme": "^4.0.1"
   }
 }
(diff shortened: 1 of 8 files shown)


packages/banner/package.json
```diff
@@ -32,6 +32,7 @@
   },
   "devDependencies": {
     "@pluralsight/ps-design-system-build": "^1.10.1",
+    "@pluralsight/ps-design-system-button": "*",
```
jaketrent: sad, it would have been great to get this in css.  I read your comment and feel the pain on why.  This is another reason esm needs to get here.

packages/banner/src/react/index.js
```diff
@@ -37,8 +44,21 @@ Banner.propTypes = {
 Banner.defaultProps = {
   color: vars.colors.blue
 }
-Banner.colors = vars.colors
 
+const Button = React.forwardRef((props, forwardRef) => (
```
jaketrent: Kinda stinky that this forwardRef API from React requires that you explicitly write this at every custom level, instead of just showing up with props or something.

Issue #661, opened 2019-09-13
build: css file output is invalid


danethurber:
CSS dist files contain invalid css selectors

ref: https://pluralsight.slack.com/archives/C70HPSJPP/p1568393333038300

PR #668, opened 2019-09-23
Patch/normalize build

danethurber:
fixes: #661 

- fix nested selectors in normalized package output
- use glamor build for fix output in button component
- upgrade postcss packages in normalize package

 _NOTE_: css vars are now included in normalize package output. this is because of the upgrades and how the postcss-preset-env package works. they have fallbacks for older browsers. i don't think it's a big deal but it is of note.


packages/normalize/postcss.config.js
```diff
@@ -0,0 +1,12 @@
+module.exports = {
+  plugins: [
+    require('postcss-import'),
+    require('postcss-preset-env')({
+      browsers: 'last 2 versions, IE <= 10',
```
jaketrent: does it matter that the browerlists in this pr are diff?

danethurber: yes....i need to remove those. good eye

PR #674, opened 2019-10-02
Feature/esm button beta

EdwardIrby:
Had to redo this PR to do publish


babel.config.js
```diff
@@ -12,7 +13,10 @@ module.exports = api => {
 
     ignore,
 
-    presets: ['@babel/preset-react', '@babel/preset-env'],
+    presets: [
+      '@babel/preset-react',
+      esm ? ['@babel/preset-env', { modules: false }] : '@babel/preset-env'
+    ],
```
danethurber: i would have do it this way

```
presets: [
      '@babel/preset-react',
      ['@babel/preset-env', { modules: !esm }]
    ],
```

your call

packages/button/package.json
```diff
@@ -1,22 +1,24 @@
...
+  "main": "react.js",
+  "module": "index.js",
   "scripts": {
-    "build": "run-s build:js build:css",
-    "build:css": "build-css --useGlamor",
-    "build:js": "babel --root-mode upward src --out-dir dist",
+    "build": "run-s build:js build:cjs build:css",
+    "build:cjs": "babel --root-mode upward src --out-dir dist/cjs",
+    "build:css": "build-css --useGlamor -i dist/cjs/css/index.js",
+    "build:js": "cross-env ESM=true babel --root-mode upward src --out-dir dist/esm",
```
danethurber: is the name `build:esm` better? maybe? idk

EdwardIrby: I think the idea is to move every component to esm so I decided to change the name of the old build to reflect it's the fallback `build:cjs`

babel.config.js
EdwardIrby: modules argument cannot be a boolean value of `true` I'll switch this to a ternary in the object value though.

PR #678, opened 2019-10-04
feat(icon): alternate caret icons

danethurber:
this was a request from gitprime. we already use this icon in the Dropdown but there isn't a version that is exposed in the Icons package. Gitprime would like these icons as they built out some custom components


danethurber:
👍

yes this are the names from the figma components

PR #682, opened 2019-10-09
Checkbox: support indeterminate

danethurber:
Screen Shot 2019-10-09 at 1 58 49 PM

### DIFF
--- a/packages/checkbox/src/css/index.js
+++ b/packages/checkbox/src/css/index.js
@@ -80,37 +80,37 @@ export default {
   '.psds-checkbox__square--error:after': {
     background: core.colors.red
   },
-  '.psds-checkbox__square--checked': {
+  '.psds-checkbox__square--active': {
     background: core.colors.blue,
     border: 'none'
   },
-  '.psds-checkbox__square--checked:focus:before, .psds-checkbox__square--checked--error:before': {
+  '.psds-checkbox__square--active:focus:before, .psds-checkbox__square--active--error:before': {
     top: '-1px',
     left: '-1px',
     right: '-1px',
     bottom: '-1px'
   },
-  [`.psds-checkbox__square--checked.psds-theme--${themeNames.light}:focus:before, .psds-checkbox__square--checked--error.psds-theme--${themeNames.light}:before`]: {
+  [`.psds-checkbox__square--active.psds-theme--${themeNames.light}:focus:before, .psds-checkbox__square--active--error.psds-theme--${themeNames.light}:before`]: {
     top: '-2px',
     left: '-2px',
     right: '-2px',
     bottom: '-2px'
   },
-  '.psds-checkbox__square--checked:focus:after, .psds-checkbox__square--checked--error:after': {
+  '.psds-checkbox__square--active:focus:after, .psds-checkbox__square--active--error:after': {
     top: '-4px',
     left: '-4px',
     right: '-4px',
     bottom: '-4px'
   },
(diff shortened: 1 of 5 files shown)


packages/checkbox/src/react/index.js
```diff
@@ -29,10 +30,22 @@ const styles = {
...
+
+  const ref = React.useRef()
+  React.useImperativeHandle(forwardedRef, () => ref.current)
+
   const square = React.createRef()
 
+  React.useEffect(
+    function updateIndetermiateAttr() {
+      if (!ref.current) return
+      ref.current.indeterminate = props.indeterminate
```
danethurber: the `indeterminate` attribute has to use the node setter - it isn't an html attribute

packages/site/pages/components/checkbox.js
```diff
@@ -185,13 +185,6 @@ export default _ => (
...
             'error state flag'
           ]),
-          PropTypes.row([
-            'innerRef',
-            'DOM element => ()',
-            null,
-            null,
-            'input field react ref callback'
-          ]),
           PropTypes.row(['name', 'string', null, null, 'form data identifier']),
```
danethurber: i forgot to update the docs. i'll do that now. one sec

packages/checkbox/src/react/index.js
```diff
@@ -68,7 +81,8 @@ const Checkbox = React.forwardRef((props, ref) => {
           tabIndex={props.disabled ? '-1' : '0'}
           {...styles.square(themeName, props)}
         >
-          {props.checked && <Checkmark />}
+          {props.indeterminate && <Indeterminate />}
+          {props.checked && !props.indeterminate && <Checkmark />}
```
jaketrent: Kind of an odd api, but apparently it matches the native checkbox api. Who knew!  https://css-tricks.com/indeterminate-checkboxes/

packages/site/pages/components/checkbox.js
```diff
@@ -232,6 +232,20 @@ export default _ => (
         ]}
       />
 
+      <SectionHeading>Indeterminate</SectionHeading>
+      <P>
+        Passing the <Text.Code>indeterminate</Text.Code> indicates that the
```
jaketrent: rm "the"

packages/site/pages/components/checkbox.js
```diff
@@ -232,6 +232,20 @@ export default _ => (
...
+        checkbox value is indeterminate.
```
jaketrent: Maybe try to avoid using the word in the definition.  Maybe something like:

"Passing the indeterminate flag indicates that the checkbox selection should be obscured as if the control was in a third, indeterminate, state."

packages/checkbox/src/css/index.js
```diff
@@ -80,37 +80,37 @@ export default {
   '.psds-checkbox__square--error:after': {
     background: core.colors.red
   },
-  '.psds-checkbox__square--checked': {
+  '.psds-checkbox__square--active': {
```
jaketrent: As I've looked at this, first I wanted to keep this as `checked`.  I see that it backs checked and indeterminate states now.  `active` is a somewhat odd name for this, but I don't have a better name.  🤷‍♂  Dunno if you had the same struggle.  "Active" really probably is the generic best.  It matches the lots of other DS components too.

danethurber:
@jaketrent my assumption is that a user doesn't trigger `indeterminate` but it's from some other app state. i'm looking at some other implementations and I think this is pretty close to the pseudo-standard but i'm posting a question in slack and seeing if I get any bites

PR #684, opened 2019-10-11
testing github actions

danethurber:
playing around. travis and me aren't getting along right now - so i'm exploring my options


danethurber:
this works and is pretty simple. not as fast as travis but there is most likely a way to share build artifacts across the different step. a commitlint integration will need to be found - i tried a few but had issues getting them to work. closing for now but can revisit in the future

Issue #687, opened 2019-10-15
DatePicker: Incorrect formatting of nextDate is preventing onBlur from being called on SubFields


cpaisan:
## Bug
  The DatePicker `onSelect` prop is not firing on `onBlur` events when a user blurs from a `SubField`. The `nextDate` object inside of the `handleSubFieldBlur` function is incorrectly formatted and omits a `mm` or `yyyy` value when either sub field is updated and is preventing the `onBlur` from firing. Please see: https://github.com/pluralsight/design-system/blob/master/packages/datepicker/src/react/index.js#L156
### Expected behavior
  The `onSelect` should be called for valid dates when a user triggers `onBlur` on any of the `SubFields` on the DatePicker
### Actual behavior
  `onSelect` is not being called when `onBlur` is triggered on the DatePicker's `SubField`s
### Steps to repro

1. Change either the `month` or `year` field manually on the DatePicker, not via the calendar component.
2. Tab or click outside of the selected input.
3. `onSelect` should be called for valid dates.

### Related packages

- datepicker@4.0.5

### Environment

- OS - MacOS 10.12
- Browser version - Chrome v60

PR #688, opened 2019-10-15
fix(datepicker): fix onblur behavior for subfields in datepicker

cpaisan:
- Fixes `nextDate` object key-values to include month and year values in the `handleSubFieldBlur` function
- Add `onSelect` unit tests for the DatePicker

### What You're Solving

Resolves issue #687 - Fixes `onBlur` not calling `onSelect` for valid dates

### How to Verify

1. Pass the DatePicker an `onSelect` prop
2. Trigger a blur event with a valid date by focusing and unfocusing out of a SubField in the DatePicker
3. Ensure that the function passed to `onSelect` was called

### DIFF
--- a/packages/datepicker/src/react/__specs__/index.spec.js
+++ b/packages/datepicker/src/react/__specs__/index.spec.js
@@ -1,5 +1,5 @@
 import React from 'react'
-import { render } from '@testing-library/react'
+import { render, fireEvent } from '@testing-library/react'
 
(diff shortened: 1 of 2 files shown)


cpaisan:
I'm not sure why the snapshot tests are failing. Any help would be greatly appreciated.

EdwardIrby:
Going to keep this open and roll back to this at the end of the week based on our meeting

PR #689, opened 2019-10-21
Feat/small inputs

### LINKED ISSUE #672: Small inputs/selects
Add a small size for inputs and selects. Need a prop for inputs and selects for size. Values will be medium and small (defaults to medium).

EdwardIrby:
resolves  #672


EdwardIrby:
> everything looks good to me. if you felt like adding some additional examples to the documentation that would be a +1

Sure, sounds like a good idea

PR #692, opened 2019-10-24
First ESM packages set

### LINKED ISSUE #690: OKR: ESM Packages Batch 2
- regex fix on glamor-to-css.js line 18
- update builds for the following:
  - build
  - core
  - filter-react-props
  - packages-api
  - prop-types
  - storybook-addon-center
  - storybook-addon-theme
  - util

EdwardIrby:
resolves #690
refactors exports for
- build
- core
- filter-react-props
- util


packages/util/src/index.js
```diff
@@ -1,9 +1,4 @@
-import * as string from './string'
-import * as propDefs from './prop-defs'
-import * as props from './props'
-
-export default {
-  string,
-  propDefs,
-  props
-}
+export { transparentize } from './color.js'
```
jaketrent: I really liked the old organization better.

`util` is very broad.  Splitting the api into functional groups is useful to for understanding.  It also gives better context for function names and makes simpler names like `transparentize` possible instead of `transparentizeColor`.

What is the reason for bringing all the functions into the top level namespace?

packages/core/js/index.js
```diff
@@ -1,7 +1,5 @@
-module.exports = {
-  accessibility: require('./accessibility'),
-  colors: require('./colors'),
-  layout: require('./layout'),
-  type: require('./type'),
-  motion: require('./motion')
-}
+export { default as accessibility } from './accessibility'
```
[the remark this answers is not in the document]
EdwardIrby: It's the ESM syntax required for treeshaking.

packages/util/src/index.js
EdwardIrby: Ergonomics more than anything. When using ESM syntax you have to do more renaming and destructing when group like that. If it was one giant repo I'd agree but each of these packages is relatively small even the `util`.  So for example: 
```js
// Works
const { transparentize } = require(`@ps-util`).color

// Nope
import { transparentize }  from `@pluralsight/ps-design-system-util`.color
```
It's actually one of the advantages of CJS is the ability to 1 line things like this. 

I'm not sure grouping really matters but names of the functions I want to quickly reference and use do. We don't document any of these functions yet. They are used internally in our work on the DS. We did talk about documenting them during retro Friday. 

Personally when I use a util library whether self made or external I've not seen grouping unless we're referring to somethings as extensive as the the node js api, but libraries like lodash do not. Thoughts?

jaketrent: Would this work with esm?

```
import { transparentize } from `@ps-util/color.js`
```

PR #694, opened 2019-10-28
Second esm packages set

EdwardIrby:
set 2: dep ds-components
- button
- avatar
- collapsible
- focusmanager
- link
- position
- theme
- tooltip
- icon


packages/icon/src/index.js
```diff
@@ -0,0 +1,3 @@
+export { default } from './react/index.js'
```
jaketrent: I'm not sure I know what this means.  How is this:

```
export { default } 
```

Different than this:

```
export default
```

packages/focusmanager/src/index.js
```diff
@@ -1,6 +1 @@
-import FocusManager from './react/index.js'
-import * as vars from './vars/index.js'
-import css from './css/index.js'
-
-export default FocusManager
-export { vars, css }
+export { default } from './react/index.js'
```
jaketrent: Good catch on the cleanups here.

packages/avatar/src/index.js
```diff
@@ -1,7 +1,4 @@
-import Avatar from './react/index.js'
-import * as vars from './vars/index.js'
-import * as js from './js/index.js'
-import css from './css/index.js'
-
-export default Avatar
-export { vars, css, js }
+export { default } from './react/index.js'
+export * from './vars/index.js'
+export * from './js/index.js'
```
jaketrent: This approach to export everything in these files to the same namespace feels a little unexpected.  Will there be naming overrides that happen unintentionally here?  

At the same time, I think it paves the way to a refactor I've been wanting to get to in the future:  `vars` feels like a pretty low-value grouping.  I think I'd rather just move vars into js.  And for packages with no vars and just js today, rename vars to js (more general).

packages/link/src/index.js
```diff
@@ -1,6 +1,3 @@
-import Link from './react/index.js'
-import * as vars from './vars/index.js'
-import css from './css/index.js'
-
-export default Link
-export { vars, css }
+export { default } from './react/index.js'
+export * from './vars/index.js'
```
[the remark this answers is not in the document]
EdwardIrby: None of our documentation pointed to main instead we've always pointed users to `
(react|vars|css).js`. However you are right this is breaking I need to go and amend a lot of commits. I was in total autopilot after writing out my steps and forgot that one. Good catch!

packages/avatar/src/index.js
EdwardIrby: A shared namespace would concern me if these were larger packages honestly. Even util is tiny. This is one of the benefits of taking this monorepo approach. We're already renaming the groups when we import these in to other files for example in the `css/index.js` files I came across the following often `import { vars as iconVars } from '@pluralsight/ps-design-system-icon'` and most often the we were using `iconVars.widths`. I figured we could save ourselves some keystrokes and go with `import { widths as iconWidths } from '@pluralsight/ps-design-system-icon'`. 

As for the dropping vars for js. I can do that let me look over current usage and see if I see any potential gotchas.

packages/theme/src/index.js
```diff
@@ -0,0 +1,2 @@
+export * from './vars/index.js'
+export { default, useTheme, withTheme } from './react/index.js'
```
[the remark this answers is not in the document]
EdwardIrby: Not sure I understand you on this one. I think I need some more clarification. The vars are static values in this one and the react component is the default always is unless it's a utility type package. The other 2 named exports here are also react specific functions that rely on react's api.

packages/site/pages/components/position-positions-example.js
```diff
@@ -4,10 +4,27 @@ import core from '@pluralsight/ps-design-system-core'
...
+  AboveRight,
+  BelowLeft,
+  BelowRight,
+  RightOf,
+  LeftOf
+} from '@pluralsight/ps-design-system-position'
 
 import { Head } from '../../src/ui/index.js'
 
+const components = [
```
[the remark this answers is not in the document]
EdwardIrby: Yea I talked to Dane about this Friday. We both agreed most react devs know `Components` have to start capitalized so it makes sense the utility function that matches it would camelCased. This is the only react package that doesn't have a default export I believe, have to check to be sure.  

Still I also see your point which is why I talked to Dane about it. I'll think about this over lunch and and maybe reach back out once I solve this test bug.

packages/avatar/src/index.js
jaketrent: The vars -> js refactor definitely doesn't happen (maybe shouldn't, to keep the diffs smaller and our effort more focused) in this pr.

PR #702, opened 2019-10-31
Fourth esm package set

### LINKED ISSUE #698: OKR: ESM Packages batch 4
- form
- halo
- linearprogress
- note
- row
- starrating
- tab
- text
- verticaltabs
 - viewtoggle

EdwardIrby:
resolves #698

- form
- halo
- linearprogress
- note
- row
- starrating
- tab
- text
- verticaltabs
 - viewtoggle


packages/text/src/index.js
```diff
@@ -0,0 +1,3 @@
+export * from './react/index.js'
```
jaketrent: wait, you *added* this?  I guess the named import wouldn't have worked before, since `main` was pointing to `index.js`, si?

jaketrent: Wait, no, this is `src/index.js`, not `/index.js`.  But then, what uses this?  I don't see it.

jaketrent: I must be missing something, because you've done this pattern many times.  Perhaps it's the late hour, but I'm not making the connection.

EdwardIrby: Oh this is so that the index.js get's built by the build script for both cjs and esm. The main and modules keys in the package.json point to these generated files. This makes it a bit easier to manage paths if they need to change versus having to remember to modify both an index.js and index.esm.js at the package root. 
```json
{
  "main": "dist/cjs/index.js",
  "module": "dist/esm/index.js"
}
```

PR #703, opened 2019-10-31
Fifth esm package set

### LINKED ISSUE #697: OKR: ESM Packages batch 5
- checkbox
- datawell
- datepicker
- dropdown
- errors
- layout
- radio
- searchinput
- switch
- table

EdwardIrby:
resolves #697 

- checkbox
- datawell
- datepicker
- dropdown
- errors
- layout
- radio
- searchinput
- switch
- table


packages/datawell/css.js
```diff
@@ -1 +1 @@
-module.exports = require('./dist/css/index.js')
+module.exports = require('./dist/cjs/css/index.js')
```
jaketrent: I'm glad we have BREAKING CHANGE marked in the commit message.  But it's not very user friendly.  This will go in the changelog.md, and in order to be useful must indicate enough to the user to understand.  

The message you created for button is much better: https://github.com/pluralsight/design-system/blob/master/packages/button/CHANGELOG.md#breaking-changes

Something like that would be nice because it's more specific and understandable.  This is more mysterious.  Though it will create the correct major version bump.

Maybe:

> a named package import now imports the React component

What do you think?

packages/layout/src/css/index.js
```diff
@@ -1,7 +1,3 @@
-import AAsideLayout from './aside-layout.js'
-import AEqualColumnLayout from './equal-column-layout.js'
-import APageHeadingLayout from './page-heading-layout.js'
-
-export const AsideLayout = AAsideLayout
-export const EqualColumnLayout = AEqualColumnLayout
-export const PageHeadingLayout = APageHeadingLayout
+export { default as asideLayoutCSS } from './aside-layout.js'
```
jaketrent: Would it be simpler to just combine these into a single `css/index.js` file (and built .css and .scss file)?

Ie, `export  * from './aside-layout.js`, etc

Ehhh, probably not, right, because there aren't named exports in these files like `aside-layout.js`, nm...

packages/layout/src/index.js
```diff
@@ -0,0 +1,3 @@
+export * from './react/index.js'
+export * from './vars/index.js'
+export * from './css/index.js'
```
jaketrent: I think I'd namespace the css like the other packages, ie:

`export { * as css } from './css/index.js'`

Does that look right?  lol

And then once I did that, I think I'd renamed things like `asideLayoutCSS` to `asideLayout`.  

I'd do this because it more closely mimics other packages.  What do you think?

packages/errors/index.js
```diff
@@ -1,4 +0,0 @@
-const css = require('./css')
```
jaketrent: Ummm, can we get rid of this file?  Seems like it's still referenced in `errors/package.json`.

packages/layout/src/index.js
jaketrent: I'm tempted to do that with the vars too.  The problem there is that `AsideLayout` (react component) and `asideLayout` (var object) are so close in naming (only differing in case).  

But if we did that, the problem created would be that it's *less* consistent with the other packages.

This makes me want to simply combine `vars/*.js` into a single `vars/index.js`.  With some var renaming (eg, `const counts` -> `const equalColumnLayoutCounts`), that would work just fine.

packages/datawell/css.js
EdwardIrby: For current usage, documented patterns, this doesn't really break anything for anyone honestly. Because the patterns we use always point to `(react|css|js|vars).js`.  Those cjs files still exist as they always have. I merely updated the path to on the root re-exports so that this upgrade was backward compatible for people using the CJS import pattern we've been following until now.

All of the ESM mods will only effect someone who is explicitly importing the package using the main module which until now was always an object with one or more of the following keys:
- react
- css
- vars
- js


However I have been working on a release post that will explain how to migrate to the new import pattern to take advantage of treeshaking. The important thing is to make that clear in the release notes so I'll probably run them by Dane on Monday to get his feedback.

packages/layout/src/css/index.js
EdwardIrby: Sorta but I think we agreed to tackle this as separate PR and not break the current patterns. Want to pair on after your leadership offsite?

packages/layout/src/index.js
EdwardIrby: Similar to position package from an earlier PR devs seem to be familiar with camel for functions/vars and PascalCase for react components. However if this proves to be an issue I say we reevaluate. Also  think maybe adding typings would help if more teams start moving to typescript.

packages/errors/index.js
EdwardIrby: looks like it is deleted and no longer referenced. I'll check to be sure.

PR #715, opened 2019-11-08
a couple refinements for verticaltab

danethurber:
- larger hit areas to show better focus state
- hover states
  - more contrast between
- animate active state for fun
- organize components into separate files
- add a gitprime story

@sweeting i made some changes to your design to give more contrast between the active/focus/inactive states

Nov-08-2019 09-59-20


sweeting:
Looks good. I made these changes in Figma library too.

PR #718, opened 2019-11-08
refactor(position): canUseDOM function updated to align with React

### LINKED ISSUE #691: Position: `inNode` default props doesn't seem to work as expected
we set the `inNode` default props based on checking if we're in the browser but it doesn't seem to be worked. exploration needed

reference thread here -> https://pluralsight.slack.com/archives/C70HPSJPP/p1571933105073800

EdwardIrby:
close  #691

### DIFF
--- a/packages/position/src/react/index.js
+++ b/packages/position/src/react/index.js
@@ -86,7 +86,7 @@ const Position = React.forwardRef((props, forwardedRef) => {
   }, [updateStyle, when])
 
   return (
-    <React.Fragment>
+    <>
       {target ? child : React.cloneElement(child, { ref })}
 
       {createUniversalPortal(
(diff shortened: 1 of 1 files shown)


packages/position/src/react/index.js
```diff
@@ -86,7 +86,7 @@ const Position = React.forwardRef((props, forwardedRef) => {
   }, [updateStyle, when])
 
   return (
-    <React.Fragment>
+    <>
```
danethurber: have we used the shorthand fragment syntax before. please double check that it compiles correctly before deployment

EdwardIrby: ??? I didn't change that. Weird??? I'll check

PR #725, opened 2019-11-22
Use VerticalTabs in site

jaketrent:
WIP


jaketrent:
It would be cool if dog fooded this.  Buuuut, no body cares if we have Vert Tab in our site.  And we're probably changing layout someday.... 

Anyway, I'm disappointed that we can't use the DS component and it makes me wonder about the generality of the component, but for today, closing this.  

No merge for you!

PR #728, opened 2019-12-03
Feat/max width dropdown

### LINKED ISSUE #487: dropdown: Add max width restriction
## Enhancement Request
Please add the ability to restrict the width of the dropdown to a max width.

### Desired behavior
Any text that expends past the max width should be truncated.

Support you wish you had.

### Today's behavior
Currently is you add a max width the text is extending out of the component.

### The Value Add
Designs where content sits to the right is currently being pushed right based of however long the dropdown need to be, with this enhancement we can decide how we wish to render content verses being subject to the dynamic data. 

Why this should be added/changed in the Design System. Where we could use it. Basically anywhere we have content sitting to the right of a dropdown. Right now the Home team is actively investigating implementing a design like this.

### The Tradeoffs

What we give up or deoptimize to get it. The costs.
The only trade off that I can think of is that on mobile this could cause some spacing concerns so this property should maybe not be available via the mobile phone.

EdwardIrby:
resolves #487

After merging master eslint kept throwing errors. I  updated eslint packages and config since commitizen started throwing errors decided to ignore key on site and stories and correct the other errors in actual components.


.eslintrc.js
```diff
@@ -25,5 +25,13 @@ module.exports = {
     'react-hooks/exhaustive-deps': WARNING,
     'react-hooks/rules-of-hooks': ERROR,
     'react/jsx-no-bind': [IGNORE]
-  }
+  },
+  overrides: [
+    {
+      files: ['packages/site/pages/**/*.js', '**/__stories__/*.js'],
+      rules: {
+        'react/jsx-key': [IGNORE]
```
jaketrent: why ignore this rule in site and stories?

EdwardIrby: Because there are 100+ places where  key has not been used. I can go through and refactor each but from now on we will have to apply a key when writing stories. Thoughts?

danethurber: WARNING is probably better than ignore

jaketrent: Edward, it sounds like we have lint debt in the stories and now we're turning off a rule because keeping the rule isn't worth it to change the 100+ places.  If that's the case, should we be turning off eslint generally for stories or removing this rule specifically from the greater codebase (If the ROI on the rule or the code linted isn't in our favor.)?

EdwardIrby: Will go with WARNING for now

EdwardIrby:
Going to close and resubmit after branching from current master. I'll make a ticket for fixing repo linting errors as well.

PR #731, opened 2019-12-03
collapsible: animation & height issues

BenGu3:
### What You're Solving

1. Fixes close animations not working. The bug was caused by the height
not being set to the calculated auto height before being set to zero. Fixed
by not setting height to 0 before calculating the height.

1. Fixes the container not resizing with its content when it starts open. The
bug was caused by the height not being set to "auto" when it starts opened.
Fixed by resolving waitForHeightTransitionToEnd() if container is already at
target height.

1. Fixes the drawer breaking if user opens while it's closing or vice versa.
The bug was caused by not checking if the collapse was in the expected
state after the animation finishes. Fixed by checking the measured height
of the element after the animation finished and only doing close/open cleanup
code if the container is closed/opened.

### How to Verify

1. When closing the collapsible it should animate smoothly instead of snapping shut.

1. Create a collapsible that starts open. Modify the content or the height of the content inside the container. The container should update its height accordingly instead of staying the same height.

1. This wasn't a bug before but we had to add some code to prevent it from happening with our other changes. When closing the collapsible, open it while it's animating closed. It should render correctly.

### DIFF
--- a/packages/collapsible/src/__specs__/index.spec.js
+++ b/packages/collapsible/src/__specs__/index.spec.js
@@ -12,7 +12,7 @@ describe('Collapsible', () => {
     )
     const wrapper = container.querySelector('[aria-hidden]')
     expect(wrapper).toHaveAttribute('aria-hidden', 'false')
-    expect(wrapper).not.toHaveStyle(`height: 0`)
+    expect(wrapper).toHaveStyle(`height: auto`)
     rerender(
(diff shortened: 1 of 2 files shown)


EdwardIrby:
@danethurber This does not appear to break any of the vertical tab stories. However I wanted you to take a look before merging due to flow usage.

PR #732, opened 2019-12-05
fix(dropdown): dev can set width thus truncating long input & options

### LINKED ISSUE #487: dropdown: Add max width restriction

EdwardIrby:
resolves #487

### DIFF
--- a/packages/util/src/useCombinedRefs.js
+++ b/packages/util/src/useCombinedRefs.js
@@ -1,0 +1,17 @@
+import { useRef, useEffect } from 'react'
+
+export const useCombinedRefs = (...refs) => {
+  const targetRef = useRef()
+  useEffect(() => {
+    refs.forEach(ref => {
+      if (!ref) return
+      if (typeof ref === 'function') {
+        ref(targetRef.current)
+      } else {
+        ref.current = targetRef.current
+      }
+    })
+  }, [refs])
+
+  return targetRef
+}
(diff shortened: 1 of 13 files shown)


packages/util/src/useCombinedRefs.js
```diff
@@ -0,0 +1,17 @@
+import { useRef, useEffect } from 'react'
+
+export const useCombinedRefs = (...refs) => {
+  const targetRef = useRef()
+  useEffect(() => {
+    refs.forEach(ref => {
+      if (!ref) return
+      if (typeof ref === 'function') {
+        ref(targetRef.current)
```
danethurber: we probably don't need to support the older ref api. you can decided to leave it in or take it out

EdwardIrby: Even though traditional class component callback refs may not be in use there is a new pattern using useCallback that is recommended for [measuring the dom for example ](https://reactjs.org/docs/hooks-faq.html#how-can-i-measure-a-dom-node). I'm going to leave this in place and add a test.

PR #735, opened 2019-12-05
Refactor/code patterns

### LINKED ISSUE #723: Audit and update components for latest code style

danethurber:
resolve  #723

- update to use new style/pattern around glamor import and style composing
- spread props based on gitprime request
- update to use forwardRef where noticed


packages/card/src/react/index.js
```diff
@@ -9,77 +9,114 @@ import { sizes as iconSizes } from '@pluralsight/ps-design-system-icon'
...
 
 if (typeof window !== 'undefined') polyfillFocusWithin(document)
 
 const styles = {
-  actionBar: ({ actionBarVisible: visible, fullOverlay }) =>
-    glamor.css({
-      ...css['.psds-card__action-bar'],
-      ...(fullOverlay &&
+  actionBar: ({ actionBarVisible: visible, fullOverlay }) => {
+    const label = 'psds-card__action-bar'
```
danethurber: i used a variable to store the base class names in a few places to make refactoring easier. i used the new label because in the future we might use the labels feature of glamor like so


> css({ label })

packages/avatar/src/react/index.js
```diff
@@ -9,10 +9,10 @@ import { getColorByName, getInitials, transformSrc } from '../js/index.js'
 
 const styles = {
   avatar: ({ size }) =>
-    css({
-      ...stylesheet['.psds-avatar'],
-      ...stylesheet[`.psds-avatar--size-${size}`]
-    }),
+    compose(
```
EdwardIrby: Couldn't this also be written as `css(stylesheet['.psds-avatar'], stylesheet[.psds-avatar--size-${size}])` Is there an advantage to using compose? Should we be using compose all the time now?

PR #737, opened 2019-12-05
feat(screenreaderonly): implementation

### LINKED ISSUE #736: ScreenReaderOnly component
i just wrote one in gitprime's repo for reference

https://github.com/Git-Prime/pluralsight-ds-ext-gp/blob/master/src/components/ScreenReaderOnly/index.tsx

danethurber:
requested by our new accessibility czar.

resolves #736


jaketrent:
I love the speed from request to something to commit.  And I'll probably love the code.  I just haven't looked yet.  

What it seems like didn't happen was fielding a request through the front door, triaging as a team, and deciding how to choose to respond as a team.  Of course, we can do that now, but with a PR already done, the decision process is far advanced.

My request for the future would be that if we get requests that they go into our triage process, probably with a GH issue.  Then we can discuss, learn the rationale for the request as a group, and decide how to approach the request given shared ownership for the DS product.

Of course, I could be wrong in my interpretation of the situation and would love to learn more, especially if my request doesn't seem appropriate.  This is probably a good retro subject too.  I'll add it there.

packages/screenreaderonly/src/react/index.js
```diff
@@ -0,0 +1,27 @@
...
+import stylesheet from '../css/index.js'
+
+const styles = {
+  screenReaderOnly: () => css(stylesheet['.psds-screen-reader-only'])
+}
+
+const ScreenReaderOnly = React.forwardRef((props, forwardedRef) => {
+  const { as: Tag, ...rest } = props
+  const ref = React.useRef()
+  React.useImperativeHandle(forwardedRef, () => ref.current)
```
EdwardIrby: Why are we including this if we are not using the ref internally?

packages/site/pages/components/screenreaderonly.js
```diff
@@ -0,0 +1,34 @@
...
+      <Code language="bash">
+        npm install @pluralsight/ps-design-system-screenreaderonly
+      </Code>
+
+      <P>Include a React component in your project:</P>
+      <Code language="javascript">
+        import Typeahead from '@pluralsight/ps-design-system-screenreaderonly'
+      </Code>
+
+      <SectionHeading>Usage example</SectionHeading>
```
EdwardIrby: In a future version of docs should we include more detailed explanation of why and how to use this?

packages/screenreaderonly/src/react/__stories__/index.story.js
```diff
@@ -0,0 +1,12 @@
+import { storiesOf } from '@storybook/react'
+import React from 'react'
+
+import ScreenReaderOnly from '../index.js'
+
+storiesOf('ScreenReaderOnly', module).add('basic', _ => (
+  <div>
+    <p>you should not see the following</p>
+
+    <ScreenReaderOnly />
```
jaketrent: I do see the text.  The story usage is surprising.  I would have thought this component would wrap children.

Screen Shot 2019-12-05 at 3 37 18 PM

packages/screenreaderonly/react.d.ts
```diff
@@ -0,0 +1,24 @@
+/// <reference types="react" />
```
jaketrent: Epic types.  I couldn't tell you if they're right or not.  Perhaps you could explain them.

packages/site/pages/components/screenreaderonly.js
```diff
@@ -0,0 +1,34 @@
...
+      <br />
```
jaketrent: Maybe

```suggestion
      <P>Wrap content that you want the visually impaired to read through a screenreader.  It will be heard through screenreader audio but not visually present on the page.</P>
```

packages/site/pages/components/screenreaderonly.js
```diff
@@ -0,0 +1,34 @@
...
+  Code,
+  Content,
+  P,
+  PageHeading,
+  SectionHeading
+} from '../../src/ui/index.js'
+
+export default () => (
+  <Chrome>
+    <Content title="ScreenReaderOnly">
```
jaketrent:
```suggestion
    <Content title="Screen Reader Only">
```

danethurber:
@jaketrent you're totally right in your concerns. i guess i meant this pr as more of the team conversation but didn't quite make that clear. also knowing about the screen reader class in core is some knowledge i didn't have

for context. i was asked if we had a global css for screen reader only. i didn't know of one and said maybe we could add a component for teams. let's discuss tomorrow about if the current class is enough or if this added component is valuable

packages/screenreaderonly/src/react/__stories__/index.story.js
danethurber: oops. i didn't do that right

packages/screenreaderonly/src/react/index.js
danethurber: because i copy/pasted. i'll fix

packages/site/pages/components/screenreaderonly.js
danethurber: yes. i was being lazy 😬 i'll update

jaketrent:
Dane, thanks for the extra context.  I like that you intent was to have a conversation.  I'd be happy to chat more tomorrow.

packages/screenreaderonly/react.d.ts
danethurber: yep these are the gnarliest types i've written. there is a prop on the component call `as` that allows a string with the element name. these type enforce that the `as` prop is a valid html or svg element. also any additional html or react attributes have to match what is valid for the html or svg element.

in typescript this could potentially replace the filterReactProps. it's different but accomplishing a similar need

PR #749, opened 2019-12-16
Fix/empty state svg export

### LINKED ISSUE #741: EmptyState: include svgs files in dist bundle
for reference: https://pluralsight.slack.com/archives/C70HPSJPP/p1576248922001100

cc @willsoto

EdwardIrby:
resolves  #741


packages/emptystate/package.json
```diff
@@ -10,12 +10,13 @@
...
   },
   "scripts": {
-    "build": "run-s build:svg build:js build:cjs build:css",
+    "build": "run-s build:svg build:js build:cjs build:css copy:svg",
     "build:cjs": "babel --root-mode upward src --out-dir dist/cjs",
     "build:css": "build-css --useGlamor -i dist/cjs/css/index.js",
     "build:js": "cross-env ESM=true babel --root-mode upward src --out-dir dist/esm",
     "build:svg": "svgr --filename-case kebab --ext \"dist.js\" -d src/react/illustrations src/svg",
     "build:watch": "npm run build:js -- --watch",
+    "copy:svg": "node copySvgs.js",
```
danethurber: nitpick: i think we've been using dashed filename. i suggest you change `copySvgs.js` to `copy-svgs.js`

packages/emptystate/copySvgs.js
```diff
@@ -0,0 +1,14 @@
+const ncp = require('ncp').ncp
```
danethurber: why use `npc` instead of just using the `cp -r` command from the terminal. is this better because it works on windows - or are there other reasons

EdwardIrby: Yep, in the past Jake reminded me as this is an opensource project we should try to be good citizens. Otherwise I would of gone with cp -r.

packages/emptystate/package.json
EdwardIrby: Will do!

PR #753, opened 2019-12-16
Layout page width

jaketrent:
Adds new LayoutPageWidth component

Supports:
- Constrains max width when prism won't
- Standard page-level gutters = 24px; 48px above 768px
- Float in center space


packages/layout/src/css/page-width-layout.js
```diff
@@ -0,0 +1,17 @@
+import * as core from '@pluralsight/ps-design-system-core'
```
EdwardIrby: Probably a good idea to use this syntax here `import { layout } from '@pluralsight/ps-design-system-core'`. We want to stop using star to ensure effective treeshaking

PR #754, opened 2019-12-17
storybook chaos addon

danethurber:
proposal: new storybook addon to bring in some chaos engineering

this is the start of a new addon.I integrated it into the carousel package for testing

the addon currently does a few basic things but i could see it doing a lot more in the future. at the moment it does two things:
- changes the global font size and line height to see how our components would respond
- changes the text to mimic translations that are longer that the source language and introduces diacritic marks

thoughts?


jaketrent:
Getting this on npm run build:

```
~/dev/design-system/packages/storybook-addon-chaos(feat/chaos ✔) nrb

> @pluralsight/ps-design-system-storybook-addon-chaos@0.0.0 build /Users/jaketrent/dev/design-system/packages/storybook-addon-chaos
> babel --root-mode upward src --out-dir dist

sh: babel: command not found
```

jaketrent:
Fixed it with clean/bootstrap

packages/storybook-addon-chaos/src/components/Panel.js
```diff
@@ -0,0 +1,47 @@
...
+      <div {...stylesheet.inner}>
+        {enabled ? (
+          <h2>Chaos Enabled</h2>
+        ) : (
+          <Button
+            appearance={Button.appearances.primary}
+            onClick={() => setEnabled(true)}
+            size={Button.sizes.large}
+          >
+            enable chaos
```
jaketrent: toggle chaos would be nice.

PR #756, opened 2019-12-17
[wont work, but review problem, plz] Theme prop to enable bleutrals

### LINKED ISSUE #745: New Colors Theme Flag
Add new theme option for colors so that button, text, and link can be phased into new colors.

jaketrent:
Resolves #745 

I think it'd be a little better to transmit *multiple* values from the theme provider.  But this would require changing the consumption.  Instead, what I did was modify the themeName value going down in context to consumers.

If enabled:
"dark" -> "darkBleutrals"
"light" -> "lightBleutrals"

This allows the api not to change.  

What it will require....

Blast, this won't work.  Because what about those components that need "light" to come through and are not updated to work with bleutrals?  We shouldn't need to update them.  Yet, their themeName is going to be modified so it won't match corresponding css in those components.

We need another idea... Something that doesn't modify themeName.  Probably means a separate value.  And that probably means breaking change on Theme.  Then, what's the upgrade process there?  Same problem, different impl detail in this scenario: what about those components that are not using the new Theme version (you can only have one in your project)?

Too bad we didn't return a more flexible data structure like an object from this context provider!

Anyone have an idea for a solution to this?

### DIFF
--- a/packages/theme/src/react/index.js
+++ b/packages/theme/src/react/index.js
@@ -19,34 +19,46 @@ export function withTheme(BaseComponent) {
   const name = getDisplayName(BaseComponent)
 
   const Forwarded = React.forwardRef((props, ref) => {
-    const themeName = useTheme()
-    return <BaseComponent {...props} ref={ref} themeName={themeName} />
+    const { enableBleutrals, ...rest } = props
+    const themeName = useTheme({ enableBleutrals })
+    return <BaseComponent {...rest} ref={ref} themeName={themeName} />
   })
   Forwarded.BaseComponent = BaseComponent
   Forwarded.displayName = `withTheme(${name})`
+  Forwarded.propTypes = {
+    enableBleutrals: PropTypes.bool
+  }
+  Forwarded.defaultProps = {
+    enableBleutrals: false
+  }
 
   return hoistNonReactStatics(Forwarded, BaseComponent)
 }
 
-export function useTheme() {
+export function useTheme(config) {
   const themeName = React.useContext(ThemeContext)
-  return themeName
+  const bleutralsAdjustedThemeName =
+    config && config.enableBleutrals ? themeName + 'Bleutrals' : themeName
+  return bleutralsAdjustedThemeName
 }
 
 // NOTE: anything that uses <Theme /> from context only (no withTheme), will be broken with this new impl
 // TODO: find this set of components ^ and update
 export default function Theme(props) {
+  const value = props.enableBleutrals ? props.name + 'Bleutrals' : props.name
   return (
-    <ThemeContext.Provider value={props.name}>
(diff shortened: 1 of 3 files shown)


packages/theme/src/react/index.js
```diff
@@ -19,34 +19,46 @@ export function withTheme(BaseComponent) {
...
+  return bleutralsAdjustedThemeName
 }
 
 // NOTE: anything that uses <Theme /> from context only (no withTheme), will be broken with this new impl
 // TODO: find this set of components ^ and update
 export default function Theme(props) {
+  const value = props.enableBleutrals ? props.name + 'Bleutrals' : props.name
   return (
-    <ThemeContext.Provider value={props.name}>
+    <ThemeContext.Provider value={value}>
```
EdwardIrby: What if this value wasn't primitive one but rather a object, one with a key **theme** `light` or `dark` and another key **bleutrals** `true` or `false`? Then all anyone would need to do to upgrade theme in addition to phase 1 components.  The **withTheme** function could pass through this new value in addition to `themeName` that could be used in refactoring button link and text during phase 1.

packages/theme/src/react/index.js
```diff
@@ -19,34 +19,46 @@ export function withTheme(BaseComponent) {
...
 
   return hoistNonReactStatics(Forwarded, BaseComponent)
 }
 
-export function useTheme() {
+export function useTheme(config) {
   const themeName = React.useContext(ThemeContext)
-  return themeName
+  const bleutralsAdjustedThemeName =
+    config && config.enableBleutrals ? themeName + 'Bleutrals' : themeName
```
EdwardIrby: Should this be bluetrals if we're going to be doing an entire new color palette?

packages/theme/src/react/index.js
jaketrent: Agreed, easier if value was already an objective.  But since it's already a primitive, changing to an obj would be a breaking change.  And since essentially all components rely on Theme, how would one upgrade Theme in a project without upgrading all the rest of the DS components (and that means *we* also have to upgrade all of them)?

packages/theme/src/react/index.js
jaketrent: Good point.  This is more than bleutrals.  What is the "new palette" called?  Do we have a name designers are using, @sweeting ?

jaketrent:
I think my fav alternate solution is simply to create a new component:

- Context Provider like Theme, but separate
- Provides the object of feature flags, straight from our feature service 
- The DS components will be compatible with it for occasions when needed
- Teams might find such a component generally useful since we do feature flagging across the product.
- Doesn't require api changes to theme
- The upgrade path for teams and for us seems straightforward:  Start using the new component.

PR #758, opened 2019-12-18
Feature Flags component

jaketrent:
Resolves: #745 

Replaces: #756 

Reasoning for this solution in #756 comments.

### DIFF
--- a/packages/featureflags/react.d.ts
+++ b/packages/featureflags/react.d.ts
@@ -1,0 +1,27 @@
+/// <reference types="react" />
+
+interface Flags {
+  [name: string]: boolean | string | number
+}
+
+interface Props {
+  flags: Flags
+}
+
+interface FeatureFlagsContext {
+  flags: Flags
+}
+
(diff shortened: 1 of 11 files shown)


packages/featureflags/react.d.ts
```diff
@@ -0,0 +1,27 @@
+/// <reference types="react" />
```
jaketrent: I want to add these, they make sense.  But I'm not a Typescript compiler.  How do we verify these are correct?

danethurber: that's the question? we'd have to have some way to validation it. maybe something exists we can integrate?

jaketrent:
Any concern about the reasoning that got us here?  Or the general approach?

danethurber:
@jaketrent i love the simplicity and do think that other teams would find by using this approach

for the new colors...will we need to write a guide about how to provide the context value. we'll need to ensure that value matches up with what our button/link/etc components expect. does our team own that flag in launch darkly?

PR #760, opened 2019-12-19
Feat/add icons

### LINKED ISSUE #739: Add all core Icons
Now that tree shaking is a thing. Let's get all the core icons in the DS

EdwardIrby:
resolves #739


packages/icon/src/react/icons/index.js
```diff
@@ -13,63 +25,132 @@ export { CaretDownIcon } from './CaretDownIcon.dist.js'
...
 export { HomeIcon } from './HomeIcon.dist.js'
 export { InfoIcon } from './InfoIcon.dist.js'
 export { InteractiveIcon } from './InteractiveIcon.dist.js'
+export { IqIcon } from './IqIcon.dist.js'
 export { LinkIcon } from './LinkIcon.dist.js'
 export { ListIcon } from './ListIcon.dist.js'
+export { LocationIcon } from './LocationIcon.dist.js'
+export { LockFillIcon } from './LockFillIcon.dist.js'
 export { LockIcon } from './LockIcon.dist.js'
-export { LogoutIcon } from './LogoutIcon.dist.js'
```
jaketrent: This will require a BREAKING CHANGE in the commit to gen a new major version on release.

.gitignore
```diff
@@ -43,3 +43,6 @@ jspm_packages
 .env*
 
 .cache
+
+
+.DS_Store
```
jaketrent: I would remove this for these reasons

https://jaketrent.com/post/gitignore-system-files/

packages/icon/src/react/icons/index.js
EdwardIrby: Done!

PR #773, opened 2020-01-13
Feat/colors 2020

### LINKED ISSUE #744: New Colors
Add new color palette to core

EdwardIrby:
resolves #744 
Please note this does not contain final button styles just a storybook example of the flag working at http://localhost:6006/?path=/story/button-appearance--primary


packages/storybook-addon-featureflags/__mocks__/index.js
```diff
@@ -0,0 +1,5 @@
+var React = require('react')
```
danethurber:
```suggestion
const React = require('react')
```

packages/storybook-addon-featureflags/__mocks__/index.js
```diff
@@ -0,0 +1,5 @@
...
+
+const MockTheme = props => React.createElement(React.Fragment, props)
```
danethurber:
```suggestion
const MockFeatureFlagDecorator = props => React.createElement(React.Fragment, props)
```

packages/storybook-addon-featureflags/__mocks__/index.js
```diff
@@ -0,0 +1,5 @@
...
+
+module.exports = storyFn => React.createElement(MockTheme, null, storyFn())
```
danethurber:
```suggestion
module.exports = storyFn => React.createElement(MockFeatureFlagDecorator, null, storyFn())
```

packages/button/src/css/index.js
```diff
@@ -50,7 +50,14 @@ export default {
       }
     }
   },
-
+  '.psds-button--psds2020Colors': {
```
danethurber: a couple naming options that might be cleaner and follow BEM standards

`psds-button--2020-colors`
`psds-button--feature-2020-colors`

🤷‍♂

packages/core/js/colors.2020.js
```diff
@@ -0,0 +1,151 @@
...
+  4: '#A883F8',
+  5: '#956CED',
+  6: '#8359DF',
+  7: '#7048C6',
+  8: '#5934AC',
+  9: '#482592',
+  10: '#351973'
+}
+
+export const colorsTextIcon = {
```
danethurber: wondering why these variables have a `colors` prefix but the vars above do not.

what if you define/export the variables in this file without a prefix and then in the `packages/core/js/index.js` you can rename the exported object like we do the other colors file

`export { default as colors2020 } from './colors.2020.js'`

I coming in a little late to the feature here. just thought this might be a better pattern

packages/storybook-addon-featureflags/src/constants.js
```diff
@@ -0,0 +1,8 @@
+export const FEATURE_FLAG_ID = 'ps-design-system-storybook-addon-featureflags'
```
danethurber:
```suggestion
export const ADDON_ID = 'ps-design-system-storybook-addon-featureflags'
```

packages/storybook-addon-featureflags/src/constants.js
```diff
@@ -0,0 +1,8 @@
...
+export const PANEL_ID = `${FEATURE_FLAG_ID}/panel`
+
+export const EVENTS = {
+  change: `${FEATURE_FLAG_ID}:change`
```
danethurber:
```suggestion
  change: `${ADDON_ID}:change`
```

packages/storybook-addon-featureflags/src/constants.js
```diff
@@ -0,0 +1,8 @@
+export const FEATURE_FLAG_ID = 'ps-design-system-storybook-addon-featureflags'
+export const PANEL_ID = `${FEATURE_FLAG_ID}/panel`
```
danethurber:
```suggestion
export const PANEL_ID = `${ADDON_ID}/panel`
```

danethurber:
Screen Shot 2020-01-13 at 2 28 08 PM

I get an error when running the storybook for `<Button />` on my machine

packages/button/src/css/index.js
EdwardIrby: Maybe will have to refactor a lot of other commit for that to work.

packages/core/js/colors.2020.js
EdwardIrby: They are tokens not just variables and this is to simply help identified them. Also should we be making an effort to reduce the size of bundles by using tree shaking patterns?

EdwardIrby:
> Screen Shot 2020-01-13 at 2 28 08 PM
> 
> I get an error when running the storybook for `<Button />` on my machine

Yep that was what was causing test to fail but wasn't showing up in my console.

packages/core/js/colors.2020.js
EdwardIrby: Maybe it should be colorBlue instead of blue for example??

jaketrent: Or maybe the `colorsTextIcon` is just `textIcon`?

jaketrent: (use the abstraction of the package name for the "color" categorization)

PR #776, opened 2020-01-14
Feat/colors 2020 text

EdwardIrby:


packages/text/src/css/index.js
```diff
@@ -28,7 +28,16 @@ export default {
     background: transparentize(0.5, core.colors.black),
     border: `1px solid ${core.colors.black}`
   },
-
+  [`.psds-text__code.psds-theme--${themeNames.light}.psds-button--2020-colors`]: {
+    color: '#0072be',
```
jaketrent: no var for this one?

EdwardIrby: > I wasn't able to detect much of a visual change from the storybook side, but everything looked well enough. I haven't made myself very aware of what the text package should do in response to the new palette. If there's something drastic I should have caught, I'd love to be informed of it.

Sorry not really a drastic change have to inspect to even notice at least for me.

EdwardIrby: No we need to revisit this as a team according to Brian as he's not familiar with these colors. I can set up some time to go over it.

EdwardIrby:
> I wasn't able to detect much of a visual change from the storybook side, but everything looked well enough. I haven't made myself very aware of what the text package should do in response to the new palette. If there's something drastic I should have caught, I'd love to be informed of it.

Sorry not really noticeable on this set at all unless you inspect.

packages/text/src/css/index.js
sweeting: I've updated the spreadsheet so every color should have a variable name. Let me know if we need to sync up on any naming differences. https://docs.google.com/spreadsheets/d/1TGxJJFTmVpAcCjcJNvxjaQb6r0Ys-PRZh_Oj__TLmJk/edit#gid=0

jaketrent: @sweeting , is this the right hext, #0072be, for the inline code block on light them in the new 2020 palette?  If so, that's enough for me.  

My original question on this line was simply because as we are going toward 2020, it seems like all the colors have had var names.  It's just a double check.  I'm not implying that they have to have var names.

sweeting: For the inline code block on dark theme we can use Blue3, and the light theme can use Blue8.

jaketrent: Awesome, Brian, thanks for the reply.  Looks like this was merged w/o that change.  I'll make a new card.

PR #777, opened 2020-01-14
feat(icon): enables treeshaking when used by parcel or webpack

EdwardIrby:
Treeshaking for Icons is not be working. After creating a test project I concluded it's two things adding `sideEffects: false` to our repo and `optimization: {usedExports: true,}` to a webpack config for webpack 4. As for Parcel I haven't tested yet but according to their docs just the sideEffects part is needed only.

### DIFF
--- a/packages/site/pages/components/icon.js
+++ b/packages/site/pages/components/icon.js
@@ -78,7 +78,11 @@ export default _ => (
       <Code language="bash">
         npm install @pluralsight/ps-design-system-icon
       </Code>
-
+      <P>
+        In your package webpack config in order to enable treeshaking add the
+        following config option:
+      </P>
+      <Code>{`optimization: {usedExports: true }`}</Code>
       <P>Include a React component in your project:</P>
       <Code language="javascript">
         {`import {CodeIcon, /* OtherNamedIcons */} from '@pluralsight/ps-design-system-icon'`}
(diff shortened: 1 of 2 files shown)


packages/site/pages/components/icon.js
```diff
@@ -78,7 +78,11 @@ export default _ => (
       <Code language="bash">
         npm install @pluralsight/ps-design-system-icon
       </Code>
-
+      <P>
+        In your package webpack config in order to enable treeshaking add the
+        following config option:
+      </P>
+      <Code>{`optimization: {usedExports: true }`}</Code>
```
jaketrent:
```suggestion
      <Code>{`optimization: { usedExports: true }`}</Code>
```

jaketrent:
for ref, docs: https://webpack.js.org/guides/tree-shaking/

packages/site/pages/components/icon.js
```diff
@@ -78,7 +78,11 @@ export default _ => (
       <Code language="bash">
         npm install @pluralsight/ps-design-system-icon
       </Code>
-
+      <P>
+        In your package webpack config in order to enable treeshaking add the
+        following config option:
+      </P>
+      <Code>{`optimization: { usedExports: true }`}</Code>
```
jaketrent: Won't teams usually be using `mode: "production"` instead, and that would enable the tree shaking optimization?

see: https://webpack.js.org/guides/tree-shaking/#conclusion

EdwardIrby: True I should reference if they want tree shaking in dev non-prod mode. I make that change.

PR #791, opened 2020-01-21
refactor(tooltip): apply 2020 colors

jaketrent:
I removed transparentize usage.  It will pkg size and runtime cost.  It seems not super important to have had it here.  The token color was black, it was on a box shadow.  It did not change in the 2020 migration and seems unlikely to change in the future.  I'd love to hear any reasons you have for replacing it.


packages/tooltip/src/css/index.js
```diff
@@ -9,10 +8,7 @@ const diamondHeight = getDiamondHeight(tailDimension, tailDimension) + 'px'
 const tailTransformTop = `translate(calc(((${diamondHeight} - ${tailDimension})) / 2), calc(50% + (${diamondHeight} - ${tailDimension}))) rotate(45deg)`
 const tailTransformBottom = `translate(calc(((${diamondHeight} - ${tailDimension})) / 2), -50%) rotate(45deg)`
 
-const boxShadow = `0px 1px ${shadowBlurWidth} ${transparentize(
-  0.5,
-  core.colors.black
-)}`
+const boxShadow = `0px 1px ${shadowBlurWidth} rgba(0, 0, 0, 0.5)`
```
danethurber: 👍

PR #794, opened 2020-01-21
feat(core): add status color vars/tokens

danethurber:


packages/core/js/colors.2020.js
```diff
@@ -163,3 +163,11 @@ export const colorsPrimaryAction = {
   textDarkTheme: colorsBlue[5],
   textLightTheme: colorsBlue[7]
 }
+
+export const colorsStatusError = colorsRed.base
```
jaketrent: Do we want these grouped?  There are categorical groupings for others like `colorsTextIcon` or `colorsBorder` or `colorsBackgroundUtility`.  Seems like a grouping would match these closer.

I am torn, however, because single var exports seem better for tree shaking.

But since we're down the grouping road, I think we shouldn't mix them.  Instead, we should decide on one or the other.

What do you think?

danethurber: i could see that. are you thinking -> 

```
export const colorsStatus = {
  error: colorsRed.base,
  info: colorsBlue.base,
  success: colorsGreen.base,
  warning: colorsYellow.base
}
```

jaketrent: Yes, exactly.

PR #795, opened 2020-01-21
refactor(table): apply 2020 color palette

jaketrent:
asdf


jaketrent:
please have a looksy at the 2 new, latest commits.

danethurber:
@jaketrent double checking.

emphasized text looks like normal cell text?
also - should links be blue?

jaketrent:
@danethurber emphasized text should be normal color, but bold.  Links should be normal color, underline on hover.  Both per Brian as of today.  (as long as I understood him correctly.)

PR #796, opened 2020-01-21
Feature/halo colors2020

danethurber:


packages/halo/src/css/index.js
```diff
@@ -20,7 +20,7 @@ export default {
       position: 'absolute',
       borderWidth: '3px',
       borderStyle: 'solid',
-      borderColor: core.colors.blue,
+      borderColor: colorsStatus.info,
```
jaketrent: Just wondering if these are the right semantics.  Do we want to think about the focus indicator as the "primary action" or an "info status"?  I'm currently thinking primary action.  What was your thinking for info status?

PR #797, opened 2020-01-21
feat(checkbox): use 2020 colors

danethurber:


packages/checkbox/src/css/index.js
```diff
@@ -20,13 +20,13 @@ export default {
...
-    background: core.colors.gray05,
+    border: `2px solid ${core.colorsTextIcon.lowOnDark}`,
+    background: core.colorsBackgroundDark[1],
     color: core.colors.white
   },
   [`.psds-checkbox__square.psds-theme--${themeNames.light}`]: {
-    background: core.colors.white,
-    borderColor: core.colors.gray04,
+    background: core.colorsBackgroundLight[3],
+    borderColor: core.colorsTextIcon.lowOnLight,
```
jaketrent: matches figma.  seems odd semantic choice from the figma point of view.  why not colorsBorder?

packages/checkbox/src/css/index.js
```diff
@@ -62,26 +62,14 @@ export default {
     right: '-7px',
     bottom: '-7px'
   },
-  '.psds-checkbox__square:focus:before': {
```
jaketrent: what were these?  why can they all go away?

seems related to focus rings.  But in that case, why are there still so many lines of css for focus rings remaining?

packages/checkbox/src/css/index.js
danethurber: it's figma the square of the checkbox is considered an icon. i agree is is weird but these colors looks better than the border colors. synced up with brian and this is what we came up with

packages/checkbox/src/css/index.js
danethurber: this was missed when we did the focus ring to halo component refactor. there is no :before element to target

PR #799, opened 2020-01-22
refactor(switch): apply 2020 color palette

jaketrent:
util/transparentize removed -- just another case where it seems we could do w/o;  small size switch had the "thumb" changed from 12px to 10px.  New border on unchecked state. Orange color made blue in anticipation of later removal of orange option.  Waiting on that later removal change so that Brian can make a determination if we want to go from 2 options (green/blue) down to just one, essentially removing the color prop.

### DIFF
--- a/packages/switch/src/css/index.js
+++ b/packages/switch/src/css/index.js
@@ -3,7 +3,6 @@ import {
   defaultName as themeDefaultName,
   names as themeNames
 } from '@pluralsight/ps-design-system-theme'
-import { transparentize } from '@pluralsight/ps-design-system-util'
 
 import * as vars from '../vars/index.js'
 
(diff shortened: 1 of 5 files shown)


packages/switch/src/css/index.js
```diff
@@ -43,28 +46,31 @@ export default {
   [`.psds-switch__track.psds-switch__track--size-${vars.sizes.large}`]: {
     height: '24px',
     width: '48px',
-    borderRadius: '12px',
-    padding: '1px'
+    borderRadius: '12px'
+    // padding: '1px'
```
danethurber: stray comment here.

packages/switch/src/css/index.js
```diff
@@ -43,28 +46,31 @@ export default {
...
+    borderColor: 'transparent'
   },
   [`.psds-switch__track--checked.psds-switch__track--color-${vars.colors.green}`]: {
-    backgroundColor: core.colors.green
+    backgroundColor: core.colorsGreen.base,
+    borderColor: 'transparent'
   },
   [`.psds-switch__track--checked.psds-switch__track--color-${vars.colors.orange}`]: {
-    backgroundColor: core.colors.orange
+    // NOTE: temp: make orange blue until we do a breaking change to remove orange
```
danethurber: noted. there is also an orange story you could comment on - or just remove

jaketrent: Makes sense to me to keep the story, since it's built on the colors enum, until the colors enum changes.

PR #801, opened 2020-01-22
row: Remove Row.Action, use DS Button instead

jaketrent:
In response to a request from Will

https://pluralsight.slack.com/archives/C70HPSJPP/p1579719708014300o
    
His use case is wanting to use another icon button that is provided as official widget from another BC.

This solution was in consultation with @sweeting .


sweeting:
👍

PR #804, opened 2020-01-22
Feat/textarea colors 2020

danethurber:


packages/textarea/src/css/index.js
```diff
@@ -17,27 +24,27 @@ export default {
...
 
-    '&:focus': { outline: 'none' }
+    '&:focus': { outline: 'none' },
+    '&::placeholder': { color: colorsTextIcon.lowOnLight }
   },
   [`.psds-text-area__field.psds-theme--${themeNames.light}`]: {
     '&:focus': { border: '1px solid transparent' }
   },
 
+  [`.psds-text-area__field.psds-theme--${themeNames.light}`]: {
```
jaketrent: This selector is a dup and can be combined with the one above.

packages/textarea/src/css/index.js
```diff
@@ -17,27 +24,25 @@ export default {
...
   [`.psds-text-area__field.psds-theme--${themeNames.light}`]: {
+    border: `1px solid ${colorsBorder.highOnLight}`,
     '&:focus': { border: '1px solid transparent' }
   },
 
   [`.psds-text-area__field--error.psds-theme--${themeNames.light}`]: {
     border: '1px solid transparent'
   },
-  [`.psds-text-area__field.psds-theme--${themeNames.light}`]: {
-    background: colors.white,
```
jaketrent: @danethurber Did you still want background: white?  I'm not seeing this transferred in the refactor.

danethurber: uh oh....i'll fix

danethurber: looks like i added a white background to the field regardless of theme. so this change is correct

PR #805, opened 2020-01-22
feat(textinput): apply 2020 colors

danethurber:


packages/textinput/src/css/index.js
```diff
@@ -1,4 +1,12 @@
-import { colors, layout, type } from '@pluralsight/ps-design-system-core'
+import {
+  colorsBackgroundDark,
```
jaketrent: It's really good that you're doing this to allow treeshaking at least parts of core.  I don't think I've been doing this.  We should call this out for a good refactor on these color conversion PRs if aren't seeing them, I think.  cc @EdwardIrby

danethurber:
@jaketrent yes i noticed the alignment issue. i'll create an issue and we can follow up

PR #812, opened 2020-01-23
refactor(row): apply 2020 color palette

jaketrent:
asdf


packages/row/src/css/index.js
```diff
@@ -25,6 +31,9 @@ export default {
       borderTop: 'none'
     }
   },
+  [`.psds-row.psds-theme--${themeNames.light}`]: {
+    borderTop: `1px solid ${colorsBorder.lowOnLight}`
```
danethurber: this border is super hard to see

jaketrent: double checking...  seems to be set as that in figma.  it is subtle.

PR #813, opened 2020-01-23
feat(avatar): new colors applied

EdwardIrby:


danethurber:
https://github.com/pluralsight/design-system/blob/master/packages/avatar/src/vars/index.js#L9

some other colors that we should update. maybe just use the base of each non-neutral color

cc @sweeting

packages/avatar/src/vars/index.js
```diff
@@ -7,15 +16,13 @@ export const sizes = {
...
-  '#6D8C36',
-  '#AE8018',
-  '#82214E'
+  colorsBlue[7],
+  colorsTeal[7],
+  colorsGreen[7],
+  colorsOrange[7],
+  colorsRed[7],
+  colorsPink[7],
+  colorsPurple[7]
```
danethurber: nitpick. colorsBlue['base'] is more readable for me. but whatever 😉

EdwardIrby: Base is index 6 not 7 Brian has it at 7in slack convo

PR #816, opened 2020-01-23
refactor(circularprogress): apply 2020 color palette

jaketrent:
approach matching linear progress.

### DIFF
--- a/packages/circularprogress/src/css/index.js
+++ b/packages/circularprogress/src/css/index.js
@@ -1,9 +1,8 @@
-import * as core from '@pluralsight/ps-design-system-core'
+import { colorsTextIcon } from '@pluralsight/ps-design-system-core'
 import {
   defaultName as themeDefaultName,
   names as themeNames
 } from '@pluralsight/ps-design-system-theme'
-import { transparentize } from '@pluralsight/ps-design-system-util'
 
 import * as vars from '../vars/index.js'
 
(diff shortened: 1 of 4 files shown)


packages/circularprogress/src/css/index.js
```diff
@@ -38,20 +37,20 @@ export default {
...
+    stroke: colorsTextIcon.highOnDark
   },
   // __bg
   [`.psds-circularprogress__bg`]: {
     fill: 'transparent',
     strokeWidth: `${vars.style.strokeWidth}px`
   },
   [`.psds-circularprogress__bg.psds-theme--${themeNames.light}`]: {
-    stroke: transparentize(0.8, core.colors.gray04)
+    stroke: 'rgba(0, 0, 0, 0.15)'
```
EdwardIrby: This is the third time I've seen this sorta usage of this color for this. Maybe it should be a core token like track/progress on lo & high

jaketrent: Perhaps.  We have used it a few times.  In my mind, it doesn't rise to the generality level of text, icon, and border though.

EdwardIrby: Should we maybe create a tokens package?

jaketrent: Isn't that what core is?

PR #827, opened 2020-01-24
dialog: fix tail

jaketrent:
- add site doc for showing dialog in context
- fix doc link
- fix dialog to show tail (was regression from overflow support a while back).  Solution is to have an inner div to handle scrolling.  This solution now doesn't automatically move the close button the left when a scrollbar is present.  For this reason, I increased the close button margin generally.  I want to double check this, design-wise, with @sweeting .  There may be better alternatives.


sweeting:
I like the position of the close button better now. 👍

jaketrent:
@sweeting like it even better!  wow, sounds great.  I'll merge.

PR #829, opened 2020-01-24
Feat/colors 2020 badge

EdwardIrby:

PR #839, opened 2020-01-28
badge: Simplify the css selector utility naming

jaketrent:
An attempt to make it more readable.  See last commit only.  Merge after #829.

### DIFF
--- a/packages/badge/src/js/index.js
+++ b/packages/badge/src/js/index.js
@@ -1,21 +1,10 @@
-import { appearances } from '../vars/index.js'
-
-export function defaultWithColor(themeName, c) {
-  return appearance(appearances.default) + theme(themeName) + color(c)
-}
-
-export function subtleThemeWithColor(themeName, c) {
-  return appearance(appearances.subtle) + theme(themeName) + color(c)
-}
-
-function appearance(a) {
-  return `.psds-badge--appearance-${a}`
-}
-
-function color(c) {
-  return `.psds-badge--color-${c}`
-}
-
-function theme(t) {
-  return `.psds-theme--${t}`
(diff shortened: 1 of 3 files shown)


packages/badge/src/js/index.js
```diff
@@ -1,21 +1,10 @@
...
-  return `.psds-theme--${t}`
+export function select(themeName, appearance, color) {
+  return (
+    '.psds-badge--apearance-' +
+    appearance +
+    '.psds-badge--color-' +
+    color +
+    '.psds-theme--' +
+    themeName
+  )
```
danethurber: i know i've built something similar in other packages. i wonder if we can find a generic/reuseable solution.

also i want to show you how i've been doing css w/cascade in gitprime's repo and see if there is a way to make it work with our repo and css build

jaketrent: Where's the other thing you built?  I'd be happy to consider it after seeing what was generalizable.  Though I do think this is so simple as to not need something very fancy.  I almost removed this string builder for this very reason.

EdwardIrby: Think this is what he's referring to https://github.com/pluralsight/design-system/blob/master/packages/emptystate/src/react/index.js#L20

jaketrent: The puurpose here seems slightly different than what was discussed in the PR.

- PR is creating a string combining selectors.
- The code ref’ed here is combining selectors in a glamor call.

The point of wanting generality still stands.  The PR just doesn’t seem to overlap with the purpose of the code referenced.

When I said as much in the Slack convo, I got the thumbs up from Dane, so I think we're ok to merge.

PR #841, opened 2020-01-28
fix(actionmenu): incorrectly named prop 'css' should be 'style'

### LINKED ISSUE #771: ActionMenu: nested menus are improperly positioned
Screen Shot 2020-01-10 at 9 48 49 AM

EdwardIrby:
resolves #771

### DIFF
--- a/packages/actionmenu/src/react/item.js
+++ b/packages/actionmenu/src/react/item.js
@@ -103,7 +103,7 @@ export default function Item(props) {
     props.nested &&
     isActive &&
     React.cloneElement(props.nested, {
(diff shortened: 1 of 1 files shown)


jaketrent:
in a mtg, but I'll get back to a proper review.

Fyi, we have used css in the past because it's a valid glamor prop.  Not sure if that's the case here.

EdwardIrby:
> in a mtg, but I'll get back to a proper review.
> 
> Fyi, we have used css in the past because it's a valid glamor prop. Not sure if that's the case here.

It's an inline style in this case so it should be style as it's going directly on the react node.

Issue #836, opened 2020-01-27
form: ButtonRow and VerticalLayout allow 1 child element


jaketrent:
Currently proptypes warns of single element.  Should allow this.

Ref:
https://pluralsight.slack.com/archives/C70HPSJPP/p1580159944027300

EdwardIrby:
resolved by #844

PR #844, opened 2020-01-28
fix(form): relaxes proptypes

EdwardIrby:
In regards to https://pluralsight.slack.com/archives/C70HPSJPP/p1580159944027300


packages/form/src/react/button-row.js
```diff
@@ -24,7 +24,10 @@ const ButtonRow = (props, context) => (
 ButtonRow.displayName = 'ButtonRow'
 ButtonRow.propTypes = {
   align: PropTypes.oneOf(Object.keys(vars.aligns).map(k => vars.aligns[k])),
-  children: PropTypes.arrayOf(PropTypes.element)
+  children: PropTypes.oneOfType([
+    PropTypes.node,
+    PropTypes.arrayOf(PropTypes.node),
+  ]),
```
danethurber: you can just do `PropTypes.node`. it includes arrays

jaketrent:
Added a couple extra commits: to simplify the node example on buttonrow and to relax the form.verticallayout proptypes

Resolves: https://github.com/pluralsight/design-system/issues/836

packages/form/src/react/vertical-layout.js
```diff
@@ -22,7 +22,10 @@ const VerticalLayout = (props, context) => (
 )
 VerticalLayout.displayName = 'VerticalLayout'
 VerticalLayout.propTypes = {
-  children: PropTypes.arrayOf(PropTypes.element)
+  children: PropTypes.oneOfType([
+    PropTypes.element,
+    PropTypes.arrayOf(PropTypes.element)
+  ])
```
EdwardIrby: @jaketrent Why not PropTypes.node on this? is it because of the inline style?

jaketrent: Yes, because we operate on elements inside the component, and node is too broad to be acceptable internally.

jaketrent:
Looks like we're good to merge.  Thanks for the review, Dane.

PR #845, opened 2020-01-28
Fix/text input icon alignment

### LINKED ISSUE #810: textinput: after field icon missaligned
the "after field w/ icon" story looks like the icon is aligned too far right

73007417-c0c3ce00-3dc9-11ea-920a-7ce7cd320bbd

EdwardIrby:
resolves #810


packages/icon/src/css/index.js
```diff
@@ -25,7 +25,9 @@ export default {
     width: vars.widths.large
   },
   [`.psds-icon > svg`]: {
-    fill: 'currentColor'
+    fill: 'currentColor',
+    width: '100%',
+    height: '100%'
```
EdwardIrby: This is optional but I noticed icon wasn't aligning correctly in it's container. We can do this style or we can do the following

```
[`.psds-icon`]: {
    display: 'inline-flex',
   alignItems: 'center',
   justifyContent: 'center,
 }
```

danethurber:
@EdwardIrby i spot checked a handful of components and they seemed to be ok

packages/textinput/src/css/index.js
```diff
@@ -44,7 +44,7 @@ export default {
     padding: `6px ${layout.spacingXSmall}`
   },
   '.psds-text-input__field--w-after': {
-    paddingRight: 0
+    paddingRight: layout.spacingXSmall
```
jaketrent: Kinda odd.  I wonder why we would have set this explicitly as 0.  Did that value work at some previous point?

packages/icon/src/css/index.js
```diff
@@ -10,7 +10,14 @@ import * as vars from '../vars/index.js'
 
 export default {
   [`.psds-icon`]: {
-    display: 'inline-block'
+    display: 'inline-flex',
```
jaketrent: Not likely, but since Icon can take a child, is this layout change potentially breaking for any of those custom child cases?  (the ones that we don't have in our storybook to check)

EdwardIrby: Not that I can think of given Icon takes default size of medium

Issue #834, opened 2020-01-27
core: make remaining old colors match the new conventions.


jaketrent:
What about white and black colors?  Are there any others?

We don't want two patterns:

1) import { colors } from 'core' // as seen in legacy

2) import { colorsPink } from 'core' . // as established in 2020 vars

Instead, I propose we push white into the 2nd pattern:

import { colorsWhite } from 'core'

@sweeting To help know what colors will be affected like this ^, which colors will remain from the legacy set in the 2020 set?

jaketrent:
From convo with Brian:

- Move the code to colorsCode
- Remove the gradientHorz and Vert, they’ve been replaced
- Move colorsBlack and colorsWhite over
- Remove the rest later (tent. 31 Mar)

PR #846, opened 2020-01-28
feat(core): add white, black, and code colors

jaketrent:
Resolves: #834

Someday the old tokens will be removed.  This moves the ones we're keeping from the old set to the new pattern.

### DIFF
--- a/packages/core/js/colors.2020.js
+++ b/packages/core/js/colors.2020.js
@@ -1,3 +1,7 @@
+export const colorsWhite = '#FFFFFF'
+
+export const colorsBlack = '#000000'
+
 export const colorsBlue = {
   base: '#0084BD',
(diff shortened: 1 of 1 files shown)


packages/core/js/colors.2020.js
```diff
@@ -1,3 +1,7 @@
+export const colorsWhite = '#FFFFFF'
+
+export const colorsBlack = '#000000'
```
EdwardIrby: Pedantic but I think we should make this an object so it mirrors other color objects maybe???

jaketrent: I checked with Brian.  These are the facts that made me keep it this one:

- It's a single value color, no spectrum like the others.
- Nobody says things like "white base"

PR #851, opened 2020-01-30
refactor(examples): update example to recent design system

culli:
This updates the example create-react-app to use more recent design system versions, and updates the icon usage to match.  I was looking for a starting place to try replicate some strange behavior I see in another app, and figured others could use the same updates.


jaketrent:
@culli Awesome, thanks for seeing this and making the update.  I can take care of the conflicts to get this merged.  Thank you!

jaketrent:
The failing lint here is weird.  The problem is apparently that lint doesn't like the commit messages for conventional commit

Since this is a single commit branch, in the example dir only, let's just merge.  

Someone wanna double check this logic and what I'm seeing? @danethurber / @EdwardIrby ?

danethurber:
@looks good to me

PR #859, opened 2020-01-31
ci: integrate commitizen with commitlint

danethurber:
use the commitlint config as the template for commitzen. try it and tell me what you think - it's different..not sure i like it

why: 
- `chore` isn't a valid scope even though commitizen tells you it is
- tab completion for package names


danethurber:
closing. i don't like the way the prompt works 😢 and nobody else has a strong opinion

PR #860, opened 2020-01-31
Site/featflags

jaketrent:
asdf


packages/site/src/ui/example/react.js
```diff
@@ -260,6 +273,7 @@ ReactExample.propTypes = {
 }
 
 ReactExample.defaultProps = {
+  featureFlags: [],
```
danethurber: shouldn't this be an `{}`

jaketrent:
Thanks for the help cleaning up lint, Dane1

PR #868, opened 2020-02-04
fix(dropdown): corrects placeholder fonter color

### LINKED ISSUE #673: Correct placeholder text color in dropdown
In the dropdown component, the placeholder text should have less contrast than a selected value.

In the dark theme, the placeholder text color is correct, but when a value is selected, the selected value text color is the same as the placeholder.

In the light theme, the placeholder text color is the same dark color as the selected text value.

EdwardIrby:
resolves #673


EdwardIrby:
@danethurber can you rerun the ci build?

PR #870, opened 2020-02-05
fix(viewtoggle): correct inactive option font color

### LINKED ISSUE #866: viewtoggle: Unselected text color incorrect on theme toggle in docs
The theme toggle unselected text color is incorrect. Should be `textIconOnDarkLow`.

image

EdwardIrby:
resolves #866


EdwardIrby:
> The change is good and does address the related bug. But only in terms of the view toggle by itself. There's something still wrong with the context shown in the screenshot in the bug.
> 
> Go to http://localhost:1337/components/badge#colors-&-appearance
> 
> And see the viewtoggle on dark theme in the react example widget. I'm still seeing:
> 
> Screen Shot 2020-02-05 at 9 46 50 AM
> 
> (Changes requested are just to complete this other part of the bug resolution.)

fixed!

packages/site/src/ui/example/react.js
```diff
@@ -234,10 +234,12 @@ class ReactExample extends React.Component {
         {this.renderError()}
 
         {this.props.themeToggle && (
-          <ThemeToggle
-            activeThemeName={this.state.themeName}
-            onSelect={this.handleThemeSelect}
-          />
+          <Theme name={this.state.themeName}>
```
jaketrent: The solution doesn't quite make sense to me.

`state.themeName` is the internal state for the theme toggling.  But it should change the *look* of the theme toggler.  Right?

jaketrent: I talked through the solution with Edward, and this solution now makes sense.  ViewToggle didn't used to consume Theme.  Now it does.  Through Context.  Page-level context is light theme.  That could be the reason that the inactive option is too dark.  

Edward is double-checking that all commits are in and works on his box.

PR #871, opened 2020-02-05
refactor(site): dynamically add page export map

danethurber:


packages/site/package.json
```diff
@@ -17,6 +17,7 @@
     "core-js": "^3.3.6",
     "debounce": "^1.2.0",
     "dotenv": "^8.2.0",
+    "glob": "^7.1.6",
```
EdwardIrby: Use fast-glob it's .... wait for it ... faster  🏎 , LOL

JK

PR #872, opened 2020-02-05
[experimental] Feat/mdx

danethurber:
this is just an example of what is involved. there would be some refactoring to make this a reality. look but **do not merge**

### What You're Solving

- simplify setup by adding chrome and styling in _app.js
- allow pages to be written in MDX format
- configure mdx provider with custom components

### How to Verify

- install
- start
- look at root page(all other pages are broken :wink:)

### DIFF
--- a/packages/site/pages/_app.js
+++ b/packages/site/pages/_app.js
@@ -1,0 +1,53 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import { MDXProvider } from '@mdx-js/react'
+
+import { Heading } from '@pluralsight/ps-design-system-text'
+
+import {
+  Code,
+  Chrome,
+  Content,
+  P,
+  PageHeading,
+  TextLink
+} from '../src/ui/index.js'
+
+const mdComponents = {
+  a: TextLink,
+  code: Code,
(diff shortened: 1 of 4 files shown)


packages/site/pages/_app.js
```diff
@@ -0,0 +1,53 @@
...
+import {
+  Code,
+  Chrome,
+  Content,
+  P,
+  PageHeading,
+  TextLink
+} from '../src/ui/index.js'
+
+const mdComponents = {
```
jaketrent: oh, this is awesome.

packages/site/pages/install.mdx
```diff
@@ -1,5 +1,3 @@
-import React from 'react'
-
 import Badge from '@pluralsight/ps-design-system-badge'
```
jaketrent: Looks great.  Makes on-off pages like this nice.  I'd love to see what a conversion might look like (if desirable at all) of one of the component doc pages.

EdwardIrby: Can we do partial/includes with this to make component pages easier? Looks simpler overall which is better.

PR #875, opened 2020-02-06
refactor site with new colors and conventions

danethurber:
- use new colors tokens
- wrap in feature flags context with new colors
- convert class to functional components
- add displayNames
- organize files with default export as first definition
- organize styled-jsx
- use named imports for core variables

NOTE: did not change some of the more complicated components (Content,
Code, etc) because i have some future plans for them. did not refactor page components because adopting mdx would be a better direction


packages/site/src/ui/page-heading.js
```diff
@@ -53,3 +31,29 @@ PageHeading.propTypes = {
...
   packageName: PropTypes.string
 }
+
+const BadgeSpacer = () => <span style={{ width: layout.spacingLarge }} />
+
+function Beta(props) {
+  return (
+    <>
+      <BadgeSpacer />
+      <Badge color={Badge.colors.blue}>Beta</Badge>
```
danethurber: @sweeting we missed one

danethurber:
@jaketrent sorry this pr is so big. i was just in the groove and didn't think to break it up 😢

packages/site/pages/_app.js
```diff
@@ -0,0 +1,17 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+
+import FeatureFlags from '@pluralsight/ps-design-system-featureflags'
+
+const App = ({ Component, pageProps }) => (
+  <FeatureFlags flags={{ psds2020Colors: true }}>
```
jaketrent: This makes it much more "apply new colors to everything", which is nice if we're optimizing for blank flag changing.  When I put it in the examples code last, I was thinking in terms of putting a flag toggle on those particular examples that would respond to it.  But, I suppose this doesn't preclude that someday, as long as they nest.

packages/site/src/ui/change-log.js
```diff
@@ -34,61 +83,17 @@ const ChangeLog = props => {
...
-ChangeLog.propTypes = {
-  packageName: PropTypes.string,
+
+ChangeLogLabel.propTypes = {
+  packageName: PropTypes.string.isRequired,
   version: PropTypes.string
 }
 
-const getPackagesApiHost = _ => {
+const getPackagesApiHost = () => {
```
jaketrent: no likey the `_`? :)

packages/site/src/ui/doc.js
```diff
@@ -1,9 +1,10 @@
-import * as core from '@pluralsight/ps-design-system-core'
 import Markdown from 'react-markdown/with-html.js'
 import PropTypes from 'prop-types'
 import React from 'react'
 import Router from 'next/router.js'
 
+import * as core from '@pluralsight/ps-design-system-core'
```
jaketrent: this used?

packages/site/src/ui/link.js
```diff
@@ -1,8 +1,9 @@
-import DSLink from '@pluralsight/ps-design-system-link'
 import NextLink from 'next/link.js'
 import PropTypes from 'prop-types'
 import React from 'react'
 
+import DSLink from '@pluralsight/ps-design-system-link'
```
jaketrent: Named imports all belong in one block in my personal opinion, whether we wrote them or not.  I believe the project started and is still mostly that way. We could talk as a team to adjust this.

packages/site/src/ui/link.js
```diff
@@ -27,7 +29,10 @@ const Link = React.forwardRef((props, ref) => {
     </NextLink>
   )
 })
+
 Link.appearances = DSLink.appearances
+Link.displayName = 'Link'
```
jaketrent: Nice catch.  How in the world did you discover that amongst everything else?

packages/site/src/ui/page-heading.js
jaketrent: No comprendo.  Missed one what?

packages/site/src/ui/prop-types.js
```diff
@@ -1,57 +1,114 @@
...
 
-const row = ([name, value, required, defaultValue, description]) => ({
-  name,
-  value,
-  required,
-  defaultValue,
-  description
-})
+        <Table className="table">
+          {Array.isArray(propsTable)
```
jaketrent: Didn't want to add "make proptypes api not terrible" too? :)

packages/site/src/ui/change-log.js
danethurber: i think it's a linting rule/warning? i know it's a linting rule in typescript so maybe i'm just in the habit

danethurber: i think it's a linting warning because it's an unused var. can't remember

packages/site/src/ui/doc.js
danethurber: yes. but I didn't end up changing this file so i shouldn't have commited this file

packages/site/src/ui/link.js
danethurber: there are linting warnings around components missing display names. i'd like to eventually enforce that rule once everything is in compliance. while pairing with teams i've noticed that it's super hard to use the react dev tools because a ton of components are just named `Unknown`

packages/site/src/ui/page-heading.js
danethurber: color change. i'm guess it should be colorsBlue.base or something

packages/site/src/ui/prop-types.js
danethurber: i get too neurotic with organizing and formatting things. it's a legit problem sometimes 😬

packages/site/src/ui/link.js
danethurber: yeah we should talk. I know we talked about it in the past and maybe we didn't lock down a final decision. i've generally been organizing in this fashion but I might have been doing that out of turn

packages/site/src/ui/page-heading.js
jaketrent: Still `blue`, see: https://design-system.pluralsight.com/components/badge#colors-&-appearance

packages/site/src/ui/link.js
jaketrent: Definitely not out of turn in terms of it not being allowed or sanctioned or whatever.  I think our styleguide is too ambiguous for that.  2 weeks in the brig should be sufficient. :)

PR #879, opened 2020-02-06
fix(icon): restore coloring of svg

jaketrent:
Regression was in 8ae52b86ed9cea90c8f204f916c20e5882115e11

I discovered an unexpected limitation/gotcha for how glamor works.  If you have a selector:

'asdf': {
  '& > svg': { attr: 'firstDeclaration' }
}

And later:

'asdf--modifier': {
  '& > svg': { attr: 'secondDeclaration' }
}

And then put them together like we do:

glamor.css({
  ...css['asdf']
  ...css['asdf--modifier']
})

Then the first `& > svg` is overwritten by the 2nd, giving attr only "secondDeclaration".

PR #881, opened 2020-02-06
fix(icon): restore svg coloring

jaketrent:
instead of #879 if you like.

### DIFF
--- a/packages/icon/src/react/index.js
+++ b/packages/icon/src/react/index.js
@@ -9,11 +9,11 @@ import * as vars from '../vars/index.js'
 
 const style = {
   icon: ({ color, size }) =>
-    css({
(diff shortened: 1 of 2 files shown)


packages/icon/src/react/index.js
```diff
@@ -9,11 +9,11 @@ import * as vars from '../vars/index.js'
 
 const style = {
   icon: ({ color, size }) =>
-    css({
-      ...stylesheet['.psds-icon'],
-      ...stylesheet[`.psds-icon--size-${size}`],
-      ...stylesheet[`.psds-icon--color-${color}`]
-    })
+    css(
```
danethurber: did you test this out. i tested out using compose but this might work as well 🤷‍♂

PR #883, opened 2020-02-07
[experimental] new CodeBlock component

danethurber:
playing around with using Prism highlighting and ReactLive. **do not merge**

I've heard that the prism highlighter does a better job with jsx. ReactLive uses prism under the hood so i through that in for fun

### DIFF
--- a/packages/site/pages/components/avatar.js
+++ b/packages/site/pages/components/avatar.js
@@ -3,14 +3,14 @@ import React from 'react'
 
 import {
   Chrome,
-  Code,
   Content,
-  Example,
   P,
   PageHeading,
   PropTypes,
   SectionHeading
 } from '../../src/ui/index.js'
+
+import CodeBlock from '../../src/ui/code-block/index.js'
 
 export default _ => (
   <Chrome>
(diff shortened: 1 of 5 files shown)


packages/site/pages/components/avatar.js
```diff
@@ -57,28 +57,45 @@ export default _ => (
...
       <P>Five avatar sizes are available.</P>
-      <Example.React
-        includes={{ Avatar }}
-        codes={Object.keys(Avatar.sizes).map(
-          size =>
-            `<Avatar size={Avatar.sizes.${size}} src="https://secure.gravatar.com/avatar/0f792a763ebf08411c7f566079e4adc7?s=400" />`
-        )}
-      />
+
+      <CodeBlock live>
```
danethurber: this would look even better with mdx 😉

packages/site/src/ui/code-block/scope.js
```diff
@@ -0,0 +1,8 @@
+import Avatar from '@pluralsight/ps-design-system-avatar'
```
danethurber: we can import and scope all the ds components here

Issue #888, opened 2020-02-10
link: hover styles don't work


jaketrent:
### Expected behavior

Use CSS selectors, hover, see style

### Actual behavior

Use CSS, hover, see that style is overwritten in dev tools.

https://codesandbox.io/s/dreamy-noether-mnezl

This is to help NaN.

This is not an issue in React land.

Has to do with ordering in stylesheet exported.

PR #890, opened 2020-02-10
fix(link): reorder selectors for hover in cascade

jaketrent:
Resolves: #888


packages/link/package.json
```diff
@@ -15,7 +15,7 @@
     "build:js": "cross-env ESM=true babel --root-mode upward src --out-dir dist/esm",
     "build:watch": "npm run build:js -- --watch",
     "prepublish": "npm run build",
-    "storybook": "start-storybook -p 6006",
+    "storybook": "start-storybook -p 6006 -s ./dist",
```
danethurber: are these storybook changes supposed to be committed. i generally don't run build along with storybook so this would be confusing when making changes

jaketrent: Yes, that was my intent.

I always build and thought one had to build.  But I guess things have changed since that habit was formed.  

I don't think there's another way to have this kind of test (testing output stylesheets) (which arguably we should not have), but I am seeing a bit of value with it.  In this case, the test case also involves interaction of hover, which is unsnapshottable, so it's less helpful.  But it at least provides an interactive environment to test and find the bug.

EdwardIrby: Jake's right about this one. Without building glamour run on src doesn't output the same and hover appears to work. We might want to refactors our other stories.

jaketrent: I think the other stories can remain as-is for now because they're all using React components in the output.

This is the first story that I'm aware of where we're using the CSS stylesheet that's built inside of storybook.

EdwardIrby: I'm not sure the storybook thing is it. I just ran this branch using a build off master and noticed it was working on my machines. I haven't yet run a build with your css mod.

jaketrent: Edward, not sure what you mean, but if you want to see the problem in action, try:

https://codesandbox.io/s/dreamy-noether-mnezl

And then to verify the fix,:

the PR should look good after a build and npm run storybook

EdwardIrby: Yea the codepen makes sense to me. 

I mean I wanted to verify the storybook change was necessary and so I first ran build of the entire DS on the master branch. Then I checked out your branch and noticed that when I run your new storybook command I'm not seeing the bug that's in the codpen link.

packages/link/src/css/index.js
```diff
@@ -37,5 +25,17 @@ export default {
...
+    '&:hover': {
+      color: colors.orange,
+      textDecoration: 'underline'
+    }
+  },
+  [`.psds-link.psds-button--2020-colors`]: {
+    '&:hover': {
+      color: colorsPrimaryAction.background,
+      textDecoration: 'underline'
+    }
```
EdwardIrby: This looks cool to me I just don't think the storybook change makes a difference in detecting the error. I tested just the storybook mod and noticed it was still working in storybook.

jaketrent: Without the mod to storybook, how are the selectors like `.psds-link` going to be usable in the story?

(The mod makes those possible, serving static files from `/dist` in the script command arg, then adding the stylesheet link to `/index.css` in the `preview-head.html`.)

EdwardIrby: When running `"storybook": "start-storybook -p 6006 -s ./dist",` using the old build from master pictured below hover still works as expected. This is without using your mod to the *css/index.js*. I think that modifying storybook will not reveal this bug at all while the codepen does. 

Screen Shot 2020-02-11 at 2 13 43 PM

PR #902, opened 2020-02-14
feat(button): removes featureFlags consumer and old styles

EdwardIrby:


packages/button/src/css/index.js
```diff
@@ -49,38 +46,23 @@ export default {
...
-        position: 'absolute',
-        top: '-4px',
-        left: '-4px',
-        right: '-4px',
-        bottom: '-4px',
-        border: `3px solid ${colors.blue}`,
-        borderRadius: '4px'
-      }
+      outline: 'none'
     }
```
danethurber: we lost our focus state

EdwardIrby: @sweeting what's that blue suppose to be?

PR #904, opened 2020-02-14
feat(link): removes featureFlags consumer and old styles

EdwardIrby:


EdwardIrby:
> the subtle on light looks wrong. can't remember what it's supposed to be. double check please

It's right the story doesn't change the default color when you switch to light theme. Subtle inherits it's font color.

PR #907, opened 2020-02-19
refactor(button): update button colors and interactions for states & themes

sweeting:
- New focus state across buttons (using box-shadow)
- Give all buttons a little push transition down on the active state
- Add hover, active, focus states for Primary button specific to light theme
- Make secondary button hover/active state more evident on light theme
- Make secondary button states for hover/active/focus consistent across themes
- Update styling of stroke button for better contrast on light and dark themes

### DIFF
--- a/packages/button/src/css/index.js
+++ b/packages/button/src/css/index.js
@@ -90,15 +90,15 @@ export default {
     color: colorsTextIcon.highOnDark,
     background: colorsPrimaryAction.background,
     '&:not([disabled]):hover': {
-      background: colorsBlue[5],
+      background: colorsBlue[7],
       cursor: 'pointer'
     },
     '&:not([disabled]):focus': {
-      background: colorsBlue[5],
+      background: colorsBlue[7],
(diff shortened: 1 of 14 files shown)


packages/button/src/css/index.js
```diff
@@ -95,53 +86,126 @@ export default {
...
+      background: colorsBlue[7],
+      cursor: 'pointer'
+    },
+    '&:not([disabled]):focus': {
+      background: colorsBlue[7],
+      cursor: 'pointer'
+    },
+    '&:not([disabled]):active': {
+      background: colorsBlue[8],
+    }
```
danethurber: should these colors be part of the primary action palette

sweeting: All states have AA or better. The only one that didn't was the primary hover state on dark.

@danethurber I did as much as I could to try to adjust the styling of the buttons inside banners to improve contrast, but was unable to get it to build so if you want to take a look and maybe we could jump on a call to review it.

PR #913, opened 2020-02-25
Portal actionmenu overlay to direct child of body

twclark0:
## Instructions

- Action menu (specifically the overlay)

### What You're Solving

This is in response to: https://pluralsight.slack.com/archives/C70HPSJPP/p1582318454236600

As a side note, I wasn't able to get storybook running so I didn't actually verify that it works correctly. Can someone try it out on their machine and see if it puts the overlay just under the body?

### DIFF
--- a/packages/actionmenu/src/react/overlay.js
+++ b/packages/actionmenu/src/react/overlay.js
@@ -11,6 +11,13 @@ const styles = {
   overlay: () => css(stylesheet['.psds-actionmenu__overlay'])
 }
 
+const hasComponentRendered = () =>
+  !!(
+    typeof window !== 'undefined' &&
+    typeof window.document !== 'undefined' &&
+    typeof window.document.body !== 'undefined'
+  )
+
 const Overlay = props => {
   function handleOverlayClick(evt) {
     if (evt.target.id === MODAL_OVERLAY_ID) {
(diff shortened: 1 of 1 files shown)


packages/actionmenu/src/react/overlay.js
```diff
@@ -11,21 +11,31 @@ const styles = {
   overlay: () => css(stylesheet['.psds-actionmenu__overlay'])
 }
 
+const hasComponentRendered = () =>
```
jaketrent: Will this break SSR?  window is not defined outside the browser.

twclark0: Right, so ideally line 16 catches that?

twclark0: This is the same format done in the position package

danethurber: https://github.com/JedWatson/exenv is the react core way to do this check

twclark0: I have never seen this. It's a package last updated almost three years ago. 

You want me to update the window check with this package?

danethurber:
merged

twclark0:
Hey @danethurber what does versioning and publishing this package look like? So we know when we can update our package.

danethurber:
@twclark0 sorry i missed you message. it's published. notes here

https://pluralsight.slack.com/archives/C70HQMTUM/p1582752534000100

PR #926, opened 2020-03-13
feat(tab): draft for adding in .d.ts files for packages

jhewlett:
I briefly discussed with Dane the possibility of adding Typescript `.d.ts` files for components. It sounds like there are plans to do this at some point. I wanted to play around with a few things and see what you think.

This adds types for the `Tab` component as a proof of concept.

To figure out what all the allowed props were, I used a combination of the prop types and the documentation. I also also looked at some usages in the wild, for example to discover that `css` and `style` seem to be props that are accepted on most components.

However, since most provided props are forwarded on to the DOM elements (if they meet the whitelist in `filter-react-props`), it would be quite challenging to capture all additional props that someone may want to add.

For this reason, I think we would need to document how to add your own props as a user. This can be done via a feature called interface augmentation, like this:

```typescript
declare module '@pluralsight/ps-design-system-tab' {
    interface ListItemProps {
        name: string                          //add "name" prop to ListItem
    }
}
```

We would probably want to adopt a naming convention like `ComponentNameProps` since the user would need to know the name of the props in order to augment them.

The "spec" is just a tsx file that will fail to compile if a prop is missing from the .d.ts or used incorrectly. You would run `npm run typecheck-dts` to typecheck.

There is a risk that this gets out of sync as new props are added or changed.

Longer term, I think the best option would be for the stories to be written in TypeScript
and have that act as the check for the .d.ts as well, since it already tries to highlight
the different states of the components.


jhewlett:
Pushed another commit that does basic type checking of the Tab components right in the existing Tab story.

This makes use of '--allowJs' and '--checkJs' compiler flags, and can catch basic things like extra properties, or missing required properties.

If we convert the story to TS we'll get even more checks (like making sure handler function for the onclick has the right shape). I played around with doing this, and had it converted and everything typechecking with `--strict` in about 15 minutes, but I couldn't figure out to get storybook to work with TS.

I think this is a better approach than separate "tests" just for typechecking, especially since we really need to run it to make sure the `.d.ts` is in sync with the actual code for the component.

packages/tab/src/index.d.ts
```diff
@@ -0,0 +1,44 @@
+import { FunctionComponent, ReactChild, Component, MouseEvent } from 'react'
+
+interface Styling {
```
jaketrent: Yeah, I think this is a good idea to separate.  Eventually, perhaps even more general, because these are meant to be applicable all UI components.

packages/tab/src/index.d.ts
```diff
@@ -0,0 +1,44 @@
...
+    css?: Record<string, string | number>
+    style?: Record<string, string | number>
+}
+
+interface ListProps {
+    className?: string
```
jaketrent: className belongs in Styling interface, imo

packages/tab/src/react/__stories__/index.story.js
```diff
@@ -1,8 +1,8 @@
-import PropTypes from 'prop-types'
-import React from 'react'
+import * as PropTypes from 'prop-types'
```
jaketrent: Does this change avoid having to set es module compatibility with TS?  If so, me likey.

packages/tab/src/index.d.ts
```diff
@@ -0,0 +1,44 @@
...
+type ListItem = FunctionComponent<ListItemProps & (ListItemAnchorProps | ListItemButtonProps) & Styling>
+
+interface PanelProps {
+    className?: string
+    labelledBy: string | number
+}
+
+type Panel = FunctionComponent<PanelProps & Styling>
+
+declare const exports: {
```
jaketrent: This is me not understanding TS.  Why `declare` here (as well as export later)?

packages/tab/src/react/__stories__/index.story.js
jhewlett: Maybe? I just know that TS seems to be stricter in ES standards and it wanted it this way when running through `tsc`. Storybook still worked OK as well.

packages/tab/src/index.d.ts
jhewlett: `declare` here is a way of declaring the type for an object that is created on the JS side.

Then the `export default` is saying that that object is the default export in the JS.

More here:
https://stackoverflow.com/a/41749803
https://www.typescriptlang.org/docs/handbook/declaration-files/by-example.html

jaketrent:
Hey @jhewlett , my votes:

- one devDep of TS in root
- convert stories (usually just index.story.js) to .tsx files per package as .d.ts are added.
- with stories in .tsx, we're already using storyshots to run the stories as snapshot tests, so our CI integration is probably close to complete (hopefully!)

jhewlett:
That all sounds good! I'll make these changes (probably next week) and work next on getting storybook to play nice with TSX files.

jhewlett:
I converted the Tabs story to `.jsx` and having it working in Storybook. Looks like the builds are failing because Jest isn't configured to handle the Typescript yet.

Reading here (https://jestjs.io/blog/2019/01/25/jest-24-refreshing-polished-typescript-friendly#typescript-support), it looks like our options are to use `@babel/preset-typescript`, which strips the annotations, but then we would need to run `tsc` separately to actually do type checking.

Alternatively, we could use `ts-jest` which would transpile and also do type checking. I can play around with both, but I'm curious if anyone has a preference here

Once I get those tests passing, the main thing left would be to ensure that anyone who installs the package gets the typings. I've tried consuming it locally using `npm link` and in its current state the types are not being discovered. I'll look into this some more.

schicks:
The search team uses `@babel/preset-typescript`, for two reasons. We always do separate type checking anyway, as part of our development/testing process, and we need the babel preset anyway when we are running for production mode through webpack. This may be a non-goal for the design system, but it seems to me like going the route of babel would allow for directly developing in typescript later, and does not actually rule out using `ts-jest`.

jaketrent:
@schicks , thanks for weighing in.

@jhewlett , I don't have a strong preference, partly because I don't fully appreciate the problem yet.  If we need to upgrade jest to get support, we can look into that.  But that might require a wider change.  If you can find a solution that allows us to do a per-package conversion/feature-add, that would be preferable.  

We don't run webpack for builds, but we do run babel in DS builds.

jhewlett:
@jaketrent Right now I can't even get Jest tests to run locally (even on master, on a clean clone). 

I did:

```
nvm use
npm i
npm run bootstrap
npm test
```

I get Hooks errors, like:

```
Storyshots › style overrides › panel disabled opacity (style)

    Invalid hook call. Hooks can only be called inside of the body of a function component. This could happen for one of the following reasons:
    1. You might have mismatching versions of React and the renderer (such as React DOM)
    2. You might be breaking the Rules of Hooks
    3. You might have more than one copy of React in the same app
    See https://fb.me/react-invalid-hook-call for tips about how to debug and fix this problem.

      33 |
      34 | function List(props) {
    > 35 |   const themeName = useTheme()
         |                     ^
      36 |   const allProps = {
      37 |     ...props,
      38 |     themeName
```

Have you seen this before?

jhewlett:
I was able to get the tests passing (in TravisCI, still doesn't work locally) by adding in `babel/preset-typescript`. (So we're not yet doing type checking.)

This was the least intrusive way I could find. AFAIK this only changes how Jest treats `.ts`/`.tsx` files.

jhewlett:
Just pushed a commit that actually ships the `.d.ts` files so they can be picked up by consumers. I verified that the types get picked up by using `npm link` in a test consumer project.

I'll explore how we might want to type check this as part of a build, but otherwise I think this is ready for review.

packages/tab/package.json
```diff
@@ -10,15 +10,16 @@
   "module": "dist/esm/index.js",
   "scripts": {
     "build": "run-s build:js build:cjs build:css",
-    "build:cjs": "babel --root-mode upward src --out-dir dist/cjs",
+    "build:cjs": "babel --root-mode upward src --out-dir dist/cjs && cp src/react/index.d.ts dist/cjs/react/index.d.ts && cp src/index.d.ts dist/cjs/index.d.ts",
```
jhewlett: I don't like having to do this, not sure if there's a better way.

packages/tab/src/index.d.ts
```diff
@@ -0,0 +1 @@
+export { default } from './react/index.js'
```
jhewlett: This is necessary to make TypeScript aware of the re-export and so it can pick up the types.

package.json
```diff
@@ -95,6 +97,7 @@
     "eslint-plugin-react": "^7.16.0",
     "eslint-plugin-react-hooks": "^2.2.0",
     "eslint-plugin-standard": "^4.0.1",
+    "fork-ts-checker-webpack-plugin": "^4.1.0",
```
jaketrent: learned: for @storybook/preset-typescript

package.json
```diff
@@ -115,6 +118,8 @@
     "raf": "^3.4.1",
     "react-test-renderer": "^16.11.0",
     "require-context.macro": "^1.2.2",
+    "ts-loader": "^6.2.1",
```
jaketrent: learned: for @storybook/preset-typescript

packages/tab/package.json
jaketrent: Yeah, this is surprisingly manual.

How do other projects include their typings?

What if we allowed src/**/*.d.ts via npmignore?

packages/tab/package.json
```diff
@@ -10,15 +10,16 @@
...
     "build:css": "build-css --useGlamor -i dist/cjs/css/index.js",
-    "build:js": "cross-env ESM=true babel --root-mode upward src --out-dir dist/esm",
+    "build:js": "cross-env ESM=true babel --root-mode upward src --out-dir dist/esm && cp src/react/index.d.ts dist/esm/react/index.d.ts && cp src/index.d.ts dist/esm/index.d.ts",
```
jaketrent: Is it important to have typings in `esm` alongside the `cjs` dir? 

I would assume the `"types"` property in the package.json (below) controls where tooling looks for typings.

packages/tab/src/react/__stories__/index.story.tsx
```diff
@@ -1,14 +1,13 @@
-import PropTypes from 'prop-types'
-import React from 'react'
+import * as React from 'react'
 import { storiesOf } from '@storybook/react'
 
 import Tab from '../index.js'
 
-function randomIntBetween(min, max) {
+function randomIntBetween(min: number, max: number) {
```
jaketrent: gettin fancy!

packages/tab/package.json
jhewlett: I don't think this is a common use case. The recommendation is to only include typings directly in the library if the library is written in TS[1]. (In the case that it's written in TS, you would have `tsc` emit the `.d.ts` files for you in the right place.) It seems like the main reason for this is that if the types are inacurrate/incomplete, you can iterate on the types independently without needing to cut a release of the library.

1: https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html

packages/tab/package.json
jhewlett: I'm not sure how to test with `esm`, or if it even makes sense to use them with TS. (What's the main benefit? Tree-shaking?)

We should probably just take them out of the `esm` folder until we know of a concrete use case for them there.

packages/tab/src/react/__stories__/index.story.tsx
jhewlett: I set `strict` in the `tsconfig`, so no implicit `any`s!

packages/tab/package.json
jhewlett: Allowing `src/**/*.d.ts` in npmignore could be an option if we can rely on the `types` property to point to the right place.

packages/tab/package.json
jhewlett: Removed

jaketrent: Hey Justin, just getting back in...

Yes, we have ESM as a format for tree shaking, mostly.  In the future, in-browser module loading is also a potential.

So, I suppose I'm a little confused and also probably didn't clearly articulate my question.  I'll try again.  Let me know how I do:

If the .d.ts file goes to npm *somewhere*, I would think that *where* would be pointed to by the `typings` package.json property -- and that alone would be sufficient for tooling/TS to pickup.  This is regardless of whether one is consuming the cjs or esm format of the DS code.  

Do you think that's ^ the case?

To your other question, yes, I *think* projects that use the ESM format would likely enjoy the TS typings as well.  TS looks more like ESM than CJS -- I would have almost thought that TS would use the ESM format, pointed to by the `module` property in package.json.

jhewlett: Got it, that makes sense. I was able to verify that your suggestion works for `cjs`. I'm not sure how to test with `esm`. Pushed up another commit with this.

jhewlett: For me at least, TypeScript is resolving `import Tab from '@pluralsight/ps-design-system-tab'` to `/dist/cjs`, so it seems to be following `main`.

jhewlett: The only way I can think of to get the `esm` is to do this:

`import Tab from '@pluralsight/ps-design-system-tab/dist/esm'`

If I do this, then it doesn't find typings if they're coming from `src`. If I place the `.d.ts` files alongside the source in `dist/esm`, then it works.

Let me know if you have a better way to test this.

packages/tab/src/react/__stories__/index.story.tsx
```diff
@@ -1,14 +1,13 @@
-import PropTypes from 'prop-types'
-import React from 'react'
+import * as React from 'react'
 import { storiesOf } from '@storybook/react'
 
-import Tab from '../index.js'
+import Tab from '../../index.js'
```
jhewlett, line 4: We're only typing the "public interface", so need to import from there

packages/tab/package.json
jaketrent: Hey Justin, again, sorry to be slow.  I've been in and out, mostly out. Thanks for helping me through on this.  You're a trooper, as always.  I miss working with you more.

I think you're testing is great.  It's good to know what format TS is resolving, using what package.json fields.  I don't have any further tests or testing methods I have thought of.

packages/tab/src/react/__stories__/index.story.tsx
jaketrent: Oh, wow, is this required to get the proper typings on `Tab` from within this story file?

packages/tab/package.json
jhewlett: @jaketrent 

All right, I have a solution that seems to check all my boxes:

Easy for consumers to add props via interface augmentation:

```typescript
declare module '@pluralsight/ps-design-system-tab' {
    interface ListItemProps {
        name: string                          //add "name" prop to ListItem
    }
}
```

Typings work with both `cjs` and `esm`

Only need to copy out a single file to the `dist` for each, like so:

```shell
cp src/index.d.ts dist/cjs/index.d.ts`
```
and
```shell
cp src/index.d.ts dist/esm/index.d.ts`
```

This is because I put the declaration straight into the "public interface" `.index.d.ts`

packages/tab/src/react/__stories__/index.story.tsx
jhewlett: So this is a side effect of my most recent change, which collapses the two separate `index.d.ts` files into one, at the outer layer. The advantage of collapsing them is that it makes it easier for consumers to augment the interface to add custom props if they need to[0]. The other advantage is that there's fewer files to copy out into `dist`

I do recognize this as a downside of the change, because it means we wouldn't be able to import anything into a story that's not exported in the other file.

Happy to discuss whether this is worth it.

[0]: With two separate files, the consumer has to do something like this:

```typescript
import Tab from '@pluralsight/ps-design-system-tab'

declare module '@pluralsight/ps-design-system-tab/dist/cjs/react' {
    interface ListItemProps {
        name: string
    }
}
```

(You have to know the internals a bit, notice the `/dist/cjs/react` in the `declare module` augmentation. This is a known bug/design limitation in TS, that it isn't able to handle the re-export for the purposes of augmentation)

With a single file, we can just do:

```typescript
import Tab from '@pluralsight/ps-design-system-tab'

declare module '@pluralsight/ps-design-system-tab' {
    interface ListItemProps {
        name: string    }
}
```

Either way, we would need to document how to do this for consumers, so maybe the first option is acceptable?

jaketrent: Hey Justin, I'm glad you're working on this.  I would not have anticipated this issue.

In my opinion, as a consumer, option 2 is by far better.  The collapsing into a single outer .d.ts file seems worth it to provide the consumer experience.

We good to merge, then?

jhewlett: Yes, I'm happy with where this ended up if you are.

Next steps, I will test this out on the search team, and type one more component to make sure this pattern scales.

PR #931, opened 2020-03-18
fix(site): update site color values to fix build

EdwardIrby:


jaketrent:
It is so weird to me that these are still cropping up.  I feel like I did this weeks ago.

PR #947, opened 2020-03-31
collapsible: ensure it expands completely on page load

### LINKED ISSUE #946: Drawer: sometimes loads only partially open
### Expected behavior
When the user loads the page with a Collapsible component with `isOpen={true}` should be completely open.

### Actual behavior
Okay, this one's a doozy.... *Some* users experience the Collapsible rendering only partially open. They need to click on the drawer to close it then open it again to see the whole thing. 

A screen recording is here: https://pluralsight.slack.com/files/U12F327V4/F01152XNCP8/channel_section_drawer_issue.mov

At the 6 second mark, you can see that all of the Collapsible on the page have rendered only 33px open when they should be 300px+. If you slow down the recording to when the page is loading, you can see that the content of the drawer along with images, fonts, and other assets across the page are not yet rendered for a split second. 

The core of the problem is that the Collapsible should be resizing after the content inside them has finished loading but, for some reason, the height stays at 33px even though the content inside is taller than that. So it looks like the `waitForHeightTransitionToEnd` function is never resolving so the height stays stuck at its hard-coded value instead of being set back to `auto`.

### Steps to repro

This problem has been occurring for some users for months. We've looked into it several times but the only way I can reproduce it is by changing the Collapsible component's code to simulate a problem with detecting when the component is already at its target height.

1. In the Collapsible component, comment out line 27:
```javascript
if (isAlreadyAtTargetHeight) return Promise.resolve()
```

2. Now we need to use collapsible in a way where the Collapsible height changes after it loads. I did that by writing a story in the Drawer's storybook:
```jsx
function DelayedContentLoadStory() {
  const [isOpen, setIsOpen] = useState(true)
  const [contentLines, setContentLines] = useState([1])

  useEffect(() => {
    setTimeout(() => {
      setContentLines([1, 2, 3, 4, 5, 6, 7, 8, 9, 10])
    }, 2000)
  }, [])

  return (
    <div>
      <button type="button" onClick={() => setIsOpen(!isOpen)}>Toggle</button>
      <Collapsible isOpen={isOpen}>
        {contentLines.map(c => (
          <div style={{background: 'blue'}}>{c} of {contentLines.length}</div>
        ))}
      </Collapsible>
    </div>
  )
}
```

3. When the page loads, the Collapsible will be open with one row, then two seconds later, when 9 more rows are added, the Collapsible doesn't expand to show them since its height is still hard-coded because the `waitForHeightTransitionToEnd` function never saw that it was already done animating and didn't resolve. 

I'm thinking the main problem here is that when the page loads, the height of the collapsible is just a tiny bit off (could even be a fraction of a pixel) from the actual number it needs to be. So the line that should know that it's already opened is failing:
```javascript
element.style.height === `${element.getBoundingClientRect().height}px`
```
So the code that sets the height back to `auto` never runs.

I'll submit a PR but one possible solution would be to wait a maximum of the transition duration time before setting the height back to `auto`. That would ensure we're always resetting back to `auto` eventually instead of being stuck at the hard-coded height.

### Related packages

- Collapsible@2.0.16

### Environment

- OS - MacOS 10.12
- Browser version - Chrome v80

KaidenR:
### What You're Solving

This hopefully fixes #946. Here's what I wrote at the bottom of that issue:

I'm thinking the main problem here is that when the page loads, the height of the collapsible is just a tiny bit off (could even be a fraction of a pixel) from the actual number it needs to be. So the line that should know that it's already opened is failing:

```javascript
element.style.height === `${element.getBoundingClientRect().height}px`
```

So the code that sets the height back to auto never runs.

The solution in this PR is to wait a maximum of the transition duration time before setting the height back to auto. That would ensure we're always resetting back to auto eventually instead of being stuck at the hard-coded height.

### How to Verify

1. Without my changes, in the Collapsible component, comment out line 27:

```javascript
if (isAlreadyAtTargetHeight) return Promise.resolve()
```

2. Now we need to use collapsible in a way where the Collapsible height changes after it loads. I did that by writing a story in the Drawer's storybook:

```jsx
function DelayedContentLoadStory() {
  const [isOpen, setIsOpen] = useState(true)
  const [contentLines, setContentLines] = useState([1])

  useEffect(() => {
    setTimeout(() => {
      setContentLines([1, 2, 3, 4, 5, 6, 7, 8, 9, 10])
    }, 2000)
  }, [])

  return (
    <div>
      <button type="button" onClick={() => setIsOpen(!isOpen)}>Toggle</button>
      <Collapsible isOpen={isOpen}>
        {contentLines.map(c => (
          <div style={{background: 'blue'}}>{c} of {contentLines.length}</div>
        ))}
      </Collapsible>
    </div>
  )
}
```

3. Verify that 2 seconds after loading the page, the box does not expand to fit the new content.

4. Pull in this code

5. Verify that 2 seconds after loading the page, the box DOES expand to fit the new content.


KaidenR:
@EdwardIrby Doesn't look like this change has been published out yet. Did I need to bump the version or something?

PR #952, opened 2020-03-31
docs(site): fix layout issue with common icon set in Safari

### LINKED ISSUE #951: site: layout issue in Safari with common icons in icon docs
Label is overlapping the icon.

image

sweeting:
Also made the icons 24x24.

Closes #951

### DIFF
--- a/packages/site/pages/components/icon.js
+++ b/packages/site/pages/components/icon.js
@@ -41,7 +41,7 @@ const CommonSet = props => (
         const Comp = Icons[id]
         return (
           <div className="icon" key={i}>
-            <Comp size={sizes.large} className="svg" />
+            <Comp size={sizes.medium} className="svg" />
             <div className="iconLabel">{id}</div>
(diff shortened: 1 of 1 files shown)


packages/site/pages/components/icon.js
```diff
@@ -41,7 +41,7 @@ const CommonSet = props => (
         const Comp = Icons[id]
         return (
           <div className="icon" key={i}>
-            <Comp size={sizes.large} className="svg" />
+            <Comp size={sizes.medium} className="svg" />
```
jaketrent:
```suggestion
            <Comp size={sizes.medium} />
```

Can also remove this (unused).

sweeting: @EdwardIrby I used the "Commit suggestion" button in Github to incorporate this feedback from Jake, but then it committed without a commit message and that's what's causing the build failure. Not sure how to fix that.

PR #956, opened 2020-04-02
fix(core): missing black color

EdwardIrby:


package.json
```diff
@@ -37,7 +37,7 @@
     }
   },
   "lint-staged": {
-    "*.{css,js}": [
+    "*.js": [
```
EdwardIrby, line 40: @danethurber ran into a weird commitlint error where I couldn't commit the css change unless I removed css from this config it kept throwing on the semicolon in`:root`  on line 1 of `packages/core/css/colors.module.css`

PR #957, opened 2020-04-03
Fix/action menu styles

### LINKED ISSUE #953: action menu styles not building correctly
CSS file built by action menu is missing `.psds-actionmenu` class. Last reported working on 9.1.2. by Carolina Powers

EdwardIrby:
resolve #953


packages/actionmenu/src/react/menu.js
```diff
@@ -16,7 +17,11 @@ const slide = css.keyframes(
 const styles = {
   menu: ({ _isNested, origin }) => {
     return compose(
-      css(stylesheet['.psds-actionmenu']({ slide })),
+      css(stylesheet['.psds-actionmenu']),
+      css({
+        animation: `${slide} ${motion.speedNormal} forwards`
+      }),
+      css(stylesheet['@keyframes psds-actionmenu__keyframes__slide']),
```
jaketrent, line 24: What does including this line here accomplish?

danethurber:
is the build fix simple to apply. i'd rather fix it there

packages/actionmenu/src/react/menu.js
EdwardIrby: `'.psds-actionmenu': ({ slide }) => ({   ` does not build css with the class `'.psds-actionmenu`. That enables both the react code to function and the css to build correctly. I can try and fix the build process but it doesn't properly handle the examples you posted above.

EdwardIrby:
> is the build fix simple to apply. i'd rather fix it there
@danethurber I'll try and dix the build process instead of refactoring the files.

jaketrent:
Here's a proposal on the fix via build: https://github.com/pluralsight/design-system/pull/964

packages/actionmenu/src/react/menu.js
jaketrent: Huh, that's unexpected.  I would have thought that the usage of `glamor.css.keyframes` on line 13 and the application of that on line 22 in `animation` would have been enough to make the React component work.  And then the CSS build doesn't use this file at all, just `css/index.js`.

PR #959, opened 2020-04-06
Core: Add support for TypeScript

jhewlett:
# Core: Add support for TypeScript

I started adding types for another `CircularProgress`, and that story depends on `Core`, so I thought it would make sense to start there.

### Design Decisions

The first commit added an `index.d.ts` following the same pattern as I did with `Tab`. The second commit takes it one step further and just converts it to TypeScript. There were no changes to the code required, just changing the extension to `.ts`, and now it uses `tsc` instead of `babel` to compile it.

If we're not comfortable taking a "production" dependency on TypeScript at this point, we could just go with the first commit. It may be tedious to keep the declaration file in sync with the actual code, but it wouldn't require any changes to the build, so that's the trade-off.

The only thing I wasn't sure about with using `tsc` was which version of EcmaScript to target. The default is `ES3`. I somewhat arbitrarily chose `ES5` but I don't think it would make a difference here.

### How to Verify

- Compare the compiled output between `babel` and `tsc`
- Ensure that running `tsc` in `npm run build` copies out `.d.ts` alongside the other `.js` files in `dist`


packages/core/js/index.d.ts
```diff
@@ -0,0 +1,147 @@
...
+  green: string
+  turquoise: string
+  blue: string
+  purple: string
+  sand: string
+}
+
+export const accessibility: {
+  screenReaderOnly: {
+    position: string
```
jaketrent: Feels like CSS props probably don't need listed out like this.  Just `screenReadOnly: object` ok?

jhewlett: The `.d.ts` no longer applies if we go with the second commit, but I could explicitly add a type to `accessibility.ts` like this:

```typescript
const screenReaderOnly : object = {
  position: 'absolute',
  top: 'auto'
  [...]
}

export default {
  screenReaderOnly
}
```

Is your thinking that this is just an object that they'll be using for some css-in-js and may as well be opaque to them? That accessing the properties within it wouldn't be very helpful (and perhaps confusing)?

jaketrent: Hey Justin, yeah, my comment isn't very applicable in light of your 2nd commit. I don't think this needs any additional treatment.

jhewlett:
@jaketrent Thanks for the review. Do you think this is ready for merge?

jaketrent:
@jhewlett Yes, I'll plan to merge/release today.

PR #960, opened 2020-04-06
refactor(drawer): experimental patterns

danethurber:
Played around with a couple different patterns to see if I could make the external api simpler to use. Ultimately i think this is overkill for the drawer component but could be beneficial for more complicated components. I'll try to put my thoughts in the code section - let me know what questions you have.


packages/drawer/rc/Drawer.js
```diff
@@ -0,0 +1,156 @@
...
+
+import stylesheet from './css/index.js'
+
+import { DrawerContext, useDrawerContext } from './context.js'
+import { actionTypes, useDrawer } from './use-drawer.js'
+import { usePrevious } from './use-previous.js'
+import { combineFns, isDefined, isFunction } from './utils.js'
+
+const styles = {
+  head: (themeName, { isOpen }) =>
```
danethurber, line 18: for style args i like to do theme first, props seconds, and component state third

packages/drawer/rc/Drawer.js
```diff
@@ -0,0 +1,156 @@
...
+    ...props.initialState,
+    ...(controlled && { controlled, isOpen: props.isOpen })
+  }
+
+  const context = useDrawer({
+    initialState,
+    stateReducer,
+
+    onRequestClose: props.onRequestClose,
+    onRequestOpen: props.onRequestOpen
```
danethurber, line 54: a couple nice event functions that be used to build a controlled component without going full on state reducer

packages/drawer/rc/Drawer.js
```diff
@@ -0,0 +1,156 @@
...
+    if (prevOpen === props.isOpen) return
+
+    reset({ isOpen: props.isOpen })
+  }, [controlled, reset, prevOpen, props.isOpen])
+
+  useEffect(() => {
+    if (!isFunction(onChange)) return
+    if (!isDefined(prevOpen) || prevOpen === isOpen) return
+
+    onChange({ isOpen })
```
danethurber, line 77: an onChange function for anytime the open open is changed. i like this prop because it matches the props of native react components like input

packages/drawer/rc/Drawer.js
```diff
@@ -0,0 +1,156 @@
...
+  }, [onChange, prevOpen, isOpen])
+
+  return (
+    <div ref={ref} {...filterReactProps(rest)}>
+      <DrawerContext.Provider value={context}>
+        {isFunction(children) ? children(context) : children}
```
danethurber, line 83: i've found that a render prop still has a lot of value and was easier to use in the stories i played around with. it's also a single line of code to add 💥

packages/drawer/rc/Drawer.js
```diff
@@ -0,0 +1,156 @@
...
+    />
+  )
+})
+Body.displayName = 'Body'
+Body.propTypes = {
+  children: PropTypes.node
+}
+
+Drawer.Head = Head
+Drawer.Head.displayName = 'Drawer.Head'
```
danethurber, line 151: use dot notation for compound component display names

packages/drawer/rc/__stories__/index.story.js
```diff
@@ -1,262 +1,225 @@
...
+      const initialState = { isOpen: true }
+      const stateReducer = (prevState, action, pendingChanges) => {
+        if (action.type === Drawer.types.close && !confirm('allow to close'))
+          return prevState
+
+        if (action.type === Drawer.types.open && !confirm('allow to open'))
+          return prevState
+
+        return pendingChanges
+      }
```
danethurber, line 187: the internal state/actions/types are now part of the public api so we'll have to ensure we mark breaking changes appropriately. might not be that big of a deal but i thought i'd point it out

packages/drawer/rc/__stories__/index.story.js
```diff
@@ -1,262 +1,225 @@
...
+
+    return <Story />
+  })
+
+storiesOf('Drawer | RC / advanced', module)
+  .addDecorator(story => <div style={{ padding: 48 }}>{story()}</div>)
+  .add('default', () => {
+    function Story() {
+      const initialState = { isOpen: true }
+      const stateReducer = (prevState, action, pendingChanges) => {
```
danethurber, line 179: moved the third arg to be the pending changes. this makes sense to me and i think it makes the reducer easier to use.

packages/drawer/rc/__stories__/index.story.js
```diff
@@ -1,262 +1,225 @@
...
+          return prevState
+
+        if (action.type === Drawer.types.open && !confirm('allow to open'))
+          return prevState
+
+        return pendingChanges
+      }
+
+      return (
+        <Drawer initialState={initialState} stateReducer={stateReducer}>
```
danethurber, line 190: this prop api of `initialState` and `stateReducer` is what i came up with. it's still coupling with useDrawer hook with the Drawer component. 

this is different from libraries like [downshift](https://github.com/downshift-js/downshift) where the interaction is in NO way coupled to the view. that would probably return a bunch of prop getter and then be used to build more customized views

packages/drawer/rc/__stories__/use-drawer.story.js
```diff
@@ -0,0 +1,54 @@
...
+import React from 'react'
+
+import { storiesOf } from '@storybook/react'
+
+import Button from '@pluralsight/ps-design-system-button'
+import * as Text from '@pluralsight/ps-design-system-text'
+
+import { useDrawer } from '../index.js'
+
+storiesOf('Drawer | RC / useDrawer', module)
```
danethurber, line 12: the useDrawer hook can be used separately from the Drawer component

packages/drawer/rc/index.js
```diff
@@ -1,151 +1,2 @@
...
-}
-Drawer.Head = Head
-Drawer.Body = Body
-Drawer.types = {
-  toggleDrawer: 'TOGGLE_DRAWER',
-  openDrawer: 'OPEN_DRAWER',
-  closeDrawer: 'CLOSE_DRAWER'
-}
+export { Drawer } from './Drawer.js'
+export { actionTypes, useDrawer } from './use-drawer.js'
```
danethurber, line 2: separated out the hook/state from the component

packages/drawer/rc/index.js
```diff
@@ -1,151 +1,2 @@
...
-  const value = React.useMemo(
-    () => ({ ...(open !== undefined ? { open, onToggle } : uncontrolled) }),
-    [open, onToggle, uncontrolled]
-  )
-  return (
-    <DrawerContext.Provider value={value}>{children}</DrawerContext.Provider>
-  )
-}
-
-Head.displayName = 'Drawer(Head)'
```
danethurber, line 127: this style of display name should be used for HOCs that "enhance" a component

packages/drawer/rc/use-drawer.js
```diff
@@ -0,0 +1,81 @@
...
+  pick,
+  removeUndefinedValues,
+  switchcaseF
+} from './utils.js'
+
+export const actionTypes = {
+  close: 'CLOSE',
+  open: 'OPEN',
+  reset: 'RESET'
+}
```
danethurber, line 16: simplified the types because adding drawer felt redundent

packages/drawer/rc/use-drawer.js
```diff
@@ -0,0 +1,81 @@
...
+  persistSynthetic,
+  pick,
+  removeUndefinedValues,
+  switchcaseF
+} from './utils.js'
+
+export const actionTypes = {
+  close: 'CLOSE',
+  open: 'OPEN',
+  reset: 'RESET'
```
danethurber, line 15: reset can be used when the component is controlled and receives new state from props

packages/drawer/rc/use-drawer.js
```diff
@@ -0,0 +1,81 @@
...
+    [actionTypes.close]: controlled ? prevState : { isOpen: false },
+    [actionTypes.open]: controlled ? prevState : { isOpen: true },
+    [actionTypes.reset]: { ...prevState, ...action.payload }
+  }
+
+  const handleUnknown = () => {
+    throw new Error(`Unhandled type: ${action.type}`)
+  }
+
+  return switchcaseF(cases)(handleUnknown)(action.type)
```
danethurber, line 32: i hate native switch cases. this is better/faster

packages/drawer/rc/use-drawer.js
```diff
@@ -0,0 +1,81 @@
...
+    reset: createActionCreator(actionTypes.reset)(dispatch)
+  }
+
+  const close = combineFns(
+    persistSynthetic,
+    onRequestClose,
+    actionCreators.close
+  )
+  const open = combineFns(persistSynthetic, onRequestOpen, actionCreators.open)
+  const toggle = state.isOpen ? close : open
```
danethurber, line 67: toggle doesn't have an action because it wasn't idempotent. it's easy enough to use open/close actions and retain idempotency(sp?)

this also means that the public api we expose(either through event props or the external stateReducer) is easier to deal with because it's either an open or close event/action

packages/drawer/rc/utils.js
```diff
@@ -0,0 +1,38 @@
+/* eslint-disable no-prototype-builtins */
+
```
danethurber, line 2: these functions are generic enough we can move to a shared utils package

packages/drawer/rc/use-drawer.js
```diff
@@ -0,0 +1,81 @@
...
+    close,
+    open,
+    toggle,
+    reset: actionCreators.reset
+  }
+}
+
+useDrawer.types = actionTypes
+
+const reducerNoop = (state, actions, pendingChanges) => pendingChanges
```
danethurber, line 81: changed the state reducer api to have the pendingChanges as the third arg

packages/drawer/rc/utils.js
```diff
@@ -0,0 +1,38 @@
...
+export const createActionCreator = type => dispatch => payload =>
+  dispatch({ payload, type })
```
danethurber, line 4: this partially implements the flux standard action pattern. we can expand with errors/meta/whatever as needed

packages/drawer/rc/use-drawer.js
```diff
@@ -0,0 +1,81 @@
...
+  switchcaseF
+} from './utils.js'
+
+export const actionTypes = {
+  close: 'CLOSE',
+  open: 'OPEN',
+  reset: 'RESET'
+}
+
+const defaultState = { controlled: false, isOpen: true }
```
danethurber, line 18: changed from `open` to `isOpen` so it was more explicit and didn't clash with the `open` action

PR #962, opened 2020-04-06
feat: simpler reimplementation of drawer

### LINKED ISSUE #937: Breakout Drawer functionality
New pattern experiment

Breakout drawer functionality so it can be assembled by a user from base elements and hooks. This does not effect existing drawer. It is intended as a test of a new pattern that allows developers to build for their needs from lower level components and functionality:

> Hey, would it be possible to add startOpento the useEffect for the Drawer component? We have a use case where we are changing startOpen conditionally based on whether the drawers are search results, and we can’t get them to rerender to be open unless we pass in isOpen which breaks the ability to close them. ~ @kristiegiles

EdwardIrby:
resolves #937


packages/drawer/rc/index.js
```diff
@@ -80,47 +81,19 @@ const Body = React.forwardRef(({ children, ...rest }, forwardedRef) => {
...
-  const uncontrolled = useDrawer()
-  const value = React.useMemo(
-    () => ({ ...(open !== undefined ? { open, onToggle } : uncontrolled) }),
-    [open, onToggle, uncontrolled]
-  )
+export const Drawer = ({ children }) => {
+  const [open, setOpen] = useState(false)
   return (
-    <DrawerContext.Provider value={value}>{children}</DrawerContext.Provider>
+    <DrawerProvider open={open} onToggle={() => setOpen(!open)}>
```
EdwardIrby, line 94: Not the best name but it occurred to me while the stateReducer patterns might be good for something like Autocomplete/Typeahead and DateInput where we have numerous microinteractions it's probably not the best for single microinteraction components like Drawer or toggle. We've talked about doing more composability and building lower level components. That why I like this pattern for simpler interactive components where a developer may want to take over control.

packages/collapsible/rc/index.js
```diff
@@ -50,7 +50,7 @@ const waitForHeightTransitionToEnd = element => {
 }
 const updateOverflowStyle = ({ element, open, isTransitioning = false }) => {
   if (element) {
-    element.style.overflow = isTransitioning || !open ? 'hidden' : 'visible'
+    element.style.overflow = isTransitioning || !open ? 'hidden' : 'auto'
```
EdwardIrby, line 53: If a user places a paragraph or another child element with margins inside a collapsible parent this style will prevent unexpected rendering such as gaps from appearing such as in the Drawer.

packages/drawer/rc/Readme.md
```diff
@@ -0,0 +1,113 @@
+# Drawer(Release Candidate)
```
EdwardIrby, line 1: I feel we should write a separate doc for useCallapsible and useToggle if this pattern is adopted maybe something on the DS site showing how the Drawer was built from these base components.

packages/collapsible/rc/index.js
```diff
@@ -0,0 +1,102 @@
+import { useCallback, useState, useEffect } from 'react'
```
EdwardIrby, line 1: I think if we accept this hook it should probably be moved to either utils or hooks specific package or should we start creating hooks packages?

danethurber: since it's part of Collapsible it's probably fine for now. if it was more generic(like useToggle) it say we put it with utils or core maybe

packages/drawer/.storybook/config.js
```diff
@@ -7,9 +7,11 @@ import themeDecorator from '@pluralsight/ps-design-system-storybook-addon-theme'
 addDecorator(themeDecorator)
 
 const req = requireContext('../src', true, /\.story\.js$/)
+const rc = requireContext('../rc', true, /\.story\.js$/)
 
 function loadStories() {
   req.keys().forEach(filename => req(filename))
+  require('../rc/__stories__/index.story');
```
danethurber, line 14:
```suggestion
  rc.keys().forEach(filename => rc(filename))
```

packages/drawer/rc/Readme.md
danethurber: agreed. if we export hooks for use they should have matching documentation on the site

packages/drawer/rc/__specs__/index.spec.js
```diff
@@ -0,0 +1,52 @@
+import React, { useState } from 'react'
+import { fireEvent, render } from '@testing-library/react'
+
+import { Drawer } from '../index.js'
+
+describe('Drawer', () => {
+  describe('when the arrow icons is clicked', () => {
```
danethurber, line 7: this test isn't testing the arrow buttons but the whole head el

packages/drawer/rc/__specs__/index.spec.js
```diff
@@ -0,0 +1,52 @@
...
+      const btn = getByTestId('head')
+      const contentWrapper = getByTestId('body')
+      expect(contentWrapper).toHaveAttribute('aria-hidden', 'true')
+      fireEvent.click(btn)
+      expect(contentWrapper).toHaveAttribute('aria-hidden', 'false')
+      fireEvent.click(btn)
+      expect(contentWrapper).toHaveAttribute('aria-hidden', 'true')
+    })
+  })
+  const Controlled = () => {
```
danethurber, line 29: i'd say just move this component into the test because it's only used in one place

packages/drawer/rc/__stories__/index.story.js
```diff
@@ -0,0 +1,268 @@
...
+  .addDecorator(story => <div style={{ padding: 48 }}>{story()}</div>)
+  .add('default', () => {
+    return (
+      <Drawer>
+        <Drawer.Head>
+          <DrawerHeadContent>Click me to open</DrawerHeadContent>
+        </Drawer.Head>
+        <Drawer.Body>
+          <DrawerBodyContent>
+            Ambulance and emergency sirens in New York City are wailing
```
danethurber, line 94: some heavy filler text...

EdwardIrby: Yeah I was just copying the first few paragraphs text from the last news article I had open didn't even realize it until now.

packages/drawer/rc/__specs__/index.spec.js
EdwardIrby: Copy/Pasta error from previous drawer

packages/collapsible/rc/index.js
EdwardIrby: I was thinking if we adopt the new drawer then the collapsible component could either stay as is but this new hook which renders no markup might need to co somewhere else. We actually could use it to update vertical tabs.

PR #964, opened 2020-04-07
fix(build): support building selectors as functions with glamor

jaketrent:
see https://github.com/pluralsight/design-system/pull/957


EdwardIrby:
closes #957 resolves #953

PR #965, opened 2020-04-09
feat(circularprogress): add .dts for typescript consumption

jhewlett:
### What You're Solving

This adds TypeScript types for `circularprogress`

With this in place, we would have 2 reference implementations, which gives us some confidence that the pattern can scale. If we're happy with both of these, we could document how to add types and empower teams to submit their own types.

### Design Decisions

- I explored putting the `Styling` interface in a common location. The problem with this is that we would then have to copy out the extracted file and have it available for the `.d.ts` to import, or else somehow inline it. Duplicating seemed like the best option for now.

### How to Verify

- Component can be consumed from TypeScript, with the `strict` flag enabled, without having to declare any types


packages/circularprogress/.storybook/main.js
```diff
@@ -0,0 +1,3 @@
+module.exports = {
+  presets: ['@storybook/preset-typescript']
+}
```
jhewlett, line 3: We might want to look into doing this in the top-level of the repo if it reduce friction to add types to future packages.

packages/circularprogress/src/index.d.ts
```diff
@@ -0,0 +1,24 @@
...
+}
+
+type Sizes = {
+  small: 'small',
+  medium: 'medium'
+}
+
+export const sizes: Sizes
+
+declare const CircularProgress: ForwardRefExoticComponent<CircularProgressProps & Styling & RefAttributes<HTMLDivElement>>
```
jhewlett, line 21: This took a bit of time to figure out the correct type. The biggest help was just seeing what `react` says is the return type of `React.forwardRef`

packages/circularprogress/src/react/__stories__/index.story.tsx
```diff
@@ -1,26 +1,26 @@
...
+import CircularProgress from '../../index.js'
 
 const valueStory = storiesOf('value', module)
 ;[0, 25, 50, 75, 100, 33, 66].forEach(value =>
   valueStory.add(`value ${value}`, _ => <CircularProgress value={value} />)
 )
 
 const sizeStory = storiesOf('size', module)
-Object.keys(CircularProgress.sizes).forEach(size =>
+Object.values(CircularProgress.sizes).forEach(size =>
```
jhewlett, line 13: Switched to `Object.values` because `Object.keys` has a return type of `string[]`, and we want `('small' | 'medium')[]` so that it will typecheck when passing in as the prop to `CircularProgress`, which only accepts `'small' | 'medium'`

jhewlett: An alternative is to do some casting

PR #973, opened 2020-04-13
CircularProgress: typing updates

danethurber:
@jhewlett let me know what you think. i was attempting to make the component typings simpler to read and replicate. i had some patterns that i've been using on another project that i wanted to run by you

### DIFF
--- a/packages/circularprogress/src/index.d.ts
+++ b/packages/circularprogress/src/index.d.ts
@@ -1,24 +1,40 @@
-import { ForwardRefExoticComponent, RefAttributes } from 'react'
+import { ForwardRefExoticComponent, HTMLAttributes, RefAttributes } from 'react'
 
-interface Styling {
-  css?: Record<string, string | number>
-  style?: Record<string, string | number>
-  className?: string
-}
+declare const CircularProgress: RefForwardingComponent<
+  CircularProgressProps,
+  HTMLDivElement,
+  CircularProgressStatics
+>
+
+export default CircularProgress
 
 export interface CircularProgressProps {
-  size?: 'small' | 'medium'
+  size?: keyof typeof sizes
   value?: number
 }
 
-type Sizes = {
-  small: 'small',
-  medium: 'medium'
+export interface CircularProgressStatics {
+  sizes: typeof sizes
 }
 
-export const sizes: Sizes
+export const sizes = keyMirror('small', 'medium')
 
-declare const CircularProgress: ForwardRefExoticComponent<CircularProgressProps & Styling & RefAttributes<HTMLDivElement>>
-  & { sizes: Sizes }
+// TODO: relocate to a global typings file
+export type RefForwardingComponent<
+  Props = {},
+  El = Element,
+  Statics = {}
+> = ForwardRefExoticComponent<Props & HTMLAttributes<El> & RefAttributes<El>> &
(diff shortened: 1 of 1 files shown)


packages/circularprogress/src/index.d.ts
```diff
@@ -1,24 +1,39 @@
-import { ForwardRefExoticComponent, RefAttributes } from 'react'
+import { ForwardRefExoticComponent, HTMLAttributes, RefAttributes } from 'react'
 
-interface Styling {
-  css?: Record<string, string | number>
```
danethurber, line 4: `css` isn't needed anymore. this is something we should have removed when we dropped glamourous

packages/circularprogress/src/index.d.ts
```diff
@@ -1,24 +1,39 @@
...
-  style?: Record<string, string | number>
-  className?: string
```
danethurber, line 6: these are react/dom native attrs and we shouldn't have to define them for all components. we can leverage the `HTMLAttributes` typing to get the list of this props. if we were to do this is our react source we wouldn't need the filterReactProps util anymore

packages/circularprogress/src/index.d.ts
```diff
@@ -1,24 +1,39 @@
...
+  HTMLDivElement,
+  Statics
+>
+
+export default CircularProgress
 
-export interface CircularProgressProps {
-  size?: 'small' | 'medium'
+export interface Props {
+  size?: keyof typeof sizes
```
danethurber, line 12: generate the types from the sizes src/const

packages/circularprogress/src/index.d.ts
```diff
@@ -1,24 +1,39 @@
...
+  Props,
+  HTMLDivElement,
+  Statics
+>
+
+export default CircularProgress
 
-export interface CircularProgressProps {
-  size?: 'small' | 'medium'
+export interface Props {
```
jhewlett, line 11: The reason I named this `CircularProgressProps` is to follow a convention of `{ComponentName}Props`, which would provide users with a consistent way to augment the props interface if they need to, like so:

```typescript
declare module '@pluralsight/ps-design-system-tab' {
    interface ListItemProps {
        name: string                          //add "name" prop to ListItem
    }
}
```

It feels redundant with single-export components, but makes more sense when you consider things like `Tab` that have multiples, and this way there would only be one convention to remember.

That said, now that you've added the `HTMLAttributes` in to the props, does that cover all the props that a user might provide (and would actually get forwarded on?) If so, then there's probably less need to augment, and I don't really care either way.

packages/circularprogress/src/index.d.ts
```diff
@@ -1,24 +1,39 @@
...
-  small: 'small',
-  medium: 'medium'
+export interface Statics {
+  sizes: typeof sizes
 }
+export const sizes = keyMirror('small', 'medium')
 
-export const sizes: Sizes
+// TODO: relocate to a global typings file
+export type RefForwardingComponent<
```
jhewlett, line 22: I really like this helper!

packages/circularprogress/src/index.d.ts
jaketrent: I like the `{ComponentName}Props` convention.

PR #979, opened 2020-04-20
Fix/search icon

EdwardIrby:


packages/layout/src/css/page-heading-layout.js
```diff
@@ -2,7 +2,7 @@ import { layout } from '@pluralsight/ps-design-system-core'
 
 export default {
   '.psds-page-heading-layout': {
-    padding: `${layout.spacingLarge} 0`
+    padding: layout.spacingLarge
```
danethurber, line 5: is the horz padding not needed anymore?

packages/searchinput/src/css/index.js
```diff
@@ -3,5 +3,10 @@ import { layout } from '@pluralsight/ps-design-system-core'
 export default {
   '.psds-searchinput-clear': {
     margin: `0 ${layout.spacingXXSmall}`
+  },
+  '.psds-searchinput-field': {
+    '&::-webkit-search-cancel-button': {
+      display: 'none'
+    }
```
danethurber, line 10: 👍

packages/layout/src/css/page-heading-layout.js
EdwardIrby: No I realized after publishing earlier merges a change I marked as breaking didn't bump to a new major. I moved this back so that people using that version would not experience breakage. I just created a new branch that will hopefully bump the version of layout to a new major.

PR #983, opened 2020-04-23
fix(searchinput): clear icon background

EdwardIrby:
Fixes https://pluralsight.slack.com/archives/C70HPSJPP/p1587589348077800


sweeting:
Contrast is good, I believe that's a flat button. The clear button should only appear when text has been entered in the field. The alignment should be shifted to the right by 4px.

Here's a screenshot of the desired state:
image

EdwardIrby:
> Contrast is good, I believe that's a flat button. The clear button should only appear when text has been entered in the field. The alignment should be shifted to the right by 5px.
> 
> Here's a screenshot of the desired state:
> image

I'll modify interaction code to fix appearance of clear button.

packages/searchinput/src/css/index.js
```diff
@@ -1,8 +1,14 @@
-import { layout } from '@pluralsight/ps-design-system-core'
+import { layout, motion } from '@pluralsight/ps-design-system-core'
 
 export default {
   '.psds-searchinput-clear': {
-    margin: `0 ${layout.spacingXXSmall}`
+    margin: `0 0 0 ${layout.spacingXXSmall}`,
```
EdwardIrby, line 5: X looked a little to far from the edge

packages/searchinput/src/react/index.js
```diff
@@ -1,72 +1,84 @@
...
 
 SearchInput.defaultProps = {
   loading: false
 }
 
 SearchInput.propTypes = {
   loading: PropTypes.bool,
-  onClear: PropTypes.func
+  onClear: PropTypes.func,
+  onChange: PropTypes.func
```
jaketrent, line 81: What's this for?  Does this address the original legibility problem from Kristie?  If not, why include it?

EdwardIrby: - **onChage**? I just need to destructure that out and decided to change the way the prop destructuring was being done so it required a propType declaration. 
- **onClear** has been part of the component prior to this. 

The reason I desctructured **onChange** out was to address the [interaction](https://github.com/pluralsight/design-system/pull/983#issuecomment-618477680) @sweeting  mentioned above. That functionality was not actually part of the component. These changes address both Brian and Carolina's [accessibility issue](https://pluralsight.slack.com/archives/C70HPSJPP/p1587589348077800) for the original slack post mentioned when I opened the PR.

jaketrent: Ah yeah, thank you for helping me remember the interaction in question.  Tried in storybook, looks great.

Issue #934, opened 2020-03-24
ps-design-system-table: Better sorting handling


matthewma7:
## Enhancement Request

Note: Here are two separate but related feature requests denoted by number 1 and 2

### Desired behavior

1. Pass original event to onClick handler for `<Table.ColumnHeader />`
2. Add un sort prop option

### Today's behavior

1. onClick event only pass sortDir, one argument
2. It can only sort between `asc` and `desc`, can't unsort

### The Value Add

1. This helps detect if a modifier key is pressed to handle sorting by multiple columns
2. Make it possible to unsort a column

### The Tradeoffs

No obvious tradeoffs

EdwardIrby:
- Break change pass event as the first param and sort direction as the second parameter. 
- Will revisit on the un sort.
@matthewma7 please contact @sweeting to provide further information about un sort use case.

jaketrent:
Idea: Just pass original event as 2nd param, don't need to make breaking change.  We have event passed last on lots of other components, nbd.

matthewma7:
I think what @jaketrent describes would work for my use case.

jaketrent:
Matthew, we think this is done.  We did the event add to the column sort handler.  We are still unsure about what unsort could/should do.  We'll close this for now.  If you still need support, please contact us with more unsort details directly.

PR #985, opened 2020-04-24
table: add sort event to sort handler

jaketrent:
Related: #934 -- resolves only one of the issues mentioned in the GH issue.

### DIFF
--- a/packages/table/src/react/__specs__/index.spec.js
+++ b/packages/table/src/react/__specs__/index.spec.js
@@ -6,8 +6,9 @@ import Table from '../index.js'
 describe('columnHeader', () => {
   test('sorted=false onClick called with asc', () => {
     return new Promise(resolve => {
-      const onClick = sort => {
+      const onClick = (sort, evt) => {
         expect(sort).toBe(Table.sorts.asc)
+        expect(evt).toBeDefined()
         resolve()
       }
       const { getByText } = render(
@@ -19,8 +20,9 @@ describe('columnHeader', () => {
 
   test('sorted=true onClick called with asc', () => {
     return new Promise(resolve => {
-      const onClick = sort => {
+      const onClick = (sort, evt) => {
         expect(sort).toBe(Table.sorts.asc)
+        expect(evt).toBeDefined()
         resolve()
       }
       const { getByText } = render(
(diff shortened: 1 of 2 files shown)


packages/table/src/react/__specs__/index.spec.js
```diff
@@ -6,8 +6,9 @@ import Table from '../index.js'
 describe('columnHeader', () => {
   test('sorted=false onClick called with asc', () => {
     return new Promise(resolve => {
-      const onClick = sort => {
+      const onClick = (sort, evt) => {
         expect(sort).toBe(Table.sorts.asc)
+        expect(evt).toBeDefined()
         resolve()
       }
```
danethurber, line 13: evt as first argument would be more consistent with react general practices...but that's a breaking change. would could break now or later 🤷

jaketrent: Yeah, I want to be consistent one way or the other.  

We have a mix today:

After:
```
drawer/src/react/index.js                                                                                      
72:    onToggle && onToggle(nextOpen, evt)                                                                     
                                                                                                               
starrating/src/react/index.js                                                                                  
34:    props.onChange(evt.target.value, evt)                                                                   
37:  function handleStarClicked(index, evt) {                                                                  
39:    props.onChange(index + 1, evt)                                                                          
```

Before:
```
checkbox/src/react/index.js                                                                                    
59:      onCheck(evt, !props.checked, props.value, props.name)                                                 
                                                                                                               
dropdown/src/react/index.js                                                                                    
146:  function handleMenuChange(evt, value, label) {                                                           
150:      allProps.onChange(evt, value, label)                                                                 
                                                                                                               
typeahead/src/react/index.js                                                                                   
84:  const handleChange = (evt, nextValue) => {                                                                
88:    if (onChange) onChange(evt, nextValue)                                                                  
98:    handleChange(evt, nextValue)                                                                            
101:  const handleSuggestionMenuChange = (evt, nextValue) => {                                                 
103:    handleChange(evt, nextValue)                                                                           
                                                                                                               
viewtoggle/src/react/index.js                                                                                  
48:  function handleSelect(evt, index) {                                                                       
51:    if (evt && isFunction(onSelect)) onSelect(evt, index)                                                   
119:    props._onSelect(evt, props._i) 
```

What's the React general practice?  Where could I observe it?

PR #989, opened 2020-04-24
fix(carousel): update tabindex for acceisbility request

### LINKED ISSUE #969: Carousel: accessibility improvements
- carousel requests
  - move controls to the first of the order
  - currently all items tab through
    - only tab through visible cards
  - pagination controls are nice to have
  - don’t focus the whole carousel, just jump to the first item

EdwardIrby:
resolves #969


danethurber:
https://www.figma.com/file/7WHzor2BQxJekPuObohvzi/Carousel-Card-Focus-States?node-id=0%3A1

so designs

packages/carousel/src/react/index.js
```diff
@@ -84,7 +85,7 @@ export default function Carousel({ controls, size, ...props }) {
 
                   return cloneElement(item, {
                     key: itemIndex,
-
+                    tabIndex: isActivePage ? 0 : -1,
```
EdwardIrby, line 88: just remove this prop all together

jaketrent:
Re the designs linked above:

Those look to be updated designs for focus states on cards.  I think those should be done separate from this carousel tab a11y update.

packages/carousel/src/react/index.js
```diff
@@ -63,6 +63,7 @@ export default function Carousel({ controls, size, ...props }) {
         {...filterReactProps(props)}
         ref={ref}
       >
+        {controls}
```
jaketrent, line 66: The ordering change is so that controls come first in tab order, right?  And the associated pointer-event changes so we can still click stuff in the carousel, right?

packages/carousel/src/react/index.js
```diff
@@ -185,36 +183,37 @@ const Pages = React.forwardRef((props, ref) => {
...
       ref={ref}
       role="region"
-      tabIndex="0"
       {...styles.pages(props)}
       {...filterReactProps(props)}
     />
   )
 })
 
+Pages.displayName = 'Carousel.Pages'
```
jaketrent, line 192: Great catch

packages/carousel/src/react/index.js
```diff
@@ -185,36 +183,37 @@ const Pages = React.forwardRef((props, ref) => {
...
   const { offset } = React.useContext(CarouselContext)
-
   return (
     <li
       ref={ref}
       {...styles.page(props)}
       {...css({ transform: `translate3d(${offset}px, 0, 0)` })}
       {...rest}
-      {...(!isActivePage && { hidden: true, tabIndex: -1 })}
+      {...(!isActivePage && { hidden: true })}
```
jaketrent, line 210: These must have been tabIndex -1 for a reason (programmatic focus).  Where was that happening?  I don't see it removed in this diff.

packages/carousel/src/react/index.js
EdwardIrby: Yep

packages/carousel/src/react/index.js
EdwardIrby: **li** by default have no tabinddex. The purpose of `tabIndex: -1` is to explicitly make sure an element is not reachable via sequential keyboard navigation, but could be focused with Javascript or visually.  This particular **li** is the Page or visible section of the **Carousel**.  One of the objectives was remove tab focus from this element. This is just a semantically clean implementation.

EdwardIrby:
> The tab order for controls has taken a step in the right direction. I don't think that part's done. I especially don't think we can say that #969 is resolved by this PR. That issue contains several issues that don't seem addressed here.
> 
> Anyway, here are the issues I see to be solved in order to resolve the bug:
> 
> 1. Move controls to the first of the order
> 
> This is addressed in the PR. They come first in order. This is very nice. But the interaction seems imperfect in this case: I tab to the "next page" arrow (the common case), I advance to the page I want, and then I tab. I would expect the tab would focus the first item visible in the current carousel page. It does not currently. Currently, it focuses the first item in the entire list (offscreen).
> 
> 2. Only tab through visible cards
> 
> Completing no. 1 will do half of no. 2 here. The other half, I believe, is that when we tab past the last visible item in the current page, the "next page" arrow should be focused. And when we shift-tab past the first visible item in the current page, the "previous page" arrow should be focused.
> 
> 1. Don’t focus the whole carousel, just jump to the first item
> 
> Done.

With Regard to point 2 in the above comment I don't think that is correct as I read the issue after confirming with @danethurber. I understood that the carousel items were also not to be tabbed. If this is incorrect I could got back and undo one of my commits where I made only visible items tabbed.

Further I don't believe that even if the items were tabbed, once one reached the end of the list they'd hit the next arrow button. I understood it as controls first both forward and backward. I'm assuming the reasoning here is that the items could be forms and having to tab through each for a carousel page would not be fun just to see the next page of items in the carousel list.  

Maybe we should check with accessibility on that. However I understood that as the controls are first both controls. What you've described is counter to that.

packages/carousel/src/react/index.js
jaketrent: Yeah, the fact that tabIndex=-1 is used for programmatic focus was the point of my question.  Where is the code that was programmatically focusing?  I don't see it removed.  Maybe it should be?

jaketrent:
Hey Edward, thanks for looking at the observations I made.  Let me continue to try to clarify what I'm thinking...

Regarding Point 2 only:

Agreed, we should not focus *Carousel Items*.  My impression was that Amber told us this in our last meeting.  The things *within* the Carousel Items should be tabbable. Ie, the title link on a Card.

Now, *which* Card titles should be tabbable?  My impression was that Amber told us in our meeting that only the Card titles within the Carousel Items of the current page should be tabbable.  This impressions seems matched by how Dane recorded the issue:

> currently all items tab through; only tab through visible cards

This also matches her rule of thumb that we should treat tabbing like a mouse-using using user would treat the mouse.  In this case, when I get to the end of a carousel page with a mouse, I click the next.  Or, when tabbing, I get to the end of Card titles visible on the current page, and I tab to the "next page" button.  When the page is advanced, I work through the next page, starting my tab focus on the first Card title in the first Carousel item of the now-visible next page.

As I've reiterated, this is my impression.  Sounds like we're either talking about different things or have different impressions on what we learned and what is preferred from an accessibility standpoint.  

I agree, let's figure that out by inviting Amber to clarify by reteaching us what should happen has a user presses tab through a Carousel.

jaketrent:
Got some [feedback from Amber Holladay in Slack](https://pluralsight.slack.com/archives/C5CMEKH5Z/p1588284923041600):

Amber:
Everything inside of the carousel needs to be keyboard accessible, which it currently is, but I have to TAB through all 20+ carousel cards to get to the next content on the page. (Screen reader access is similar with having to DOWN ARROW through all of the cards.)
For accessibility, I should be able to TAB (or DOWN ARROW) through the visible cards then TAB to the "next" button.

We had also discussed "hiding" the mouseover links (Play, Bookmark, More menu) so a user does not have ~4 TAB stops for every card. This is lower priority for me, but it does make it quicker to get to content using keyboard only.

In that instance, there would be a link on the card to "show more" which would give users access to those 3 links.

There is not an accessibility reason to have a TAB stop on the entire carousel before TABbing to the cards inside of the carousel. We had discussed removing that (again lower priority)

jaketrent:
@EdwardIrby Did this get discussed further in DS office hrs with Amber?  

Just wondering given the conversations that have happened what your intention is with this going forward.  What direction will you take it?

EdwardIrby:
@jaketrent, @sweeting and @danethurber best way to review this is run the site and go to http://localhost:1337/components/carousel

packages/carousel/src/react/index.js
```diff
@@ -179,45 +177,57 @@ const Pages = React.forwardRef((props, ref) => {
...
+const Page = props => {
   const ref = React.useRef()
 
-  const { children, isActivePage, ...rest } = props
+  const { children, isActivePage, paged, ...rest } = props
+  React.useEffect(() => {
+    if (paged && isActivePage) {
+      const page = ref.current
+      const focusFirstChild = () => page.firstElementChild.focus()
+      page.addEventListener('transitionend', focusFirstChild)
```
jaketrent, line 207: nice

PR #990, opened 2020-04-25
fix(position): no longer clones child styles onto tooltip

### LINKED ISSUE #893: Tooltip copies props of wrapped element
https://github.com/pluralsight/design-system/blob/65ffbfc430f20525c2b512593e5f11069a890db3/packages/tooltip/src/react/index.js#L77

Any reason this doesn't allow another element?

The XY problem we're trying to solve here is, the element we're wrapping in `Above` has `color:#fff`. (https://github.com/ps-dev/plan-analytics/blob/skills/client/lib/reports/all-skills/index.tsx#L480) That style is being applied to the tooltip for some reason (I'm having trouble finding in the `Tooltip` code where/why that might be happening (maybe `filterReactProps`?)) so the text in the tooltip is not visible. So we're wrapping our tooltip text in a `p` tag with `color:#000`. (Putting `style` directly on `Tooltip` is being overridden.)

I see other issues where the response is to use a dialog, but this is semantically a tooltip and is non-actionable.

We don't need the `p` element if our styles don't get copied to the tooltip. Any suggestions?

cc @twclark0

EdwardIrby:
resolves #893

### DIFF
--- a/packages/position/src/react/index.js
+++ b/packages/position/src/react/index.js
@@ -59,7 +59,7 @@ const Position = React.forwardRef((props, forwardedRef) => {
   const showRef = React.useRef()
   const showEl = React.cloneElement(props.show, {
     ref: showRef,
-    style: { ...child.props.style, ...style }
+    style: { ...style }
   })
 
   const updateStyle = React.useCallback(() => {
--- a/packages/position/src/react/index.js
+++ b/packages/position/src/react/index.js
@@ -59,7 +59,7 @@ const Position = React.forwardRef((props, forwardedRef) => {
   const showRef = React.useRef()
   const showEl = React.cloneElement(props.show, {
     ref: showRef,
-    style: { ...style }
+    style: { ...props.show.props.style, ...style }
   })
 
   const updateStyle = React.useCallback(() => {
--- a/package-lock.json
+++ b/package-lock.json
@@ -467,6 +467,15 @@
         "@babel/helper-plugin-utils": "^7.8.0"
       }
     },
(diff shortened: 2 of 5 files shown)


packages/position/src/react/index.js
```diff
@@ -59,7 +59,7 @@ const Position = React.forwardRef((props, forwardedRef) => {
   const showRef = React.useRef()
   const showEl = React.cloneElement(props.show, {
     ref: showRef,
-    style: { ...child.props.style, ...style }
+    style: { ...style }
```
jaketrent:
```suggestion
    style
```

packages/position/src/react/index.js
```diff
@@ -59,7 +59,7 @@ const Position = React.forwardRef((props, forwardedRef) => {
   const showRef = React.useRef()
   const showEl = React.cloneElement(props.show, {
     ref: showRef,
-    style: { ...child.props.style, ...style }
+    style: { ...props.show.props.style, ...style }
```
jaketrent, line 62: I believe that the fix should be transfer `show.props.style` to the cloned `props.show` element instead of removing it.

[cloneElement](https://reactjs.org/docs/react-api.html#cloneelement) does shallow merge.  But, remember, shallow is at the first level properties, where the style property is.  It doesn't merge any deeper, including the style object properties.  I added a story to verify this fact and need to keep this.  Please verify and see what you think.

jaketrent:
@EdwardIrby , if you're cool with this one and it solves the bug you're solving, I think it's ready to merge.

PR #992, opened 2020-04-30
Fix/dropdown add portal

### LINKED ISSUE #779: dialog: Inner dimensions too large when using a Dropdown
### Expected behavior

See a `Dropdown` inside of the `Dialog` without resizing the `Dialog`

### Actual behavior

Scrollbars appear on the `Dialog`. When opening the `Dropdown`, the internal dimensions of the `Dialog` become as large as the viewport.

Before opening the `Dropdown`:

Before opening the dropdown

After opening the `Dropdown`:

After opening the dropdown

[Repository which reproduces the problem](https://github.com/MCluck90/dropdown-in-dialog-ps-design-system)

### Steps to repro

1. Create a basic `Dialog` (ex. `<Dialog aria-label="Dialog"> ... </Dialog>`)
2. [Copy the example code for a `Dropdown`](https://design-system.pluralsight.com/components/dropdown)
3. Paste the code in to the `Dialog`

### Related packages

- actionmenu@^9.0.9
- core@^5.1.0
- dialog@^7.0.13
- dropdown@^5.0.11

### Environment

- OS - Windows 10
- Browser version - Chrome v79

EdwardIrby:
resolves #779


packages/dropdown/src/react/__stories__/index.story.js
```diff
@@ -8,7 +8,7 @@ import * as Icon from '@pluralsight/ps-design-system-icon'
 
 import Dropdown from '../index.js'
 
-const Empty = () => null
+const Empty = React.forwardRef((props, forwardedRef) => <span />)
```
EdwardIrby, line 11: In our propTypes we have `{ menu: PropTypes.element.isRequired }` so I modified this to represent that and also portal really does require and element.

packages/dropdown/src/react/index.js
```diff
@@ -157,9 +158,33 @@ const Dropdown = React.forwardRef((props, forwardedRef) => {
...
+      window.addEventListener('resize', requestAnimationFrame, {
+        passive: true
+      })
+      window.addEventListener('scroll', clear, { passive: true })
+    }
+    return () => {
+      document.removeEventListener('click', handleClickOutsideMenu)
+      document.removeEventListener('contextmenu', handleClickOutsideMenu)
+      window.removeEventListener('resize', requestAnimationFrame)
+      window.removeEventListener('scroll', clear)
```
EdwardIrby, line 186: Update conditions on which we close the dropdown automatically. In addition to **click out side**, we now have:
 - **right click outside**
- **scroll event**
- **resize event**

packages/dropdown/src/react/index.js
```diff
@@ -319,3 +354,16 @@ export const appearances = vars.appearances
...
+    typeof window !== 'undefined' &&
+    typeof window.document !== 'undefined' &&
+    typeof window.document.createElement !== 'undefined'
+  )
+}
+
+function createUniversalPortal() {
+  if (!canUseDOM()) return null
+  return ReactDOM.createPortal(...arguments)
+}
```
EdwardIrby, line 369: These two function were copied over from Position. Wondering if we should add to utils package?

packages/dropdown/src/react/index.js
danethurber: i know you mentioned in standup you felt this hook was too large. one simple change could be separating it into multiple hooks. that would also make the code more readable and the hooks would be reusable. here is some code from the flow extensions repo that illustrates how you could organize the hooks into smaller and more reusable chunks

```
import { canUseEventListeners, canUseDOM } from 'exenv';
import { useEffect, useRef } from 'react';

const ESCAPE_KEY = 27;

export function useGlobalEventListener<EventName extends keyof WindowEventMap>(name: EventName, fn: (evt: WindowEventMap[EventName]) => unknown): void {
    const handlerRef = useRef<typeof fn>();

    useEffect(() => {
        handlerRef.current = fn;
    }, [fn]);

    useEffect(() => {
        if (!canUseEventListeners || !canUseDOM) return;

        const handler = handlerRef.current!;

        window.addEventListener(name, handler, true);

        return () => {
            window.removeEventListener(name, handler, true);
        };
    }, [name]);
}

export function useOnDocumentClick(handler: (evt: any) => unknown) {
    useGlobalEventListener('mousedown', handler);
    useGlobalEventListener('touchstart', handler);
}

export function useOnEscape(handler: (evt: KeyboardEvent) => unknown) {
    function handleKeyDown(evt: KeyboardEvent) {
        if (evt.keyCode !== ESCAPE_KEY) return;
        handler(evt);
    }

    useGlobalEventListener('keydown', handler);
}
```

packages/dropdown/src/react/index.js
danethurber: i'd propose using [exenv](https://www.npmjs.com/package/exenv) for the first function. and i do think the portal code could be moved to the utils package

packages/dropdown/src/react/__stories__/index.story.js
danethurber: i'm surprised you hit an error. i've alway though `proptypes.element` just needed any react element so i would have thought a component that returned null would still be valid

danethurber: did the functionality of proptypes change and we upgrade the package recently and that's where the error is coming from?

packages/dropdown/src/react/index.js
```diff
@@ -157,9 +158,33 @@ const Dropdown = React.forwardRef((props, forwardedRef) => {
...
+      return currentAnimationFrame
+    }
+    let timeout = null
+    const clear = () => {
+      clearTimeout(timeout)
+      timeout = setTimeout(() => setOpen(false), 50)
+    }
+    if (isOpen) {
+      document.addEventListener('click', handleClickOutsideMenu)
+      document.addEventListener('contextmenu', handleClickOutsideMenu)
```
danethurber, line 176: i don't think we've closed on right click before and i'm not sure if that's correct. my rule of thumb has been to close on events that change focus(and the focus should be returned to the prev active element). a right click doesn't change focus so i'm unsure that this change is correct

EdwardIrby: I just cloned the interaction behavior of a native select element here. It's probably not necessary.

packages/dropdown/src/react/index.js
EdwardIrby: I'll update the PR to use this pattern

packages/dropdown/src/react/__stories__/index.story.js
EdwardIrby: I think it will take a react element that return null but createPortal will not.

packages/dropdown/src/react/index.js
EdwardIrby: Going to pass on this suggestion. Their is a conditional requirement for these functions making the pattern above more complex in it's implementation than I think is necessary.

PR #996, opened 2020-05-05
refactor(button): use low color on flat button text/icon to match figma

sweeting:
Changed label & icon colors in flat buttons to "low" to match Figma.

Also made a tweak to the active state animation which uses `scale` instead of `translateY`, which I think is just slightly nicer. :)

### DIFF
--- a/packages/banner/src/react/__specs__/__snapshots__/storyshots.spec.js.snap
+++ b/packages/banner/src/react/__specs__/__snapshots__/storyshots.spec.js.snap
@@ -12,8 +12,8 @@ exports[`Storyshots Button blue 1`] = `
     >
       this is the text with a 
       <button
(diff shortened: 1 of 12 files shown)


sweeting:
@jaketrent I merged in your updated snapshots, but I'm not sure why this is still failing and what I need to do before I can merge this.

jaketrent:
ah, yes, this problem...

If you click the Travis CI "Details" link above the merge button, you can see what's failing.  It's in the "All checks have failed" section.

Link goes here: https://travis-ci.org/github/pluralsight/design-system/jobs/683900264

It's showing that there are snapshots for other packages failing.  These other packages depend on button.  They have button in their snapshots, so the changes in the CSS has affected these as well.

I'll update them and commit, then hopefully it'll go green.

PR #998, opened 2020-05-06
CircularProgress/Tab: Run typechecking as part of npm test

jhewlett:
### What You're Solving

- There was a regression in a `.d.ts` file (https://github.com/pluralsight/design-system/pull/997) that made it through because we aren't yet running `tsc` to check types as part of CI. 

### Design Decisions

- I considered adding a separate `typecheck` script to the `package.json`, but ended up adding it to `test` because that's an easy way to hook into the CI pipeline, and we had discussed doing it this way previously. I verified that `test` and `test:watch` still have the correct behavior.

### How to Verify

- Tests run (presumably without flagging any type errors, though you could introduce one to verify the new step)


package.json
```diff
@@ -79,6 +79,8 @@
     "@testing-library/jest-dom": "^5.1.0",
     "@testing-library/react": "^9.4.0",
     "@testing-library/react-hooks": "^3.2.1",
+    "@types/react": "^16.8.6",
+    "@types/react-dom": "^16.8.6",
```
jhewlett, line 83: I had to add a dev dependency on `@types/react` to get typechecking to work in our `.tsx` stories since React doesn't bundle their own types

packages/circularprogress/package.json
```diff
@@ -16,7 +16,7 @@
     "build:watch": "npm run build:js -- --watch",
     "prepublish": "npm run build",
     "storybook": "start-storybook -p 6006",
-    "test": "jest",
+    "test": "tsc --noEmit --project ./tsconfig.json && jest",
```
jhewlett, line 19: `--noEmit` means "don't emit JS files, just typecheck"

jaketrent:
Conceptually, I like that test has `tsc` as a part of the verification.

@danethurber or @EdwardIrby  Any opinions on the CI implications?

PR #1001, opened 2020-05-07
Refactor/avatar

sweeting:
• Show only first initial in fallback
• Make type style size/weight changes
• Update color set
• Fix issue that prevented all possible colors from being used
• Update docs to use a single example displaying all sizes and fallback states (this could probably be done in a much more efficient way than I did it.)


packages/avatar/src/js/index.js
```diff
@@ -1,27 +1,17 @@
...
 
 export const getColorByName = name => {
-  const defaultColorMapKey = 'A'
-  const stableRandLetter =
-    sha1(name)
-      .toUpperCase()
-      .replace(/[^A-Z]/g, '')[0] || defaultColorMapKey
-  return colorByLetter[stableRandLetter]
+  const stableRandNumber = parseInt(sha1(name).replace(/[^0-9]/g, '')[0], 10)
+  // console.log(stableRandNumber,sha1(name))
```
jaketrent, line 13: nit(blocker): please remove commented code.

packages/site/pages/components/avatar.js
```diff
@@ -55,28 +57,82 @@ export default _ => (
...
+          gap: core.layout.spacingXXLarge,
+          gridTemplateColumns: `auto auto auto auto auto`,
+          alignItems: 'center'
+        }}
+        outputChildStyle={{ margin: 0 }}
         codes={[
-          `<Avatar name="Alan Turing" />`,
-          `<Avatar name="Grace Hopper" />`,
-          `<Avatar name="Tim Berners-Lee" />`
+          `<Avatar size="xLarge" name="Assis Silva" src="https://secure.gravatar.com/avatar/0f792a763ebf08411c7f566079e4adc7?s=400"  />`,
```
jaketrent, line 81: opinion(non-blocker): Showing the permutations seems like a fine idea to me.  Just personally, it feels like too much.  Also, I have to scroll 2 pages before I can see what code generates the visuals, and it's hard to pair code to visuals.

sweeting: Good feedback. I adjusted the examples in the docs.

PR #1002, opened 2020-05-08
fix(dropdown): position needs to be caluclated each time dropdown opens

EdwardIrby:
https://pluralsight.slack.com/archives/C70HQMTUM/p1588891991012500


jaketrent:
Looks like you switched this to WIP after you asked for a review.  I just noticed that.

My initial test in storybook seems to fix the problem in the scenario here:
- open dropdown
- scroll
- open again

master had problem; branch does not

EdwardIrby:
Moved eventListeners to Actionmenu. Scroll listener is not working in storybook but is working on next.js site.

resolves  #1004

packages/actionmenu/src/react/index.js
```diff
@@ -32,7 +30,8 @@ ActionMenu.propTypes = {
 ActionMenu.defaultProps = {
   isKeyboarding: false,
   origin: vars.origins.topLeft,
-  shouldFocusOnMount: true
+  shouldFocusOnMount: true,
+  onClose: () => {}
```
EdwardIrby, line 34: @jaketrent this was overriding the defaultProps I set in menu.js so I moved the onClose up to this indes.js file

PR #1010, opened 2020-05-13
Fix/radio controlled uncontrolled

### LINKED ISSUE #1003: radio: use new props.value to update selected radio
### Expected behavior

Can use value props updates to control the component.

### Actual behavior

https://pluralsight.slack.com/archives/C70HPSJPP/p1588977099121600
https://codesandbox.io/s/ps-ds-radioexample-vlizo?file=/src/App.js

### Steps to repro

Change Radio.Group value prop twice.  First time will set.  Second time will not.

### LINKED ISSUE #1005: radio: selecting a radio button temporarily focuses the 0th item
# Bug

### Expected behavior

1. Click an item
2. Focus the clicked item without temp focus on 0th radio

### Actual behavior

1. Click an item at index 1+
2. Focus the 0th item temporarily
3. Focus the clicked item

EdwardIrby:
BREAKING CHANGE
- **onSelect** renamed **onChange**
- **checked** prop removed from `<Radio.button/>` use **on handlers** to grab or use `<Radio.Group />` as a Controlled component
- `<Radio.Button />` must be used with `<Radio.Group/>`
- **name** prop required on `<Radio.Group />`
- **onFocus, onKeyDown** removed from `<Radio.Group />` if you need focuse or keydown handlers on `<RadioButton />` please attach directly

resolves #1005 
resolves #1003 

Met with Amber to verify tab and key up and down interactions


packages/radio/src/react/button.js
```diff
@@ -1,24 +1,24 @@
-import { compose, css } from 'glamor'
+import { compose, css, checked } from 'glamor'
```
danethurber, line 1: `checked` coming from glamor is probably not what you want. it's always truthy so all buttons are shown as checked

packages/radio/src/react/button.js
```diff
@@ -28,86 +28,80 @@ const styles = {
...
 
-const Button = React.forwardRef((props, forwardedRef) => {
-  const themeName = useTheme()
+const isChecked = (a, b) => a === b
 
+const Button = React.forwardRef(({ value, label, ...props }, forwardedRef) => {
+  const themeName = useTheme()
+  const { checkedValue, onSelect, disabled, error, name } = useRadioContext()
   const ref = React.useRef()
   React.useImperativeHandle(forwardedRef, () => ref.current)
```
danethurber, line 38: you probably wanted to calculate `checked` doing something like

```
const checked = checkedValue === value
```

packages/radio/src/react/button.js
```diff
@@ -1,24 +1,24 @@
...
 import { useTheme } from '@pluralsight/ps-design-system-theme'
-
+import { combineFns } from '@pluralsight/ps-design-system-util'
+import { useRadioContext } from './context.js'
 import stylesheet from '../css/index.js'
 
 const styles = {
   button: () => css(stylesheet['.psds-radio-button']),
-  circle: (themeName, { checked }) =>
+  circle: (themeName, checked) =>
```
danethurber, line 14: we've been trying to standardize on the params in these style calls. it usually looks like this

`(themeName, objectContainingStateOrProps) => `

please update all functions to match the pattern

packages/radio/src/react/button.js
danethurber: oh...looks like you have an isChecked function. so i suggest using that once here to have a single variable with the boolean. and then making sure that's used everywhere - right now the functionality is broken because of the `checked` coming from glamor

```
const checked = isChecked(value, checkedValue)
```

packages/radio/src/react/group.js
```diff
@@ -25,117 +26,65 @@ const styles = {
...
 Group.propTypes = {
   children: PropTypes.node, // TODO: children only Radio.Button
   disabled: PropTypes.bool,
   error: PropTypes.bool,
   label: PropTypes.node,
-  name: PropTypes.string,
-  onFocus: PropTypes.func,
-  onKeyDown: PropTypes.func,
+  name: PropTypes.string.isRequired,
   onSelect: PropTypes.func,
```
danethurber, line 88: as long as we're doing a breaking change, let's use `onChange` instead of `onSelect` so we match up with the react standard props

packages/radio/src/react/button.js
EdwardIrby: why?

packages/radio/src/react/button.js
EdwardIrby: That isn't taking props. I mean i can p[ass it a object and then let is access the checked key/value from it but that seems inefficient.  Also every time we do that it's inefficient if we've already desctrutured the value out form props assuming that is where it is coming from.

packages/radio/src/react/group.js
EdwardIrby: technically I'm using onClick happy to change it to that if we want to be really accurate.

EdwardIrby: nevermind

packages/radio/src/react/button.js
danethurber: it is taking state which is the patterns we've been using in other places. i could also make the argument around refactor/extensibility around passing objects instead of literal fixed types that lead to prop creep. i'm actually ok if you want to disregard my suggestions and merge as is. eventually we will rethink css so all these patterns will change

PR #1017, opened 2020-05-21
Fix/caorusel transition tabbing

### LINKED ISSUE #1013: Determine why buttons are not tabbable on Carousel site example

EdwardIrby:
resolves #1013

### DIFF
--- a/packages/card/src/css/index.js
+++ b/packages/card/src/css/index.js
@@ -147,10 +147,12 @@ export default {
     transition: `opacity ${motion.speedNormal}`,
     pointerEvents: 'none',
     opacity: 0,
+    visibility: 'visible',
     zIndex: 10,
 
     '> *': {
-      pointerEvents: 'all'
+      pointerEvents: 'auto',
+      opacity: 1
     },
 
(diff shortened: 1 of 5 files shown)


EdwardIrby:
No longer using typescript switch to js and simply adding type def files for better inference.

EdwardIrby:
> Tests good. Storybook looks good. You link to #1013, which I have reviewed, but there is little info there. Which "buttons" are untabbable is unknown to me. But now apparently they are tabbable and all the stories seem to indicate that.
> 
> One other observation. This tab scenario:
> 
>     1. Tab to next page button, press enter, next page advances
> 
>     2. Shift-tab to prev page button, press enter, back to prev page
> 
>     3. Shift-tab again ends up on prev page button.
> 
> 
> I would have expected:
> 3. Shift-tab ends up on the last focusable element on the last item in the current page.
> 
> Not sure if this is correct, it's just something unexpected that I ran into.

It's expected. Once a user has begun navigating with the page buttons focus will always go to the first item on the current page once the transition completes. If there is a prev page button available shift-tab will move focus there. If not it will move to the preceding tabable item on the page.

PR #1019, opened 2020-05-26
fix(tooltip): accept child node proptypes

alexzywiak:
### What You're Solving
Updates Tooltip component to not throw prop types error when rendering a valid React node.

### Design Decisions

Implementation allows any valid node to be used as a child.  Updated propTypes to reflect that.

### How to Verify

In development, render a tool tip with a react node as a child.  No propTypes error should be visible in the console.


jaketrent:
Not sure your what your page design is.  Generally, we recommend you use Dialog for content that is more than a string. The UI is similar to Tooltip, but it is intended for more than strings:

https://design-system.pluralsight.com/components/dialog

Here's how we're differentiating them today.  We plan to further differentiate them in the future, but it's not a current priority.

Dialog:

> The purpose of a Dialog is to provide actionable messaging and may appear contextually or as a modal. The Dialog adapts to various amounts and types of content. For contextual, non-actionable messaging, consider the Tooltip instead.

Tooltip: 

> The purpose of a tooltip is to provide context and explain the function of a user interface element or feature. The content of a tooltip is limited to styled text. If more customization is necessary, consider the Dialog component which builds on the patterns of the tooltip.

Please check it out and see if it fits your needs. Closing for now. If there's something I'm missing, let's reopen and keep chatting about this.

alexzywiak:
@jaketrent I was using formatted text inside of the tooltip through JSX.  Still feels tooltippy to me as it's just display text without any actions, but understand the distinction you have between dialog and tooltip.  And can make the switch to better reflect the component's intent.

Implementation looks like this:

Screen Shot 2020-05-26 at 4 05 37 PM

jakemwood:
@jaketrent I was *just* about to open a PR to resolve this, as well, to clear up the propTypes errors. I believe this PR should be re-considered. As you said, according to the documentation:

```
The purpose of a tooltip is to provide context and explain the function of a user interface element or feature. The content of a tooltip is limited to styled text. If more customization is necessary, consider the Dialog component which builds on the patterns of the tooltip.
```

My own usage of the component is very similar to Alex's. Since this component simply adds additional context to the page, and is not actionable, it certainly seems `Tooltip` is the appropriate component to use. In fact, the `Dialog` documentation explicitly discourages what we are trying to do:

```
The purpose of a Dialog is to provide actionable messaging and may appear contextually or as a modal.
```

We are not providing actionable messaging *or* a modal, thus leading me to believe a Tooltip is the correct component to use.

Any clarity you can provide would be helpful.

jaketrent:
Hey @jakemwood , yeah, I like the reasoning you have here.  And I think we need to do a better job at differentiating these two components.

For now, we should probably just loosen this up and then readdress when we can do that differentiation justice.

Because we've stayed with string-only, let me double check with the team on a merge.  If we're good, then I'll merge it, and we'll kick this can a bit down the road.  

Thanks for re-engaging on an issue and not assuming it was "solved".

PR #1020, opened 2020-05-28
build: update package.json

EdwardIrby:


jaketrent:
@EdwardIrby Where are we on this?  Are you waiting for something from Dane or myself? Thanks.

PR #1023, opened 2020-06-09
Fix/list style

### LINKED ISSUE #1021: Text.List: bulleted lists can not be nested in numbered lists
### Expected behavior

When nesting a list within a list it is expected that the nested list could be either `numbered` or `bulleted` and not dictated by its parent list.

This block, using html elements, renders the list with nesting as expected.
```html
<ol>
  <li>One</li>
  <li>Two</li>
  <ul>
    <li>first bullet</li>
    <li>second bullet</li>
  </ul>
  <li>Three</li>
</ol>
```

1. One
2. Two
- first bullet
- second bullet 
3. Three

### Actual behavior

When nesting a `bulleted` list within a `numbered` list the nested list should have bullet markers instead of numbered markers

This code

image

produces this

image

while nesting a numbered list within a bulleted list does this

image

image

### Steps to repro

1. Using Text.List from the design system nest a bulleted list inside a numbered list.

### Related packages

-text@16.0.4

### Environment

- OS - MacOS 10.15.4
- Browser version - Version 83.0.4103.61 (Official Build) (64-bit)

EdwardIrby:
resolve #1021


EdwardIrby:
@sweeting do we have designs for nested bulleted/numbered list or should we just allow the consumer of this component to style the layout and list-style-type as as they need?

sweeting:
@EdwardIrby I think nested lists should be able to be styled by the consumer of the component as needed.

EdwardIrby:
> the changes to the List component look fine. i'm wondering why there are also changes to the Carousel and Radio components?

Their just new storybook stories. I started on one ticket figured out their wasn't a bug and then felt the stories should stay regardless so I committed them and moved onto the next.

PR #1025, opened 2020-06-12
fix(table): update column header click event arg order and aria-sort

### LINKED ISSUE #1024: Table: a11y columnheaders
the current column headers when rendered as a button fail a11y checks because a button can't have a role of columnheader

also. we should change the event argument order of the `onClick` prop so that the click event is the first argument

related thread -> https://pluralsight.slack.com/archives/C70HPSJPP/p1591279249202900

EdwardIrby:
resolves #1024

### DIFF
--- a/packages/table/src/react/index.js
+++ b/packages/table/src/react/index.js
@@ -40,6 +40,16 @@ const styles = {
       onClick && css(stylesheet[`.${label}--onclick.psds-theme--${themeName}`])
     )
   },
+  columnHeaderButton: () =>
+    css(stylesheet['.psds-table__column-header__button']),
+  columnHeaderButtonInner: props => {
+    const { align } = props
+    const label = 'psds-table__column-header'
+    return compose(
+      css(stylesheet[`.${label}__button-inner`]),
+      css(stylesheet[`.${label}--align-${align}`])
+    )
+  },
   columnHeaderIcon: _ => css(stylesheet['.psds-table__column-header__icon']),
   row: (themeName, { _tableHasDrawers }) => {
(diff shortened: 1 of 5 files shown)


packages/table/src/react/index.js
```diff
@@ -100,27 +110,44 @@ function ColumnHeader(props) {
...
-  return (
-    <Tag
-      role="columnheader"
-      {...styles.columnHeader(themeName, props, { active })}
-      {...filterReactProps(props, { tagName: Tag })}
-      onClick={handleClick}
-      style={style}
-      aria-sort={
+  const children = props.onClick ? (
+    <button {...styles.columnHeaderButton()} onClick={handleClick}>
```
EdwardIrby, line 119: Not sure if we shouldn't just leave this on the TableColoumn header div

packages/table/src/react/index.js
```diff
@@ -100,27 +110,44 @@ function ColumnHeader(props) {
...
+        {props.sort && getSortIcon(props)}
+      </div>
+    </button>
+  ) : (
+    <>
+      {props.children}
+      {props.sort && getSortIcon(props)}
+    </>
+  )
+  const sort = {
```
EdwardIrby, line 131: Looked at dequeuniversity and they weren't event showing this attribute if it wasn't sorted. I went halfway and basically am not applying it unless the user has set sort of some sort on the column.

packages/table/src/react/index.js
danethurber: do you mean having the onClick event on the columnheader div. that seems fine to me

PR #1029, opened 2020-06-19
Nav Item

### LINKED ISSUE #1032: navitem: new component
A nav component. From Prism topnav. Check with upcoming design changes: https://www.figma.com/file/sqIeRqEuTdKZSfyTj3Qvyo/Nav-System-Components?node-id=527%3A342

Perhaps wait so that we can punt. There are isolated non-react teams today. But product orgs will have to share code, hopefully in a health way.  Perhaps this is a problem we won't have to solve for. Procrastinating...

~As first component, need to figure out cross-tech story.  What about non-React framework support?~

jaketrent:
Waiting on final designs. Transferring Prism visuals.

- [x] ~Theme support~ ...There is none.
- [x] ~Doc page for component~ ... Keeping this on the DL for now.
- [x] Adjust to new design

Resolves #1032 

Screen Shot 2020-06-24 at 2 08 54 PM
Screen Shot 2020-06-24 at 2 09 12 PM


sweeting:
These look great @jaketrent. For vertical, there is no label only version. Can we hide that in storybook?

I'm making a note for myself to add some CSS transitions in a future polish refactor.

packages/navitem/src/react/common.js
```diff
@@ -0,0 +1,66 @@
...
+          ...styles.button(props),
+          ...filterReactProps(props, { tagName }),
+          ref
+        },
+        props.children
+      )}
+    </Halo>
+  )
+})
+Button.displayName = 'NavItemButton'
```
danethurber, line 51:
```suggestion
Button.displayName = 'NavItem.Button'
```

PR #1035, opened 2020-06-26
[experimental] gatsby/playroom for docs 

danethurber:
an example/poc of using gatsby+playroom for docs

there are some codeblock and preview components from the apollo docs plugin but we might want to roll our own to make them better. i also setup a playroom instance to show that we can open a new playroom page directly from a codeblock

## setup
you'll need to create an `.env.development` file and we don't have an algolia account but you can use apollo's to get search results

```
ALGOLIA_API_KEY=768e823959d35bbd51e4b2439be13fb7
ALGOLIA_INDEX_NAME=apollodata
```


jaketrent:
Dane's findings:

- Gatsby, playroom could be used for docs
- Docsearch integratable with storybook or gatsby or anything

jaketrent:
Closed as a spike.  Use findings in future PR.

PR #1043, opened 2020-06-29
NavBar, NavBrand, NavItem, NavUser

### LINKED ISSUE #1034: NavBar: new component
Current thinking, perhaps mostly a container element with the content coming from product:

ContainerWithGuidelines

- Make the menu button an optional/guideline thing? What's its purpose?  Collapsing topnav into mobile menu? Or toggling sidenav?  Or both?
- Based on the Youtube behavior
- [ ] Make it clear what the new interaction pattern is: toggling and locking.
- [ ] Decide if the semantics are right? (eg, "Brand", "Nav", "Utility", "User")
- [ ] If we leave the "User" section to a Profile product team, what about the previously-in-footer links that we were planning to house in the profile menu? (We won't control the existence of a profile menu.). This is problematic because we were hoping to house a bunch of non-profile stuff there.  We could let the User prop take an element which we build an onClick handler onto and pop a menu.  Then, we could ensure the menu existence while keeping the UI in the hands of a Profile product team.

jaketrent:
Highlights
- New NavBar comp 
- NavItem supports icon only
- New NavBrand comp
- New NavUser comp

Details
- Decided to rename NavBar#nav to NavBar#items.  Is that the right semantic name?  (It's full of NavItems as the main use case for that component.  Everything in the NavBar was nav, so that name didn't make sense.)
- Toggling/locking pattern question(s) are bypassed for the moment.  This component doesn't yet answer that question, maybe never will.  Perhaps we leave it to guidelines and let the product nav codebase contain that state to support the interaction guideline.
- NavBrand and NavUser are likely to either be replaced someday or taken over by another team and removed from the general DS.  Their separateness and composability here should assist in this longterm migration.

Questions
- [x] Biggest semantic question on naming remains "utility".  We called those things "Utility buttons" when we first put them there.  Not sure if that's the concept we want to use going forward because no one except me seems to use that term any more.

Resolves #1034


packages/navuser/src/css/index.js
```diff
@@ -0,0 +1,45 @@
...
+  '.psds-navuser': {
+    display: 'inline-flex',
+    alignItems: 'center',
+    padding: `${layout.spacingXXSmall} ${layout.spacingXSmall}`,
+    textDecoration: 'none',
+    border: 'none',
+    background: 'none',
+    borderRadius: '2px'
+  },
+  '.psds-navuser--clickable': {
```
danethurber, line 19:
```suggestion
  '.psds-navuser--clickable': {
    cursor: 'pointer',
    
```

packages/navuser/src/react/index.js
```diff
@@ -0,0 +1,72 @@
...
+        </>
+      )}
+    </Halo>
+  )
+})
+NavUser.displayName = 'NavUser'
+NavUser.propTypes = {
+  href: PropTypes.string,
+  onClick: PropTypes.func,
+  planName: PropTypes.string,
```
danethurber, line 52: is this always going to be the plan name. is it more like a secondary text or subtitle area

packages/navuser/src/css/index.js
```diff
@@ -0,0 +1,45 @@
...
+      background: transparentize(0.85, colorsBackgroundUtility.base)
+    },
+    '&:focus': {
+      outline: 'none'
+    },
+    '&:active': {
+      background: transparentize(0.75, colorsBackgroundUtility.base)
+    }
+  },
+  '.psds-navuser__plan-name': {
```
danethurber, line 30: wondering about an upperlimit on text. should we cap/truncate

cc @sweeting

jaketrent: We are going to do a text overflow on a maxwidth on the container.  I need to add that.

packages/navbar/src/react/index.js
```diff
@@ -0,0 +1,47 @@
...
+        </div>
+      )}
+      <div {...styles.brand()}>{props.brand}</div>
+      <div {...styles.items()}>{props.items}</div>
+      <div {...styles.utility()}>{props.utility}</div>
+      <div {...styles.user()}>{props.user}</div>
+    </div>
+  )
+})
+
```
EdwardIrby, line 38: Needs displayName

packages/navbar/src/css/index.js
```diff
@@ -0,0 +1,26 @@
+import { layout } from '@pluralsight/ps-design-system-core'
+
+export default {
+  '.psds-navbar': {
```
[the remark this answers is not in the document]
sweeting: Yes, it should be colorsBackgroundDark[2]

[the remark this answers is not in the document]
sweeting: There's also a 1px bottom border `colorsBackgroundDark[3]` and a box shadow `box-shadow: rgba(0, 0, 0, 0.5) 0px 0 8px 0px;`

jaketrent:
Not sure why travis tests still unhealthy.  Local branch has fully-merged master and all tests pass:

```
Test Suites: 106 passed, 106 total
Tests:       1998 passed, 1998 total
Snapshots:   980 passed, 980 total
Time:        35.913s, estimated 45s
Ran all test suites.
```

packages/navuser/src/react/index.js
jaketrent: I made a call here.  Still not sure it's the right one.  

I called it planName because:

- Most foundationally, I started thinking about the NavUser component as a "temporary" DS component.  It'll help product teams get over the transition to product navs.  Then it'll be owned by someone else, likely the profile team, at the point that team has an opinion on how it should evolve.  At that point, if they believe that the plan name should not appear there or the space should be repurposed, they can. It'll probably be a breaking change to the api, and life will go on. We won't own it beyond the time it's `planName`.
- User `name` is also a semantic prop that I think is nice.  Easier to know what goes there.  
- Suggesting semantics through the API will help with consistent use.

packages/navbrand/src/css/index.js
```diff
@@ -0,0 +1,51 @@
...
+
+import { transparentize } from '@pluralsight/ps-design-system-util'
+
+export default {
+  '.psds-navbrand': {
+    display: 'flex',
+    alignItems: 'center',
+    padding: layout.spacingXSmall
+  },
+  // TODO: handle hover and focus
```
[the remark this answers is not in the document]
jaketrent: nope, I need to rm.

packages/navbrand/src/react/index.js
```diff
@@ -0,0 +1,85 @@
...
+
+function Logo(props) {
+  return <div {...styles.logo(props)}>{props.logo}</div>
+}
+Logo.displayName = 'NavBrand.Logo'
+Logo.propTypes = {
+  logo: PropTypes.node
+}
+
+function Wordmark(props) {
```
EdwardIrby, line 79: Is this for a11y if so what about using the ScreenReaderOnly component?

packages/navuser/src/react/index.js
jaketrent: cc @sweeting ... I see that you updated the figma to read "primary" and "secondary" again this morning.  Will you please weigh in on your current thinking and respond to my reasoning above?

packages/navbrand/src/react/index.js
jaketrent: The purpose of Wordmark component is not a11y.  It's just naming some part of the subcomponent.

EdwardIrby:
> Not sure why travis tests still unhealthy. Local branch has fully-merged master and all tests pass:
> 
> ```
> Test Suites: 106 passed, 106 total
> Tests:       1998 passed, 1998 total
> Snapshots:   980 passed, 980 total
> Time:        35.913s, estimated 45s
> Ran all test suites.
> ```

When I pulled this I had a local failing snapshot after running clean & bootstrap

EdwardIrby:
Are we not including these components in the pluralsight design system site?

jaketrent:
@EdwardIrby hmm, maybe I need to clean/bootstrap. Thanks!

@EdwardIrby correct, for now we're not adding these to the DS site.  We're waiting until:

1) The design for the nav docs is complete.
2) All known nav components are implemented and known to work well together.  If we end up doing a ref implementation ourselves, creating the docs could even come after that point.

packages/navuser/src/react/index.js
sweeting: @jaketrent Even if it is temporary, I like the primary and secondary naming for those lines of text.

packages/navbrand/src/react/index.js
```diff
@@ -0,0 +1,85 @@
...
+      {React.createElement(
+        tagName,
+        {
+          ...styles.navBrand(props),
+          ...filterReactProps(props, { tagName }),
+          ref
+        },
+        <>
+          <Logo {...props} />
+          <Wordmark {...props} />
```
sweeting, line 54: Even though we don't currently have the heaviest weight of Gotham, I think we should use text here. The weight issue can be addressed and using text will help keep the usage more consistent and accessibile.

```
color: #fff;
font-weight: 900;
font-size: 20px;
letter-spacing: -.03em;
line-height: 24px;
```

packages/navbrand/src/css/index.js
```diff
@@ -0,0 +1,51 @@
+import {
+  colorsBackgroundUtility,
+  layout
+} from '@pluralsight/ps-design-system-core'
+
+import { transparentize } from '@pluralsight/ps-design-system-util'
+
+export default {
+  '.psds-navbrand': {
```
sweeting, line 9: This element needs `border-radius: 2px` and height should always be 40px.

packages/navuser/src/react/index.js
jaketrent: @sweeting Why do you like "primary" and "secondary" for naming those properties?

packages/navbrand/src/css/index.js
jaketrent: done

packages/navbrand/src/react/index.js
jaketrent: done

packages/navuser/src/react/index.js
jaketrent: ^ bump.  @sweeting I'm still interested in continuing this ^ discussion. I'm interested in your reasoning on the naming preference. I believe this is the last point of discussion around the current PR before merge.

sweeting: @jaketrent I was just thinking it might be better to be less specific. `name` is probably fine since that's likely all it would be used for. `planName` may be more fraught since sometimes it's referred to internally and externally both as a "plan" and "account". Maybe we just rename `planName` to `meta`. That way it could be the the name of a plan or account or even used for an email address if that's what they decide to do.

Issue #1036, opened 2020-06-29
Autoposition utility & Tooltip refactor


EdwardIrby:
In regards to #932. The solution was to use BelowRight instead of Below because using below would cause the window width to change if the tooltip is wider than a target flexed to the right of the screen. After making this prop change the dev simply need to change the tip position to topRight. 

However this illustrates that a developer must be aware of where a target is on the screen in various context in which this position might change. A solution to this manual position management would be to allow portal like elements such as tooltip or dropdown to auto-position. 

To that effect I propose adding some auto-position utilities/hooks to the design system that can be used with these portal like elements.

sweeting:
Can it just work like this? Position can be specified, but automatically handles viewport. https://codesandbox.io/s/6hc12

jaketrent:
Triage:

- Remains: Still specify by design where the tooltip/other thing should go
- Add: Avoid clipping by shifting automatically

EdwardIrby:
Cl

PR #1046, opened 2020-07-01
Feat/autoposition

### LINKED ISSUE #932: position: Position/Tooltip miscalculation
> I think there's some kind of bug in the [Position|Tooltip] element (not sure which, but thinking Position). If you got to https://app-stage.pluralsight.com/plan-analytics/1800contacts/skills the info icon as shown below should show a tooltip. It works when the browser window has plenty of space to the right. When the browser window width shrinks, it takes a few mouse overs for the position to get calculated correctly. If you start out with a smaller browser width, the tip does not even show at all until you've moused in and out a lot while something is calibrating position and size and eventually the tip starts to show up, narrowing in on the correct position. (edited) 

cc https://pluralsight.slack.com/archives/C70HPSJPP/p1584550569101100

EdwardIrby:
resolves #932 #1036

This does not get rid of the tip. @sweeting should we do that in this PR https://pluralsight.slack.com/archives/C5CMEKH5Z/p1593465151076300

### DIFF
--- a/packages/position/src/js/__specs__/index.spec.js
+++ b/packages/position/src/js/__specs__/index.spec.js
@@ -26,7 +26,7 @@ describe('#below', () => {
     expect(below(targetEl).styleFor(el)).toEqual({
       position: 'absolute',
       top: '12px',
-      left: '-50px'
+      left: '0px'
     })
   })
 
@@ -185,7 +185,7 @@ describe('#belowRight', () => {
     expect(belowRight(targetEl).styleFor(el)).toEqual({
       position: 'absolute',
       top: '12px',
-      left: '-150px'
+      left: '0px'
     })
   })
 
@@ -202,7 +202,7 @@ describe('#belowRight', () => {
     expect(belowRight(targetEl).styleFor(el)).toEqual({
       position: 'absolute',
       top: '212px',
-      left: '-50px'
+      left: '0px'
     })
   })
 
@@ -286,7 +286,7 @@ describe('#above', () => {
     expect(above(targetEl).styleFor(el)).toEqual({
       position: 'absolute',
       top: '38px',
-      left: '-25px'
+      left: '0px'
     })
   })
 })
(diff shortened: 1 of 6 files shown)


packages/position/src/react/__stories__/index.story.js
```diff
@@ -232,3 +232,98 @@ function JsStory({ positionType, ...rest }) {
 JsStory.propTypes = {
   positionType: PropTypes.oneOf(Object.keys(positionFns)).isRequired
 }
+
+const edgeCaseSotries = storiesOf('Position Edge', module)
```
danethurber, line 236: typo 😉 `edgeCaseSotries `

packages/position/src/react/__stories__/index.story.js
```diff
@@ -232,3 +232,98 @@ function JsStory({ positionType, ...rest }) {
...
+    const [b2hover, setb2hover] = React.useState(false)
+    const [b3hover, setb3hover] = React.useState(false)
+    const [b4hover, setb4hover] = React.useState(false)
+
+    return (
+      <div
+        style={{
+          display: 'grid',
+          justifyItems: 'start',
+          grid: '1fr 1fr/ 1fr 1fr',
```
danethurber, line 251: what does the `/` do in this style?

packages/position/src/js/__specs__/index.spec.js
```diff
@@ -26,7 +26,7 @@ describe('#below', () => {
     expect(below(targetEl).styleFor(el)).toEqual({
       position: 'absolute',
       top: '12px',
-      left: '-50px'
+      left: '0px'
```
danethurber, line 29: i was hoping these snapshots wouldn't change. maybe we need to setup the window width/height for these as well

EdwardIrby: Actually the point was for these to change. There should never be time when the tooltip is offscreen.

packages/position/src/react/__stories__/index.story.js
EdwardIrby: It's required for the css prop grid. Grid is a shorthand rule and the / signifies the division between rows and columns when writing the rules.

PR #1049, opened 2020-07-03
fix(table): doubled up click handler nulling sort

EdwardIrby:
Resolves https://pluralsight.slack.com/archives/C70HPSJPP/p1593792781288200

### DIFF
--- a/packages/site/pages/components/table.js
+++ b/packages/site/pages/components/table.js
@@ -94,18 +94,20 @@ class InAppExample extends React.Component {
         <Theme name={props.themeName}>
           <Table>
             <Table.Row>
-              {state.fields.map(field => (
-                <Table.ColumnHeader
-                  align={field.align}
-                  key={field.key}
-                  onClick={evt => handleSortClick(field.key, evt)}
-                  sort={
-                    state.sortKey === field.key ? state.sortDirection : true
-                  }
-                >
-                  {field.name}
-                </Table.ColumnHeader>
-              ))}
+              {state.fields.map(field => {
+                return (
+                  <Table.ColumnHeader
+                    align={field.align}
+                    key={field.key}
+                    onClick={(_, dir) => handleSortClick(field.key, dir)}
+                    sort={
+                      state.sortKey === field.key ? state.sortDirection : true
+                    }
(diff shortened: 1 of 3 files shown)


packages/table/src/react/__specs__/__snapshots__/storyshots.spec.js.snap
```diff
@@ -867,8 +867,7 @@ exports[`Storyshots columnHeader sort asc w/ onClick 1`] = `
   >
     <div
       aria-sort="ascending"
-      data-css-1yecw7w=""
-      onClick={[Function]}
```
EdwardIrby, line 871: This was the cause of the bug we had click handler being passed by filterReactProps to the wrapping div and and the same one on the button.

packages/table/src/react/index.js
```diff
@@ -100,7 +100,7 @@ const getToggledSort = props =>
     [vars.sorts.desc]: vars.sorts.asc
   }[props.sort] || vars.sorts.asc)
 
-function ColumnHeader(props) {
+function ColumnHeader({ onClick, ...props }) {
```
EdwardIrby, line 103: destructured onClick so that it is not passed to filterReactProps

packages/site/pages/components/table.js
```diff
@@ -137,6 +139,94 @@ class InAppExample extends React.Component {
               : core.colorsBackgroundDark[2]};
           }
         `}</style>
+        <Code
```
EdwardIrby, line 142: Added an in App code example

packages/site/pages/components/table.js
```diff
@@ -94,18 +94,20 @@ class InAppExample extends React.Component {
...
-                >
-                  {field.name}
-                </Table.ColumnHeader>
-              ))}
+              {state.fields.map(field => {
+                return (
+                  <Table.ColumnHeader
+                    align={field.align}
+                    key={field.key}
+                    onClick={(_, dir) => handleSortClick(field.key, dir)}
```
EdwardIrby, line 102: This wasn't updated to the 9.0 api where dir is the second arg

PR #1052, opened 2020-07-16
explore splitting out html

### LINKED ISSUE #1037: navitem: explore how to split out markup
Goal: Supply non-React projects/products with as much support through nav DS as possible.

Existing: They already have CSS separately built.  We can split out logic easily enough as needed.  There likely won't be a lot in the nav stuff, but who knows.

Current criticality: I think this would be good for the *story* of nav and DS overall.  Currently, it likely is not critical because both products will likely have a React implementation.  But, maybe not.  We still don't know how they'll plan to share it.

`htm` lib is a solution that has been pitched previously.

Is there a solution that would work in prism today (hyperapp)?

EdwardIrby:
# Findings

Three approaches were tried: 
1. **vanilla html** combined with the portal feature which is present in most modern ui libraries.
2. **template literals** using [htm](https://github.com/developit/htm)
3. **web components** for style and structure only, **no state or interaction. Uses tiny micro utils from previous open source work by @EdwardIrby.

## Vanilla HTML

In this approach we acknowledge the Nav and app shell are static and thus can be built into the index.html in a given PXT's Webpack/parcel/next/CRA setup. Then dynamic content can be rendered to the Nav using the portal like features available in modern UI libraries.  

### Pro's
Old school and simple

### Con's
- So old school most us might have forgot how to do this. 
- Varied project configs means lots of troubleshooting at onset of usage.

## Template literals

In this approach we use template literals native to javascript in combination with htm a library that allows UI libraries supporting the hyperscript pattern to share the same component code.

### Pro's
- At the atomic level is great for things like react, vue, preact, and hyperapp.

### Con's
- Can't figure out how to account for forwardRef as it changes the function signature
- Must put keys on fragment direct children and linter does not catch this or proptype errors.
- class and className are not universal across UI libraries

### Web component

In this approach we acknowledge that while we still use React non-scalar data and imperative apis on web components can not be used with react. (See https://custom-elements-everywhere.com/) We'd have to result to using useEffect and DOM API's to get expected behavior for interactivity. (See https://stenciljs.com/docs/react)  However if it's just structure and styling then we can split this out atomically using web components. 

## Pro's
- future proof
- Framework/Library agnostic

## Con's
- Needs polyfills for ie11
- may require rethinking our styling practices such as using css files again, a mix of css and glamour, or a build step that auto-injects styles and generates a root stylesheet for variables.

# Run
Navigate to `packages/universal-nav`
run `npm start`

---
resolves #1037

### DIFF
--- a/packages/universal-nav/src/htm/Navbar.js
+++ b/packages/universal-nav/src/htm/Navbar.js
@@ -1,0 +1,34 @@
+import { html } from './html.js'
+import PropTypes from 'prop-types'
+import styles from './styles.css'
+console.log(styles)
+
+/**
+ * TODO: how to work with forwardRef
+ * Do all frameworks use className?
+ *  */
+
+const Slots = ({ brand, items, user, utility, onMobileMenuClick }) =>
+  html`
+    <div className="${styles['psds-navbar__mobile-menu']}">
+      ${onMobileMenuClick}
+    </div>
+    <div class="${styles['psds-navbar__brand']}">${brand}</div>
+    <div class="${styles['psds-navbar__items']}">${items}</div>
+    <div class="${styles['psds-navbar__utility']}">${utility}</div>
+    <div class="${styles['psds-navbar__user']}">${user}</div>
+  `
+
+export const Navbar = props => html`
+  <nav className="${styles['psds-navbar']}">
+    <${Slots} ...${props} />
+  </nav>
+`
+
+Slots.propTypes = {
+  brand: PropTypes.node,
+  items: PropTypes.node,
+  onMobileMenuClick: PropTypes.func,
+  utility: PropTypes.node,
+  user: PropTypes.node
+}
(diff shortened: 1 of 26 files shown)


packages/universal-nav/src/htm/Navbar.js
```diff
@@ -0,0 +1,34 @@
+import { html } from './html.js'
+import PropTypes from 'prop-types'
+import styles from './styles.css'
```
EdwardIrby, line 3: Uses css-modules for the template literal example

packages/universal-nav/src/webcomponent/Nav.js
```diff
@@ -0,0 +1,35 @@
...
+    </div>
+  )
+}
+
+Filler.propTypes = {
+  children: PropTypes.node,
+  slot: PropTypes.string
+}
+export const Nav = () => (
+  <Navbar>
```
EdwardIrby, line 29: Got the react component to export properly from the registry without throwing any errors.

EdwardIrby: You could inline style, add classNames or data-attributes on this to manipulate internal styles in shadowDOM that are observing these attributes.

jaketrent:
Triage:  Closing for now.  These were great experiments to help us look at options. We aren't planning to take any of them forward at this point.  We'll wait for the point where we need to have extracted html for some concrete feature.

PR #1056, opened 2020-07-20
Decrease vertical NavItem height

jaketrent:
### What You're Solving

- Vert NavItem was too tall
- Horz NavItem was too short
- Both should come out to 56px in order to fit in and align vertically inside NavBar

### DIFF
--- a/packages/navitem/src/react/horz.js
+++ b/packages/navitem/src/react/horz.js
@@ -3,9 +3,11 @@ import { css } from 'glamor'
 import PropTypes from 'prop-types'
 import React from 'react'
 
+import { Bar, Button } from './common.js'
 import stylesheet from '../css/index.js'
 
 const styles = {
+  container: () => css(stylesheet['.psds-navitem__horz-container']),
   caret: () => css(stylesheet['.psds-navitem__horz-caret']),
   icon: props =>
     css(
@@ -20,6 +22,14 @@ const styles = {
     )
 }
 
+export function HorzContainer(props) {
+  return <span className={styles.container(props)}>{props.children}</span>
+}
+HorzContainer.displayName = 'NavItem.HorzContainer'
+HorzContainer.propTypes = {
+  children: PropTypes.node
+}
+
 export function HorzCaret(props) {
   return props.menu ? (
     <span className={styles.caret()}>
@@ -27,6 +37,7 @@ export function HorzCaret(props) {
     </span>
   ) : null
 }
+HorzCaret.displayName = 'NavItem.HorzCaret'
 HorzCaret.propTypes = {
   menu: PropTypes.element
 }
@@ -36,6 +47,7 @@ export function HorzIcon(props) {
     <span className={styles.icon(props)}>{props.icon}</span>
   ) : null
 }
+HorzIcon.displayName = 'NavItem.HorzIcon'
 HorzIcon.propTypes = {
   icon: PropTypes.element
 }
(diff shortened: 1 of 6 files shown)


packages/navitem/src/react/index.js
```diff
@@ -2,22 +2,14 @@ import React from 'react'
...
-      <Button {...props} ref={forwardedRef}>
-        {props.alignment === 'vertical' ? (
-          <VertLayout {...props} />
-        ) : (
-          <HorzLayout {...props} />
-        )}
-      </Button>
-      <Bar {...props} />
-    </Container>
+  return props.alignment === 'vertical' ? (
```
jaketrent, line 9: Now Container is not common, but vert/horz-specific.  Moved this whole "layout" into the *Layout functions.

packages/navitem/src/react/horz.js
```diff
@@ -20,13 +22,22 @@ const styles = {
...
+}
+
 export function HorzCaret(props) {
   return props.menu ? (
     <span className={styles.caret()}>
       <CaretDownIcon size={CaretDownIcon.sizes.small} />
     </span>
   ) : null
 }
+HorzCaret.displayName = 'NavItem.HorzCaret'
```
jaketrent, line 40: Just a bunch o displayNames that were missing too.

packages/navitem/src/react/index.js
```diff
@@ -2,22 +2,14 @@ import React from 'react'
...
+    <VertLayout ref={forwardedRef} {...props} />
+  ) : (
+    <HorzLayout ref={forwardedRef} {...props} />
```
jaketrent, line 12: "ref drilling"? :)

danethurber: drill baby drill

packages/navitem/src/react/index.js
```diff
@@ -2,22 +2,14 @@ import React from 'react'
...
-        {props.alignment === 'vertical' ? (
-          <VertLayout {...props} />
-        ) : (
-          <HorzLayout {...props} />
-        )}
-      </Button>
-      <Bar {...props} />
-    </Container>
+  return props.alignment === 'vertical' ? (
+    <VertLayout ref={forwardedRef} {...props} />
```
danethurber, line 10: the layout component props don't support the `alignment` prop but we're passing it...maybe we shouldn't

jaketrent: @danethurber ... I'm just passing everything with the spread operator.  It's an internal component.  Do you think we need to pass each prop explicitly instead?

danethurber: react will still throw a warning because `alignment` isn't a dom attrs. i'd just do a some rest args and spread

```
const {alignment, ...rest} = props

<Layout {...rest} />
```

you know the dril

jaketrent: React seems happy.  `npm run lint` has no errors and Travis CI is agreeing.  Does that change your mind?

danethurber: well i trust you...so now i'm doubting myself. one sec

danethurber: looks like the `Button` is filtering props and so alignment is not added to the dom because of other prop filtering. that seems like a side-effect and we should be more explicit - thoughts

jaketrent: Sure, I'll be explicit.

Just contrasting that explicitness principle in my mind to the potential use cases for spreading props: If we can't spread to internal components which we fully control because of a lack of explicitness, I'm not sure where spreading is appropriate.

Prepping a commit...

danethurber: yeah. it's also kinda of an edge case and really just a react warnings. not sure the best course of action but talking through it will continue to frame the issue

PR #1057, opened 2020-07-21
[tab]: Update Tab.ListItem type to use ReactNode as children

willsoto:
By making it a string typescript was not allowing the following content:

```
My items (50)
```

where 50 is a `number`. By changing it to `React.ReactNode`, that case is fixed and we now match the standard `PropsWithChildren` type in React:

```typescript
type PropsWithChildren<P> = P & { children?: ReactNode };
```

**I do notice that the `PropTypes` for the `ListItem` only wants a `string` - can that be updated?**

https://github.com/pluralsight/design-system/blob/dcf5395bee91dc24cd03e74dbcd100d30fa8025e/packages/tab/src/react/list-item.js#L55

### DIFF
--- a/packages/tab/src/index.d.ts
+++ b/packages/tab/src/index.d.ts
@@ -13,7 +13,7 @@ type List = FunctionComponent<ListProps & Styling>
 export interface ListItemProps {
     id: string | number
     active?: boolean
-    children?: string
+    children?: React.ReactNode
     onClick?: (i: number, event: MouseEvent) => void
 }
 interface ListItemAnchorProps {
--- a/packages/tab/src/react/list-item.js
+++ b/packages/tab/src/react/list-item.js
@@ -52,7 +52,7 @@ const ListItem = React.forwardRef((props, ref) => {
 })
 ListItem.propTypes = {
   active: PropTypes.bool,
-  children: PropTypes.string,
+  children: PropTypes.node,
   href: PropTypes.string,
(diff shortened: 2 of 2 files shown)


willsoto:
One thing to note here is that the following will raise a proptypes error:

```tsx
const pendingAdditions = 10;

<Tab.List>
  <Tab.ListItem
    id="add"
    onClick={() => setActiveTab("add")}
    active={activeTab === "add"}
  >
    Content to be added ({pendingAdditions})
  </Tab.ListItem>
</Tab.List>
```

```
Warning: Failed prop type: Invalid prop `children` of type `array` supplied to `ForwardRef`, expected `string`.
    in ForwardRef (created by PendingChangesPage)
    in PendingChangesPage
    in Router (created by MemoryRouter)
    in MemoryRouter
    in ApolloProvider (created by MockedProvider)
    in MockedProvider

  107 |         <Box padding="lg" className="container mx-auto">
  108 |           <Tab.List>
> 109 |             <Tab.ListItem
      |             ^
  110 |               id="add"
  111 |               onClick={() => setActiveTab("add")}
  112 |               active={activeTab === "add"}>
```

willsoto:
@jaketrent does that mean I am okay to change the `propTypes`?

jaketrent:
@willsoto yes, please update the proptypes too, and we'll get it merged.

willsoto:
@jaketrent - done

PR #1058, opened 2020-07-21
Style bar override on selected NavItem

jaketrent:
This is really the main use case for bar style overrides. We show colored bars per product when the navitem is selected.

### DIFF
--- a/packages/navitem/src/react/common.js
+++ b/packages/navitem/src/react/common.js
@@ -11,7 +11,8 @@ const styles = {
     css(
       stylesheet['.psds-navitem__bar'],
       props.selected && stylesheet['.psds-navitem__bar--selected'],
-      stylesFor('bar', props)
+      stylesFor('bar', props),
+      props.selected && stylesFor('bar--selected', props)
     ),
   button: props =>
     css(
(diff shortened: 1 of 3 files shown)


packages/navitem/src/react/common.js
```diff
@@ -11,7 +11,8 @@ const styles = {
     css(
       stylesheet['.psds-navitem__bar'],
       props.selected && stylesheet['.psds-navitem__bar--selected'],
-      stylesFor('bar', props)
+      stylesFor('bar', props),
```
EdwardIrby, line 14: I'm not sure this is a great idea. I haven't seen us do an override this way before. This breaks the pattern people use now of of passing in a className or inline style

jaketrent: Correct!  This is a new pattern.  You approved it the first time:

https://github.com/pluralsight/design-system/pull/1029

Yes, we currently allow className and style props.  This doesn't change that.  But it does enhance it.  

It allows us to specify how to style *sub components*.  If you're using style props, you can't do that.  If you use className, you're limited to what you know about the markup and coupling to that non-public interface.  Eg, 

```
.navitem > div {
  background: var(--flow-bg);
}
```

But what if there's more than one div or thing that you want to target?  Another limitation.

I don't know if this is the best solution.  But I do feel like there's precedent to it.  Styled-components and other css-in-js solutions provide themes this way.  As do some css module solutions.

Do you have an option for product-specific styles overrides in this or the general case that prefer?

EdwardIrby: Nope not really. I was concerned about platform lock in mostly.m Let's go with it for now.

PR #1059, opened 2020-07-22
feat(drawer): exposes underlying util useDrawerContext of drawer

EdwardIrby:
In regards to https://pluralsight.slack.com/archives/C70HPSJPP/p1595369904332500. Need this for a codepen fork. Modified the Drawer RC slightly removing unnecessary div and exposed useDrawerContext hook

### DIFF
--- a/packages/drawer/rc/index.js
+++ b/packages/drawer/rc/index.js
@@ -36,9 +36,9 @@ const styles = {
     ),
   collapsible: () => css(stylesheet['.psds-drawer__collapsible'])
 }
-const DrawerContext = createContext()
+export const DrawerContext = createContext()
 
-const useDrawerContext = () => {
+export const useDrawerContext = () => {
   const context = useContext(DrawerContext)
   if (!context) {
     throw new Error(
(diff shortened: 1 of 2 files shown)


packages/drawer/rc/index.js
```diff
@@ -36,9 +36,9 @@ const styles = {
     ),
   collapsible: () => css(stylesheet['.psds-drawer__collapsible'])
 }
-const DrawerContext = createContext()
+export const DrawerContext = createContext()
```
jaketrent, line 39: I don't understand how exposing these will help solve Ben's problem.

EdwardIrby: Actually I'll have it in my codepen example I realized won't work without it. Basically I'm going to let him turn the table row into a drawer head with a simple onClick callback. I though I had already exposed that but I didn't. I also removed the wrapping div because it serves no purpose and breaks accessibility on table Rows. Please not this is not the Default Drawer export but an RC I made some months back using the patterns we've been discussing.

EdwardIrby: This is currently broke but will work with this PR https://codesandbox.io/s/divine-feather-1mosn?file=/src/App.js.

I'm also going to include a row that does use the Drawer Head but I'll need to write some more custom CSS for that one and include the aria attribute for a a table row.

jaketrent: Thanks for the context.

PR #1065, opened 2020-07-27
NavCookieBanner

jaketrent:
### What You're Solving

- Creating new component for cookie banner, fulfilling a standard nav needs

### Design Decisions

- Hard code privacyUpdatedDate in the component with a prop to override.  Previously hard-coded in the prism server. Was put here so that Flow wouldn't have to consume the Prism server and its payload. Still unclear whether that will pan out or not. If not, this default can be removed.
- In the future, when the privacy policy is updated, a new version of this component can be released and used. Or teams could pass a new prop.


packages/navcookiebanner/src/react/index.js
```diff
@@ -0,0 +1,80 @@
...
+      media('print', css(stylesheet['@media print']['.psds-navcookiebanner']))
+    ),
+  message: () => css(stylesheet['.psds-navcookiebanner__message']),
+  dismiss: () => css(stylesheet['.psds-navcookiebanner__dismiss'])
+}
+
+const NavCookieBanner = React.forwardRef((props, forwardedRef) => {
+  const ref = React.useRef()
+  React.useImperativeHandle(forwardedRef, () => ref.current)
+  const [cookies, setCookie] = useCookies(['cookie-name'])
```
danethurber, line 30: what is `cookie-name` used for here

danethurber: in the src - the array is for dependencies for when to update. i'm guessing we can just pass it nothing?

jaketrent: It's supposed to be the name of the cookie I'm setting so that state updates can be detected.  Thanks for catching this.  This lets me get rid of the secondary state update for `clicked`.  This was my misuse of the lib.

EdwardIrby:
Latest change is cause the following error

**Cookie “prism-cookienotify” will be soon rejected because it has the “sameSite” attribute set to “none” or an invalid value, without the “secure” attribute. To know more about the “sameSite“ attribute, read https://developer.mozilla.org/docs/Web/HTTP/Headers/Set-Cookie/SameSite Cookies.js:48**

jaketrent:
@EdwardIrby Thank you.  Updated.  Please take another look.

PR #1076, opened 2020-08-03
Patch/position refs

### LINKED ISSUE #1075: Position: provided refs aren't usable
reported by @willsoto 

https://pluralsight.slack.com/archives/C70HPSJPP/p1595254691323600

danethurber:
resolves #1075


packages/util/src/clone-element-with-ref.js
```diff
@@ -0,0 +1,17 @@
+import { cloneElement } from 'react'
+
+import { isFunction, isRef } from './index.js'
+
+export function cloneElementWithRef(element, outerRef, props) {
+  const { ref } = element
+
+  return cloneElement(element, {
+    ...props,
+    ref: node => {
```
jaketrent, line 10: Would you please help me understand the need for the different cases here?

packages/util/src/clone-element-with-ref.js
```diff
@@ -0,0 +1,17 @@
+import { cloneElement } from 'react'
+
+import { isFunction, isRef } from './index.js'
+
+export function cloneElementWithRef(element, outerRef, props) {
+  const { ref } = element
```
jaketrent, line 6: Huh, I didn't know you could get the ref from react elements like this.  I'm surprised this is exposed.  :themoreyouknow:

packages/position/src/react/index.js
```diff
@@ -88,7 +91,7 @@ const Position = React.forwardRef((props, forwardedRef) => {
 
   return (
     <>
-      {target ? child : React.cloneElement(child, { ref })}
```
jaketrent, line 91: Apparently this didn't work to pass ref as a prop.  Did you learn why?  If so, would you mind explaining for my benefit, please?

danethurber: this overwrites the ref with the new ref we're providing.

packages/util/src/clone-element-with-ref.js
danethurber: sure can. 

`if (isRef(outerRef)) outerRef.current = node` just checks that the new ref we're appending exists

`if (isFunction(ref)) ref(node)` supports the older ref style
`else if (isRef(ref)) ref.current = node` supports the newer refs from `createRef`

packages/position/src/react/index.js
jaketrent: ah

EdwardIrby:
Just curious if position used context to provide inline styles to portaled element would we need to clone?

danethurber:
@EdwardIrby context wouldn't help to much in this case i don't think. we could provide the styles through a render prop and expect the implementor to get add the styles that way. or we might be able to get away from cloning by adding extra divs around the child nodes

PR #1083, opened 2020-08-06
Feat/actionmenu rebuild

EdwardIrby:
~~Need to rework Dropdown to work with action menu. Right now onClick works for both key and mouse events and returns evt, and value. To meet Nav needs we need to be able to  accept all sorts of content so label isn't really a think anymore.~~

Works with Dropdown which still uses onChange but no longer requires an onClick.
Please review stories under actionmenu and dropdowndown to better understand usage.


packages/actionmenu/src/css/index.js
```diff
@@ -48,85 +34,77 @@ export default {
...
-  [`.psds-actionmenu--origin-${vars.origins.topLeft}`]: {
-    ...menuOriginTop,
-    left: 0,
-    top: 0
-  },
-  [`.psds-actionmenu--origin-${vars.origins.bottomLeft}`]: {
-    ...menuOriginBottom,
-    left: 0,
-    bottom: 0
-  },
```
EdwardIrby, line 71: Need to add this back to get animations working better.

packages/actionmenu/src/js/index.js
```diff
@@ -1,62 +1,34 @@
...
-      bottom: `calc(-1 * ${vars.style.menuPaddingVert})`
+      right: parent.width,
+      top: -Math.abs(parent.height - menu.height)
     },
     bottomLeft: {
-      left: menuWidth + vars.style.nestedMenuHorzOverlap,
-      bottom: `calc(-1 * ${vars.style.menuPaddingVert})`
+      left: -parent.width,
+      top: -Math.abs(parent.height - menu.height)
     }
```
EdwardIrby, line 21: These may change once animations are fixed

packages/actionmenu/src/js/index.js
```diff
@@ -1,62 +1,34 @@
...
+export const ellipsesStyles = {
+  whiteSpace: 'nowrap',
+  overflow: 'hidden',
+  textOverflow: 'ellipsis'
+}
+
+export const iconLeftStyles = {
+  marginRight: layout.spacingXSmall,
+  flex: 'none'
+}
```
EdwardIrby, line 34: Not sure about these there styles that make adding things other than default text easy to style and align. Was considering adding a `iconRightStyles`  but wanted feedback on this approach first.

Issue #1038, opened 2020-06-29
actionmenu: support arbitrary content


jaketrent:
- [x] Arbitrary stuff, including non-list items
- [ ] Accessible access via keyboard for clickable stuff

Not doing:
- For now, or a part of this card: Support mobile experience ("slide, replace, back" experience)

Ref impl, including arbitrary content (see privacy and terms area):
<img width="1276" alt="Screen Shot 2020-08-10 at 10 42 12 AM" src="https://user-images.githubusercontent.com/552044/89807853-751f0a80-daf6-11ea-8b51-6f091dea484a.png">

jaketrent:
@sweeting Will you please add the use cases that you think we need to support here?  (Anything beyond current capability to support what is in your most recent designs.)

jaketrent:
Team meeting last Fri: This component might be a good place to test some of our more composable component concepts.  We don't want to have to update ActionMenu to support all the many new features that could/should be added.

Possibilities:
- "Popover" component?
- "ActionMenu" component inside?
- Atomic design principles
- Name matching with Figma symbols

PR #1088, opened 2020-08-07
feat(actionmenu): rebuild actionmenu to support variable uses

### LINKED ISSUE #1053: Actionmenu Rebuild
1. Rebuild menu.js using new menu-keydown-listener.js
2. Rebuild item.js to take arbitrary content
3. create a branch with youtube settings menu interaction

In regards to https://pluralsight.slack.com/archives/C70HPSJPP/p1594908803300800

Code Sandbox: https://codesandbox.io/s/friendly-glade-0qczd

Initial investigation shows that we are not using pseudo-event `hover` for the transition and instead detecting onMouseOver to trigger a class change. Initial recommendation is switching to pure css but first we need to detect unaccounted for side-effects of modifying this functionality. 

Might be related to keyboard focus state check for conflicts.

EdwardIrby:
Resolves  #1053
Pre-work for #1038

**Please review stories in ActionMenu and Dropdown.** 
Created a new Dropdown.Item based on conversation with Brian. Recommend ActionMenu becoming more of a base atom that provides interaction functionality but is used as seen in Dropdown.  

**Questions:** Why did Dropdown not just include ActionMenu internally versus cloning? Was it because there wasn't a way to do fragments yet and a root node was needed for passing in Items?


packages/actionmenu/src/css/index.js
```diff
@@ -10,19 +10,6 @@ import {
...
-const menuOriginLeft = {
-  transform: `translate(calc(-1 * ${layout.spacingXSmall}), 0)`
-}
-const menuOriginTop = {
-  transform: `translate(0, calc(-1 * ${layout.spacingXSmall}))`
-}
-const menuOriginBottom = {
-  transform: `translate(0, ${layout.spacingXSmall})`
-}
-
```
EdwardIrby, line 25: Working to add back these animations wanted to get 90% of functionality back

packages/note/src/react/__stories__/index.story.js
```diff
@@ -210,23 +210,21 @@ storiesOf('Note/actions', module)
         <>
           <NoteWithDefaults
             actionBar={[
-              <Below
+              <BelowLeft
                 when={isOpen}
                 show={
-                  <div style={{ position: 'relative' }}>
```
EdwardIrby, line 216: @danethurber I believe your Position fix mean we no longer have to wrap ActionMenu

packages/dropdown/src/react/item.js
```diff
@@ -0,0 +1,22 @@
...
+import ActionMenu from '@pluralsight/ps-design-system-actionmenu'
+
+import stylesheet from '../css/index.js'
+const styles = () => css(stylesheet[`.psds-dropdown__item__icon`])
+
+export const Item = ({ value, icon, children, ...rest }) => {
+  return (
+    <ActionMenu.Item tag="button" value={value} {...rest}>
+      {icon && <span {...styles()}>{icon}</span>}
+      <ActionMenu.Ellipsis>{children}</ActionMenu.Ellipsis>
```
EdwardIrby, line 13: How do you both feel about doing ellipsis this way in order to handle various menu item layouts?

packages/actionmenu/src/css/index.js
EdwardIrby: I now know what **origin** is for!

EdwardIrby:
> couple things i noticed:
> 
>     * icons are getting squished. maybe limit their flex shrink?
> 
> 
> Screen Shot 2020-08-10 at 8 30 31 AM
> 
>     * text is not vertically aligned to the center
> 
> 
> Screen Shot 2020-08-10 at 8 30 39 AM
> 
>     * the right arrow icon is not visible on hover.
> 
> 
> Screen Shot 2020-08-10 at 8 46 16 AM
> 
>     * i can use right arrow to enter a nested menu. i would have assumed i can use left arrow to go back but that doesn't work.
> 
>     * there are some test failures you'll want to look into
> 
>     * i'm wondering if we should have a new hover state on the selected item. going from blue to gray doesn't seem right cc @sweeting

- For the selected state talked to Brian about that last week and there was no change recommended. 
- Good catch on the arrow need to add another css rule
- I need to remove the tabindex from the button/anchor in the **ActionMenu.Item** and verify root menu if final point of focus before blur/close events.
- I'm not getting failing test locally after clean and bootstrap. I'll investigate. 
- For the icon I actually have question about the best way to handle that. If you look at `packages/dropdown/src/react
/item.js` you'll see I created a new component that composes **ActionMenu.Item**. To accept arbitrary content one needs to create and item that properly handles flexing content and this **Dropdown.Item** does that. Alternatively we could just take off the inline-flex display and make it block and then allow a user to do flex on their own by passing a **className** to the **ActionMenu.Item**. Thoughts?

danethurber:
> For the icon I actually have question about the best way to handle that. If you look at packages/dropdown/src/react /item.js you'll see I created a new component that composes ActionMenu.Item. To accept arbitrary content one needs to create and item that properly handles flexing content and this Dropdown.Item does that. Alternatively we could just take off the inline-flex display and make it block and then allow a user to do flex on their own by passing a className to the ActionMenu.Item. Thoughts?

I like either:

1: removing inline-flex and letting the developer handle layout...pending any gotchas i don't know about
2: having an ActionMenu.Icon component, to go along with the ActionMenu.Ellipsis that is used for text

EdwardIrby:
> > For the icon I actually have question about the best way to handle that. If you look at packages/dropdown/src/react /item.js you'll see I created a new component that composes ActionMenu.Item. To accept arbitrary content one needs to create and item that properly handles flexing content and this Dropdown.Item does that. Alternatively we could just take off the inline-flex display and make it block and then allow a user to do flex on their own by passing a className to the ActionMenu.Item. Thoughts?
> 
> I like either:
> 
> 1: removing inline-flex and letting the developer handle layout...pending any gotchas i don't know about
> 2: having an ActionMenu.Icon component, to go along with the ActionMenu.Ellipsis that is used for text

I threw this in vars when I submitted this PR

```js
export const iconLeftStyle = {
  marginRight: layout.spacingXSmall,
  display: 'inline-flex',
  alignItems: 'center'
}
```

EdwardIrby:
🤷🏿‍♂️

> > For the icon I actually have question about the best way to handle that. If you look at packages/dropdown/src/react /item.js you'll see I created a new component that composes ActionMenu.Item. To accept arbitrary content one needs to create and item that properly handles flexing content and this Dropdown.Item does that. Alternatively we could just take off the inline-flex display and make it block and then allow a user to do flex on their own by passing a className to the ActionMenu.Item. Thoughts?
> 
> I like either:
> 
> 1: removing inline-flex and letting the developer handle layout...pending any gotchas i don't know about
> 2: having an ActionMenu.Icon component, to go along with the ActionMenu.Ellipsis that is used for text

Decided to go with **2**. I want **1** but the last office hours makes me hesitant about expecting devs to build the UI parts not provided even if it is simple. Such as the following for both flex and to ensure ellipses

```js
display: 'flex',
    alignItems: 'center',
    textAlign: 'left',
    minWidth: 0
```

packages/util/src/__specs__/menu-keyup-listener.spec.js
```diff
@@ -68,12 +65,12 @@ describe('useMenuKeyEvents()', () => {
     let focused = getByTestId('root-list-item-0')
 
     // ArrowUp
-    fireEvent.keyDown(focused, { key: 'ArrowUp' })
+    fireEvent.keyUp(focused, { key: 'ArrowUp' })
```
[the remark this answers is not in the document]
EdwardIrby: I went with keyup because on rightArrow  I need the keydown to toggle attribute because ie11 does not support focus-within. I could however make them all keydown but have a special synthetic even callback event on Menu for rightArrow for the rightArrow. What do you think? Require a small mod.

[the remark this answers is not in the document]
EdwardIrby: Scratch that need to rework close logic to make that work.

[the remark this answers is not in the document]
danethurber: if it's not a big change i think it would be good just to keep the event similar to what they were...but i'll leave that decision up to you

PR #1102, opened 2020-08-14
Screen Reader for Progress Components

### LINKED ISSUE #1089: Linear/Circular Progress Screen Reader Text
Add text for screen readers to Linear and Circular progress components.

`"n% complete"`

jaketrent:
Resolves #1089


packages/circularprogress/src/react/index.js
```diff
@@ -55,6 +57,9 @@ const CircularProgress = React.forwardRef((props, ref) => {
       {...styles.circularprogress(themeName, props)}
       {...filterReactProps(props)}
     >
+      <ScreenReaderOnly role="region" aria-live="off">
```
EdwardIrby, line 60: should the aria-live be false here?

jaketrent: I set it to `off because of the MDN counsel:

> For regions which are not important, or would be annoying because of rapid updates or other reasons, silence them with aria-live="off".

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions

I imagined that "n% complete" would potentially be updated many, many times and not help to hear about in real time.

What do you think?

EdwardIrby: Reasoning makes sense to me 👍🏿

PR #1103, opened 2020-08-14
fix(actionmenu): ensures a util grabbing ref and correctly update coords

EdwardIrby:
See story: http://localhost:6006/?path=/story/navbar--desktop and click profile button


packages/actionmenu/src/react/item.js
```diff
@@ -134,8 +136,7 @@ export const Item = forwardRef(
 
 Item.displayName = 'ActionMenu.Item'
 Item.defaultProps = {
-  tagName: tagName.a,
-  origin: origins.topRight
```
jaketrent, line 138: I believe there should be a default origin and it should be topLeft.  Does this change enable that? 

I'm having a hard time telling.  Does origin come from anywhere besides a prop? Doesn't seem like it.

packages/navbar/src/react/__stories__/index.story.js
```diff
@@ -83,7 +83,7 @@ storiesOf('Navbar', module).add('desktop', _ => {
           user={
             <BelowRight
               show={
-                <ActionMenu>
+                <ActionMenu origin={ActionMenu.origins.topLeft}>
```
jaketrent, line 86: Shouldn't this be topRight?

packages/site/pages/components/actionmenu.js
```diff
@@ -271,10 +272,10 @@ export default _ => (
               'origin',
               PropTypes.union(ActionMenu.origins),
               null,
-              <code>topLeft</code>,
+              null,
               <span>
-                orientation (from <code>ActionMenu.origins</code>) of nested
-                menus
+                override orientation (from <code>ActionMenu.origins</code>) of
```
jaketrent, line 277: `origin` controls orientation.  But if we remove the default, I think we leave open the question of what the *default* behavior is.  I'd ensure topLeft as the default origin.

packages/actionmenu/src/react/item.js
EdwardIrby: Fixed in a different PR

packages/navbar/src/react/__stories__/index.story.js
EdwardIrby: Fixed in a different PR

EdwardIrby: Though technically correct given the flopflop

packages/site/pages/components/actionmenu.js
EdwardIrby: It's passed in via context on items no need for a default.

EdwardIrby: So the default is whatever you set on the Menu component. I could kill this override all together. I just figured someone will wan to do it.

jaketrent: The default is topLeft (from what I saw in your latest PR https://github.com/pluralsight/design-system/pull/1105).  I think we should keep that exposed as a defaultProps or at least in documentation.  This would serve the user of ActionMenu in learning what the observed default is called, helping him conceptualize what "topLeft" must mean and help him from there extrapolate what the others might mean.

PR #1114, opened 2020-08-21
Feat/dropdown selected item

### LINKED ISSUE #1108: dropdown: selected item state
Update the dropdown to use the following selected state (checkmark icon with Blue[7] fill). Also, when opening the dropdown menu, the focus should be on the selected item in the menu.

image

EdwardIrby:
resolve #1108 

See -> http://localhost:6006/?path=/story/menu--w-value


sweeting:
@EdwardIrby 

I expected the focus to be on the selected item when the menu is opened instead of the first item. Is that something we can do?

When the menu is closed, the focus should be on the select input, so hitting enter will open the menu again.

Our Dropdown

See: https://react-spectrum.adobe.com/react-spectrum/Picker.html
react spectrum picker

EdwardIrby:
> @EdwardIrby
> 
> I expected the focus to be on the selected item when the menu is opened instead of the first item. Is that something we can do?
> 
> When the menu is closed, the focus should be on the select input, so hitting enter will open the menu again.
> 
> Our Dropdown
> 
> See: https://react-spectrum.adobe.com/react-spectrum/Picker.html
> React Spectrum Picker



> @EdwardIrby
> 
> I expected the focus to be on the selected item when the menu is opened instead of the first item. Is that something we can do?
> 
> When the menu is closed, the focus should be on the select input, so hitting enter will open the menu again.
> 
> Our Dropdown
> 
> See: https://react-spectrum.adobe.com/react-spectrum/Picker.html
> React Spectrum Picker

Thanks for catching that. I'll debug and figure out why storybook is doing one thing and this another. Thanks

PR #1123, opened 2020-08-26
dialog: allow users to change styling of content div

jakemwood:
## Instructions

- fill in title: format of "<package name>: short description"
- label: attach bug or enhancement request label

### What You're Solving

- Flow designers are requesting the usage of different margins and paddings on dialog boxes. This will allow us to change styling of the inner content div created by the dialog component.

### Design Decisions

This seems like a pretty straightforward change, but am definitely open to suggestions.

### How to Verify

- Steps needed to independently verify additions

1. Create a dialog component with some content
2. Observe that margins are something like 24px.
3. Add a prop, `contentStyles={{ margin: 0 }}`
4. The inner content div rendered by the component should compose a new CSS class and assign it to the dialog contents div.

### DIFF
--- a/packages/dialog/src/react/index.js
+++ b/packages/dialog/src/react/index.js
@@ -27,7 +27,8 @@ const styles = {
           stylesheet[`.psds-dialog--tailPosition-${tailPosition}`]
         )
     ),
-  content: () => css(stylesheet['.psds-dialog__content']),
+  content: contentStyles =>
+    css(compose(stylesheet['.psds-dialog__content'], contentStyles)),
   close: _ => css(stylesheet['.psds-dialog__close']),
   overlay: _ => css(stylesheet['.psds-dialog__overlay'])
 }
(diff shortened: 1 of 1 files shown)


jaketrent:
@jakemwood Do you have a visual spec of what you're trying to accomplish? Maybe a screenshot in context of what you're trying to accomplish?

I'd like to learn why you or the your designer feels like the standard padding won't work.

packages/dialog/src/react/index.js
```diff
@@ -95,7 +96,8 @@ Dialog.propTypes = {
   modal: PropTypes.bool,
   onClose: PropTypes.func,
   tailPosition: PropTypes.oneOf(Object.values(Dialog.tailPositions)),
-  returnFocus: PropTypes.bool
+  returnFocus: PropTypes.bool,
+  contentStyles: PropTypes.object
```
jaketrent, line 100: This is a new style override pattern.  Looks pretty specific to this particular case as well.  I'd like to avoid adding a new prop for every dom node that devs want to restyle.

jakemwood:
Here is one design spec:

image

The designer is the Flow UI director, Michelle Heffron, if you'd like to reach out to her for your other questions.

jaketrent:
Triage: 

For this and future components where style overrides are desired, we will create this api:

UNSAFE_stylesFor={{
  "component": {},
  "component__child-selector--state": {
    [attr]: 'vals'
  } as CSSProperties
 }}

Break change: Remove style and change className to UNSAFE_className

Since this approach is different from the submitted PR, we will close this and create a card for our team to do this work for Dialog.

jaketrent:
Carded: https://github.com/pluralsight/design-system/issues/1127

PR #1125, opened 2020-08-28
Feat/dropdown patterns

EdwardIrby:
Refactors Dropdown to use Atomic pattern so hat devs can change behavior such as fixed icon in button or dynamic icon in button.

**See custom stories**


packages/dropdown/src/react/__stories__/index.story.js
```diff
@@ -578,3 +580,167 @@ storiesOf('portal', module).add('position', _ => (
...
+      }
+      onChange={handleChange}
+      menu={
+        <>
+          {Object.values(values).map(({ value, icon, label }) => (
+            <Dropdown.Item value={value} key={value} icon={icon}>
+              {label}
+            </Dropdown.Item>
+          ))}
+        </>
```
EdwardIrby, line 724: So there's some code around menu.props I need to review so that we can pass an array instead of just a fragment to the menu prop on Dropdown. This has nothing to do wit the refactor and is instead a  preexisting bug.

packages/dropdown/src/react/__stories__/index.story.js
```diff
@@ -578,3 +580,167 @@ storiesOf('portal', module).add('position', _ => (
...
+  }
+  return (
+    <DropdownWithIcon
+      icon={
+        values[selected] ? (
+          values[selected].icon
+        ) : (
+          <div style={{ width: 24, height: 24, marginRight: 8 }} />
+        )
+      }
```
EdwardIrby, line 715: Is this to much boilerplate? Should we also export some base styles for icon spacing and fallback when there is no icon?

packages/dropdown/src/react/__stories__/index.story.js
```diff
@@ -578,3 +580,167 @@ storiesOf('portal', module).add('position', _ => (
     ultrices eros in cursus turpis massa.
   </div>
 ))
+const DropdownWithIcon = forwardRef(
```
EdwardIrby, line 583: So this component maps to the refactor with the addition of an icon prop. After finishing everything I just realized I could just add the prop to the existing index.js in the react dir in Dropdown and it wouldn't be horrible prop creep.

jaketrent:
@EdwardIrby Ready for this to be reviewed for real?  (Seems like you may have indicated otherwise in chat.) If ready, we can review and get merged whiled you're out.

EdwardIrby:
> @EdwardIrby Ready for this to be reviewed for real? (Seems like you may have indicated otherwise in chat.) If ready, we can review and get merged whiled you're out.

Going to submit another PR in a short while that allows for the same but without the refactor. Choose whichever one you think works best for the DS.

EdwardIrby:
> @EdwardIrby Ready for this to be reviewed for real? (Seems like you may have indicated otherwise in chat.) If ready, we can review and get merged whiled you're out.

Go ahead a review I was mistaken on something being simpler would still require a refactor just not as extensive. I'm happy with this being reviewed.

packages/dropdown/src/react/__stories__/index.story.js
```diff
@@ -578,3 +580,167 @@ storiesOf('portal', module).add('position', _ => (
...
+                appearance={appearance}
+                label={longestMenuItemState.label}
+                placeholder={placeholder}
+                selectedLabel={selectedLabel}
+                size={size}
+              />
+            </div>
+          </Dropdown.Button>
+        }
+      />
```
EdwardIrby, line 675: If we wanted to not do this refactor we'd need to modify this part of the existing component structure to support icon and add the current `css/index.js` mods.

packages/dropdown/src/react/__stories__/index.story.js
danethurber: love the composability. any thoughts on ways we could avoid destructing the whole context and then passing the props on. maybe groups of like `menuProps`, `buttonProps`, etc. 

ie https://github.com/downshift-js/downshift/tree/master/src/hooks/useSelect#returned-props

packages/dropdown/src/js/index.js
```diff
@@ -0,0 +1,155 @@
...
+            : currentLongest
+        }
+      }, startLongest)
+    }
+
+    return getLongestState({ hasIcon: false, label: placeholder || '' }, menu)
+  }
+
+  const longestMenuItemState = getLongestMenuLabelState()
+  console.log(longestMenuItemState)
```
jaketrent, line 111: rm

packages/dropdown/src/react/__stories__/index.story.js
```diff
@@ -578,3 +580,167 @@ storiesOf('portal', module).add('position', _ => (
...
+        </>
+      }
+    />
+  )
+}
+
+storiesOf('custom', module)
+  .add('fixed icon', () => (
+    <DropdownWithIcon
+      icon={<Icon.CalendarIcon style={{ marginRight: 8 }} />}
```
jaketrent, line 733: Screen Shot 2020-09-01 at 4 05 14 PM

Padding left seems really big to me.  Perhaps double check design?

packages/dropdown/src/react/__stories__/index.story.js
jaketrent: Yeah, it's a little odd.  Perhaps we could detect an `icon` prop but a non-`<Icon />` value and then internally place a blank icon in the else case?

packages/dropdown/src/react/__stories__/index.story.js
jaketrent: Agreed, array support would be nice.

packages/dropdown/src/react/__stories__/index.story.js
jaketrent: The downshift pattern is interesting.  I could go along with it, but I don't know that we need that extra abstraction.

I would go the opposite route to cut down on destructuring:

```
const DropdownWithIcon = forwardRef((props, forwardedRef) => ...

// then...
<Dropdown.Button appearance={props.appearance}
```

And 

```
const context = useDropdown({ ... })

// then...
<Dropdown.Menu ref={context.menuRef} />
```

In other words, I think the style of destructuring every property out of giant objects leads to this problem.

packages/dropdown/src/react/__stories__/index.story.js
EdwardIrby: It's 16px on a default dropdown 8px on a small. I reduced this as it was larger on on our current dropdowns and didn't match the designs after pairing with Brian to verify.  However it looks like this component's padding has been reduced since I was I on PTO to 12px on default. Will update 👍🏿

packages/dropdown/src/react/__stories__/index.story.js
EdwardIrby: Decided to leave this as doing that is prop creep. However the suggestions below on destructuring make this less annoying and also stick to the goal making it so devs can add new behavior via composability.

PR #1126, opened 2020-08-31
[experimental] convert project to yarn

danethurber:
i think this is working but i could use a couple extra eyes

this is so we can speed up development and also because gatsby doesn't work great with the npm+lerna module resolution


package.json
```diff
@@ -14,20 +14,28 @@
...
+    "packages": [
+      "docs",
+      "packages/**"
+    ]
+  },
   "main": "packages/site/index.js",
   "scripts": {
-    "bootstrap": "lerna bootstrap --hoist && link-parent-bin",
-    "clean": "lerna clean",
+    "build": "lerna run build",
```
danethurber, line 26: i would like to use `wsrun` in the future but we have some cyclical dependencies we need to fix first

packages/util/package.json
```diff
@@ -28,7 +28,6 @@
   },
   "devDependencies": {
     "@pluralsight/ps-design-system-build": "^1.11.15",
-    "@pluralsight/ps-design-system-position": "^4.1.10",
```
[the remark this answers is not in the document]
danethurber: there was a stray circular dep i removed

package.json
```diff
@@ -14,20 +14,28 @@
...
+    "clean": "del '{,!(node_modules)/}**/dist'",
```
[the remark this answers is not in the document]
danethurber: we shouldn't need bootstrap because yarn is linking the workspaces together...that's how i understand it

package.json
```diff
@@ -5,7 +5,7 @@
   "license": "Apache-2.0",
   "private": true,
   "engines": {
-    "node": ">= 7.9",
+    "node": ">= 11.4",
     "npm": ">= 4.2"
   },
```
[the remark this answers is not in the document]
danethurber: good point. i'll update

.travis.yml
```diff
@@ -17,12 +22,12 @@ jobs:
     - stage: validate
       name: Lint Commits
       script:
-        - npm install @commitlint/travis-cli
+        - yarn global add @commitlint/travis-cli
```
jaketrent, line 26: why global?

.travis.yml
```diff
@@ -12,7 +12,7 @@ language: node_js
 if: type != push OR (tag IS blank AND branch = master)
 
 install:
-  - yarn
+  - yarn install --ignore-engines
```
jaketrent, line 16: Why is this on by default in yarn?  And if it's a good reason, why don't we fix the problems we're having with it?

package.json
```diff
@@ -14,16 +14,16 @@
...
     "nohoist": [],
     "packages": [
       "docs",
       "packages/**"
     ]
   },
-  "main": "packages/site/index.js",
   "scripts": {
-    "build": "lerna run build",
+    "build": "lerna run build --ignore @pluralsight/ps-design-system-docs",
```
jaketrent, line 22: why ignore docs on build?

package.json
```diff
@@ -137,6 +136,7 @@
     "ts-loader": "^6.2.2",
     "typescript": "^3.9.3",
     "webpack": "^4.41.2",
+    "wsrun": "^5.2.1",
```
jaketrent, line 135: I didn't think we were using this yet.

plop-templates/component/package.json.hbs
```diff
@@ -14,11 +14,11 @@
     "build:cjs": "babel --root-mode upward src --out-dir dist/cjs",
     "build:css": "build-css --useGlamor -i dist/cjs/css/index.js",
     "build:js": "cross-env ESM=true babel --root-mode upward src --out-dir dist/esm",
-    "build:watch": "npm run build:js -- --watch",
-    "prepublish": "npm run build",
+    "build:watch": "yarn build:js -- --watch",
```
jaketrent, line 17: no ` run` required here, right?

tsconfig.json
```diff
@@ -9,6 +9,7 @@
     "moduleResolution": "node",
     "outDir": "dist",
     "rootDir": "src",
+    "skipLibCheck": true,
```
jaketrent, line 12: we start skipping the type check of typings at this point?

Dockerfile.docs
```diff
@@ -1,31 +1,19 @@
...
 
-RUN npm set progress=false
-RUN npm set unsafe-perm true
-
-RUN npm config set depth 0
-
-## Use app/ subdir so root tsconfig can be extended
-RUN mkdir -p ./app
-COPY docs/. ./app/
-COPY tsconfig.json .
```
jaketrent, line 18: Is there not a need to have the base tsconfig.json copied over any more?

Dockerfile.docs
```diff
@@ -70,8 +56,7 @@ USER node
 HEALTHCHECK --interval=5s --timeout=30s --retries=3 \
   CMD node /usr/src/app/health-check.js
 
-ENTRYPOINT ["/usr/local/bin/npm", "run"]
-CMD ["start"]
+CMD ["yarn", "start"]
```
jaketrent, line 59: Read an interesting bit about using yarn vs node in the CMD of dockerfiles today.

See num 2: https://dev.to/nodepractices/docker-best-practices-with-node-js-4ln4

packages/core/tsconfig.json
```diff
@@ -2,13 +2,10 @@
   "extends": "../../tsconfig.json",
   "compilerOptions": {
     "declaration": true,
-    "module": "CommonJS",
```
jaketrent: Justin set these up and they were working.  Are we positive the package will still be compatible with current usage based on these changes? (ie "module", "target" and "rootDir" changing)

.gitlab-ci.yml
```diff
@@ -1,6 +1,6 @@
 workflow:
   rules:
-    - if: $CI_COMMIT_MESSAGE =~ /build: publish/
+    - if: $CI_COMMIT_MESSAGE =~ /publish/
```
jaketrent, line 3: "publish" is not an uncommon word in programming, esp in a codebase with pubsub in it.  These doesn't seem specific enough.

.eslintrc.js
```diff
@@ -27,6 +27,7 @@ module.exports = {
     'react-hooks/rules-of-hooks': ERROR,
     'react/display-name': WARNING,
     'react/jsx-no-bind': IGNORE,
+    'react/jsx-pascal-case': IGNORE,
```
jaketrent, line 30: where are we breaking this rule?

jaketrent (commented):
Lots of questions. There were lots of changes. I bet this took you more than 1/2 an hr. ;)

plop-templates/component/package.json.hbs
danethurber: right. yarn gets rid of all that run silliness

package.json
danethurber: i will remove

package.json
danethurber: because it was slow. i'm actually going to change up the postinstall not to build. there will now be a build step you'll need to do after installation

.travis.yml
danethurber: because yarn is weird with engines. not sure how to fix it

.travis.yml
danethurber: copy/pasta. i'll just dev install it on travis

Dockerfile.docs
danethurber: 👀

.gitlab-ci.yml
danethurber: good point. i'll fix once i can get the regex passed the ci linter

.eslintrc.js
danethurber: it's mad about `Text.P`. seems like a bug in the plugin

Dockerfile.docs
danethurber: oops. nice catch

.eslintrc.js
EdwardIrby: This fixes itself if you run npm i

EdwardIrby (approved):

PR #1130, opened 2020-09-02
docs layout

danethurber:
- Provides basic layout elements(header, aside, main)


docs/gatsby-config.js
```diff
@@ -4,22 +4,37 @@
...
+  siteMetadata: {
+    title: 'Pluralsight Design System'
+  },
   plugins: [
     {
-      resolve: `gatsby-source-filesystem`,
+      resolve: 'gatsby-source-filesystem',
       options: {
-        name: `pages`,
+        name: 'pages',
```
danethurber, line 14: key if/when we have different layouts per content type

docs/src/components/frame/aside.module.css
```diff
@@ -0,0 +1,58 @@
+@import '@pluralsight/ps-design-system-core';
+
+@custom-media --breakpoint-small (width >= 768px);
```
danethurber, line 3: can't figure out a way to use cssvars in media queries 🤷 

also i need to figure out how to make these media definitions globally defined/available

docs/src/components/frame/aside.tsx
```diff
@@ -0,0 +1,33 @@
...
+import useOnEscape from '../../hooks/use-on-escape'
+
+import styles from './aside.module.css'
+
+interface Props extends HTMLAttributes<HTMLDivElement> {
+  onRequestClose: () => void
+  open: boolean
+}
+
+export const Aside: React.FC<Props> = (props) => {
```
danethurber, line 16: i'm doing all named exports for components....thoughts?

docs/src/components/layouts/default.tsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react'
+import { MDXProvider } from '@mdx-js/react'
+
+import { Frame } from '../frame'
+import { SideNav } from '../side-nav'
+
+const shortcodes = {}
```
danethurber, line 7: this is just a placeholder for if/when we use markdown shortcodes

docs/src/hooks/use-match-media.ts
```diff
@@ -0,0 +1,42 @@
+import { useCallback, useEffect, useState } from 'react'
+import { canUseEventListeners } from 'exenv'
+
+import useHasMounted from './use-has-mounted'
+
+export default function useMatchMedia(query: string): boolean {
```
danethurber, line 6: i didn't end up using this hook but i left it in just in case

danethurber (commented):
two things i'll fix in follow prs

- general page elements (headings, p, etc) are not theme aware
- eslint in the docs dir is unhappy (typescript integration needed)

docs/src/components/frame/aside.tsx
jaketrent: I dig.

docs/src/components/frame/aside.module.css
jaketrent: Yeah, I think you found the solution: https://cssnext.github.io/features/#custom-media-queries

Apparently you can't use var() in media queries: https://stackoverflow.com/a/40723269

docs/package.json
```diff
@@ -28,17 +28,27 @@
     "dotenv": "^8.2.0",
     "express": "^4.17.1",
     "gatsby-plugin-express": "^1.1.6",
+    "gatsby-plugin-react-helmet": "^3.3.10",
```
jaketrent, line 31: magic helmet?! https://www.youtube.com/watch?v=5kH0Bag0akc

docs/postcss.config.js
```diff
@@ -0,0 +1,17 @@
+/* eslint-env node */
+/* eslint-disable */
+
+const mediaPlugin = require('postcss-custom-media')
```
jaketrent, line 4: looks like a good set to me.

docs/src/hooks/use-on-document-click.ts
```diff
@@ -0,0 +1,18 @@
+import { useEffect } from 'react'
+import { canUseEventListeners } from 'exenv'
+
+export default function useOnDocumentClick(
```
jaketrent, line 4: I think you could use this instead: https://github.com/pluralsight/design-system/blob/master/packages/util/src/use-close-on-document-events.js

docs/src/hooks/use-on-escape.ts
```diff
@@ -0,0 +1,25 @@
+import { useCallback, useEffect } from 'react'
+import { canUseEventListeners } from 'exenv'
+
+const ESCAPE_KEY = 27
+
+export default function useOnEscape(
```
jaketrent, line 6: Let's util this instead of making yet another copy.

docs/tsconfig.json
```diff
@@ -1,8 +1,10 @@
 {
   "extends": "../tsconfig.json",
   "compilerOptions": {
+    "jsx": "react",
+    "outDir": "dist",
     "rootDir": "src",
-    "outDir": "dist"
+    "target": "es5"
```
jaketrent, line 7: Tell me about how you came to this "target" option.

docs/src/hooks/use-has-mounted.ts
```diff
@@ -0,0 +1,17 @@
+import { useEffect, useRef } from 'react'
```
jaketrent, line 1: Util candidate

jaketrent (approved):
Looks fun!  We get to start fresh.

docs/src/hooks/use-on-document-click.ts
danethurber: yep. these are all copied from other parts of the project. i'll follow up with a pr that centralizes them

docs/tsconfig.json
danethurber: looking at the gatsby docs

docs/typings/pluralsight.d.ts
```diff
@@ -0,0 +1,8 @@
+declare module '@pluralsight/*'
+
+declare module '@pluralsight/ps-design-system-theme' {
```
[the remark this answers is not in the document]
danethurber: for now i'd just like to fix the ts errors...and then we can publish actual typings with all the packages

PR #1132, opened 2020-09-03
Sidenav links and Color & Type Page

jaketrent:
asdf


danethurber (changes requested):
just one nitpick. the truncation of the variable name is gross. can we find a better way to do it

Screen Shot 2020-09-08 at 3 13 29 PM

jaketrent:
@danethurber yes, we can look at better ways to make that all fit.  Can we merge the old design before we redesign?

danethurber (approved):

jaketrent:
@danethurber what do you think of this redesign? https://github.com/pluralsight/design-system/pull/1132/commits/fe3d11fc9308964654205c0d36e602e2eabe878c

Adds more horz space for var names. Adds JS name in too.

If we go with it, I should probably tweak the rest of the page to match.

docs/src/components/typography/font-weight.tsx
```diff
@@ -3,30 +3,56 @@ import React from 'react'
...
+            className={styles.weightLetters}
+          >
+            Aa
+          </div>
+          <div className={styles.weightLabel}>
+            {w.label} ({w.weight})
+          </div>
+          <div className={styles.labels}>
+            <div>JS: {w.jsVarName}</div>
+            <div>CS: {w.varName}</div>
```
danethurber, line 54:
```suggestion
            <div>CSS: {w.varName}</div>
```

danethurber (approved):
it works for me. @sweeting should probably chime in

jaketrent:
@sweeting do you like this new design for the font weight section? or keep the old (see above)? or create something original and fresh?

Screen Shot 2020-09-08 at 3 49 46 PM

sweeting:
Yeah, that accommodates the space for the variables without truncating. Let's use this style for the variables. 
image

PR #1141, opened 2020-09-08
convert Avatar to typescript

danethurber:
todo:

- [x] ~generate proptypes~ i need to evaluate the following options. this will come in another pr
  - https://github.com/gcanti/prop-types-ts: requires a whole new ts library/syntax
  - https://www.npmjs.com/package/babel-plugin-typescript-to-proptypes: will require using babel to build
- [x] fix build
- [x] ~fix storybook addons~ i hit some snags with the storybook addons. i'm going to open another pr than includes storybook updates and new addons

PR #1143, opened 2020-09-08
emptystate: add TypeScript types

KaidenR:
### What You're Solving

- Adds TypeScript types for the EmptyState component

### Design Decisions

- Tried to follow the example from CircularProgress


.gitignore
```diff
@@ -40,6 +40,9 @@ jspm_packages
 # Optional REPL history
 .node_repl_history
 
+# IDE config directories
```
jaketrent, line 43: We don't include IDE paths for this reason: https://jaketrent.com/post/gitignore-system-files/

jaketrent (changes requested):
Just a lil nit. 

From here, we're interested in converting this package (and all the rest, eventually) to TS.  Any interest in doing that w/in this PR?

.gitignore
KaidenR: So is the suggestion that it should be in your global .gitignore instead of in the repo?
In DevStar repos we actually do commit parts of the .idea folder (code style settings, run configs, live templates, etc...) so we don't have it to our global .gitignore.

jaketrent: yeah, at the least, please don't include `.idea` in this project's gitignore.

KaidenR:
> Just a lil nit.
> 
> From here, we're interested in converting this package (and all the rest, eventually) to TS. Any interest in doing that w/in this PR?

Sounds like fun. Do you have an example of a component that has already been converted that I can use as an example?

.gitignore
KaidenR: Any suggestions on how I can prevent those files from being committed here but not in the other repos I work in?

jaketrent:
@KaidenR Well, I was going to point you to circular progress, but it hasn't had a *complete* transformation yet.  It does a few more things:

- Create tsconfig.json
- Make storybook run happily w/ TS

Maybe we just take it that far for the moment?

.gitignore
jaketrent: If you don't want to use a system gitignore for this, I would just be more specific in what you `git add` w/in this project.

KaidenR:
@jaketrent Okay the story for that component is converted to TS. I attempted to convert the whole component over but I was having to make too many choices that I felt your team should be the deciders for:
- How the build steps work
- How component types should be set up and organized
- How (if?) the SVG generation in that component would be converted to TS

So I just opted to convert the story over which was still nice because it helped me find a bug with my `.d.ts` file. But this change should give your team a good place to start converting a component entirely to TS.

A couple decisions I made:
- The main problem I was struggling with was how deal with imports of non-TS modules. I ended up creating a `typings/untyped-ds-modules.ts` file in the root dir which will turn those modules into anys.
- I created a `base-tsconfig.json` in the root dir which each module should be able to extend to give it all the common settings.

.gitignore
KaidenR: Figured it out. I just added an entry to my global .gitignore for just the .idea folder in that repo.

packages/emptystate/tsconfig.json
```diff
@@ -0,0 +1,7 @@
+{
+  "extends": "../../base-tsconfig.json",
```
danethurber, line 2: you should just be able to use the tsconfig that is already in the root. this pr that i'm working on is doing just that

https://github.com/pluralsight/design-system/pull/1141

@KaidenR do you want to sync up maybe 😉

packages/emptystate/src/react/__stories__/index.story.tsx
```diff
@@ -1,15 +1,15 @@
 import { storiesOf } from '@storybook/react'
-import PropTypes from 'prop-types'
-import React from 'react'
+import * as React from 'react'
```
danethurber, line 2: 😢

danethurber: because the `*` is going to confuse me 😆

KaidenR: Oh I think I might be able to change that back. I think I made that change before turning on the `esModuleInterop` flag.

packages/emptystate/tsconfig.json
KaidenR: Oh I didn't realize you were already working on converting something to TS 🤦  
Should I just wait for that one to wrap up and then look at how that affects this PR?

danethurber: sorry 😬  - i'm going to switch gears and try to wrap up my pr today. i'll be asking you for a review...you've gotten yourself into

jhewlett:
@KaidenR Not sure if it's helpful at this point, but here's a PR you can look at for reference: https://github.com/pluralsight/design-system/pull/965/files

KaidenR:
@jhewlett Good suggestion. I think since the only TS file is the storybook story we could wait until the rest of the component is TypeScript before adding a compile check.

jhewlett:
> @jhewlett Good suggestion. I think since the only TS file is the storybook story we could wait until the rest of the component is TypeScript before adding a compile check.

It's a good way to check the story against the `.d.ts` file and detect regressions. Otherwise you're relying on someone manually checking that their editor doesn't report any errors. A few months ago we shipped a broken type definition because I accepted a change to my PR on github without thinking to pull it down and check it locally. Had the `tsc` been in place at the time it would have failed CI.

danethurber:
closing in favor of https://github.com/pluralsight/design-system/pull/1146/files

PR #1144, opened 2020-09-09
Motion Page Doceroos

jaketrent:
Doc it up!


docs/src/components/motion/easing.module.css
```diff
@@ -0,0 +1,166 @@
+b {
```
danethurber, line 1: haven't seen a `b` in a while. why use this instead of `a strong`?

danethurber (approved):

docs/src/components/motion/easing.module.css
jaketrent: no reason.  it's been there for years.

PR #1150, opened 2020-09-11
Type table UI for docs

jaketrent:
And a few more tsc fix-it-whatsits for the spacing page


danethurber (approved):
looking good. the colors are pretty hard to read in the dark theme - not sure if @sweeting would have any thoughts there

jaketrent:
Great point on the dark theme support.  Previously there wasn't any.  I'll take a stab at putting in new colors for dark theme.

PR #1151, opened 2020-09-12
storybook upgrade and refactor

danethurber:
ok...there are a lot of changes here but it's a lot of the same. here are the big things that actually changed

- converted `Avatar` to typescript. i did this to test out the storybook changes
- bump babel and storybook dependencies
- convert storybook addons(center and theme) to use new storybook apis
- create a new storybook preset package to centralize as much storybook config as i could
- use the new preset in all component packages
- include tsconfig files for all packages for storybook and for future ts conversion
- the new storybook addons don't show up in storyshots anymore....so there are lots of snapshot changes
- remove chaos and featureflags addons that aren't used

The new theme toggle for storybook is in the topbar and looks like this
Screen Shot 2020-09-14 at 1 15 22 PM


packages/actionmenu/src/react/__specs__/storyshots.spec.js
```diff
@@ -3,7 +3,7 @@ import initStoryshots, {
   snapshotWithOptions
 } from '@storybook/addon-storyshots'
 
-const createNodeMock = el => document.createElement('div')
+const createNodeMock = () => document.createElement('div')
```
danethurber, line 6: this makes the linter happier

packages/appframe/src/react/__specs__/storyshots.spec.js
```diff
@@ -3,9 +3,7 @@ import initStoryshots, {
   snapshotWithOptions
 } from '@storybook/addon-storyshots'
 
-jest.mock('@pluralsight/ps-design-system-storybook-addon-theme')
```
danethurber, line 6: as far as i can tell, we no longer need to mock addons in the storyshots

packages/carousel/.storybook/main.js
```diff
@@ -0,0 +1,13 @@
+module.exports = {
+  addons: [
+    {
+      name: '@pluralsight/ps-design-system-storybook-preset',
+      options: { center: false }
```
danethurber, line 5: i've setup the preset so we can disable addons as needed. there are some packages that don't look good when centered. i've disabled the centering for the stories i could find - hopefully i didn't miss any

jaketrent (approved):
After clean and bootstrap, 
npm test is clean
storybook looks good for avatar and actionmenu, I hope that's representative
Only 488 files changed, nbd.

PR #1152, opened 2020-09-14
Mind your Dos and Donts

jaketrent:
asdf


docs/src/components/guideline.module.css
```diff
@@ -0,0 +1,18 @@
+.guideline {
```
danethurber, line 1: i think some bottom margin would help 

Screen Shot 2020-09-14 at 1 43 50 PM

danethurber (changes requested):
good to merge after you look at my one comment

also there is no dark theme change - @sweeting any thoughts on that?

docs/src/components/guideline.module.css
jaketrent: mmm, yes, always helps

danethurber (approved):

jaketrent:
add lots o margin and dark theme suPPort

PR #1158, opened 2020-09-15
Example switcher

jaketrent:
- Wanted to author in markdown since this is the common case for code authoring in docs
- Chose code block in markdown because it allows an opening and closing symbol
-  Chose a --- / hr symbol to split examples
- Downside to markdown is lack of rich data structure. Best idea so far: in a following PR, parse jsdocs for each example to parse out title and description

### DIFF
--- a/docs/src/components/mdx/code-block.tsx
+++ b/docs/src/components/mdx/code-block.tsx
@@ -1,4 +1,5 @@
 import Button from '@pluralsight/ps-design-system-button'
+import Dropdown from '@pluralsight/ps-design-system-dropdown'
 import Theme, { useTheme } from '@pluralsight/ps-design-system-theme'
 import cx from 'classnames'
 import Prism from 'prismjs/components/prism-core'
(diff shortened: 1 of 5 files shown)


docs/src/components/mdx/code-block.tsx
```diff
@@ -1,22 +1,80 @@
...
 export const CodeBlock: React.FC<CodeBlockProps> = (props) => {
-  const { children: code } = props
+  const isSwitcher = /switcher/.test(props.metastring)
   const language = props.className.replace(/language-/, '')
 
+  const codes = props.children.split('\n\n---\n\n')
+  const options = isSwitcher
+    ? Array.from(Array(codes.length)).map((_, i) => ({
+        value: i,
+        label: 'Example #' + i,
```
danethurber, line 27: obviously a way to define a title will be nice to add down the rpad

docs/src/components/mdx/code-block.tsx
```diff
@@ -1,22 +1,80 @@
...
+      }))
+    : [{ value: 0, label: 'Single example' }]
+
+  const [selectedOption, setSelectedOption] = React.useState(options[0].value)
```
danethurber, line 31: i'm not getting the initially selected option in the dropdown value

Screen Shot 2020-09-15 at 9 37 24 AM

docs/src/pages/index.mdx
```diff
@@ -56,18 +28,35 @@ const Example: React.FC = () => {
...
-```typescript codesandbox=default
+---
+
+/*
+ * For now, the description can live in a code block.
+ * 
+ * If we don't put this in markdown, we will be authoring in .tsx instead of .mdx.
+ * 
+ * Later, we can update code-block.tsx to extract these, parse markdown and display as html
+ */
```
danethurber, line 40: maybe we write in frontmatter style and parse it out with something like https://github.com/docs/frontmatter

danethurber (approved):
love the approach. couple comments. nice work!

jaketrent:
Frontmatter was a great idea. I was going to use a kind of jsdoc. Fixed the selected options too.  There may be a bug or 2 in the dropdown. If the value is just a number, the item label gets followed by a 0 and the label is not shown in the dropdown when selected. Fixed that by prepending 'value' to the item values.

PR #1162, opened 2020-09-15
We'll do it live!

jaketrent:
- Should be compatible with how we've had to structure codesandbox integration
- Is not currently compatible with TS.  Went down the road of @babel/standalone, and that creating many a hurdle. Backed it out so we could go this far. Not convinced we need TS inside examples anyway.


danethurber:
stupid typescript https://github.com/uber/react-view has some typescript support if we want to pivot

jaketrent:
Looking at uber/react-view, I think I'll avoid because:

- Adding TS support to it is similar to FormidableLabs/react-live, via babel plugins
- react-live is more flexible with imperative code (ie using `noInline`).  uber/react-view requires a single expression or component.

I'll experiment with babeling with react-live, but I don't feel like the ROI is great for me to spend much time on it.

jaketrent:
TS now LIVE! Plz2Review...

danethurber (approved):

jaketrent:
Hey @danethurber , would you mind having another look.  I didn't branch as I thought I did.  The latest 4 commits are button page-related with an important tweak to the handling of imports on the code block component.

danethurber (approved):
well-do-it-live


two nitpicks....feel free to merge when you've addressed them

--- 
### values are cut off
Screen Shot 2020-09-16 at 2 12 45 PM

---
### i think the include code should just wrap the single words and not create a whole block
Screen Shot 2020-09-16 at 2 12 51 PM

PR #1166, opened 2020-09-16
refactor(screenreaderonly): convert to typescript

danethurber:


packages/screenreaderonly/src/react/index.tsx
```diff
@@ -0,0 +1,18 @@
...
+interface Props {
+  children?: ReactNode
+  as?: React.ElementType | keyof JSX.IntrinsicElements
+}
+
+const ScreenReaderOnly = forwardRef<any, Props>((props, ref) => {
+  const { as: Tag = 'div', ...rest } = props
+
+  return <Tag ref={ref} {...css(accessibility.screenReaderOnly)} {...rest} />
+})
```
danethurber, line 14: some comments.

- typescript makes it really hard to get true type safety with the `as` polymorphic prop. srsly...i went down some rabbit holes. i just gave up
- `forwardRef` does not have a good way to do generics or overloading
- a better solution in the future for the polymorphic `as` component is not to use it...but instead use a render prop https://blog.andrewbran.ch/polymorphic-react-components/#an-alternative-approach

jaketrent: I'm fine to move the render prop approach now or later.

jaketrent (approved):

PR #1169, opened 2020-09-17
feat(docs): persist theme selection

jaketrent:
## To validate
- Load site
- Click theme toggle
- Refresh and/or go to other pages, see that theme is kept


EdwardIrby (approved):
Cookies change correctly

PR #1170, opened 2020-09-17
Fix non-TS examples from being rendered, then refactor code block

jaketrent:
CSS was trying to render live
Code block needed a cleanup


jaketrent:
large rewrite already done on code block. closing.

PR #1175, opened 2020-09-18
fix(docs): stop code from causing page to overflow past viewport

danethurber:


docs/src/components/frame/frame.module.css
```diff
@@ -24,9 +24,11 @@
 }
 
 .light {
+  background: var(--psColorsWhite);
```
danethurber, line 27: just because 🤷

EdwardIrby (approved):

PR #1177, opened 2020-09-18
feat(docs): change switcher delimiter

danethurber:
proposal: change the delimiter for segmenting examples in the code example switcher to be easy to read

`---` is used for frontmatter...so i'm thinking that `===` is different enough to make the codeblocks easier to understand

### DIFF
--- a/docs/src/pages/index.mdx
+++ b/docs/src/pages/index.mdx
@@ -14,7 +14,7 @@ const Comp = React.FC = () => <p>example 1</p>
 
 export default Comp
 
----
+===
 
 ---
 title: some title 2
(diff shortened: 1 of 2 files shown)


docs/src/pages/index.mdx
```diff
@@ -14,7 +14,7 @@ const Comp = React.FC = () => <p>example 1</p>
 
 export default Comp
 
----
+===
```
EdwardIrby, line 17: What's this do?

EdwardIrby (approved):

docs/src/pages/index.mdx
jaketrent: separates examples in a single markdown code block.

jaketrent (approved):
yeah, works for me. Good idea.

PR #1179, opened 2020-09-18
fix heading alignment in example switcher

danethurber:
before:
Screen Shot 2020-09-18 at 9 09 59 AM

after:
Screen Shot 2020-09-18 at 9 12 35 AM

### DIFF
--- a/docs/src/components/code-block/styles.module.css
+++ b/docs/src/components/code-block/styles.module.css
@@ -1,10 +1,12 @@
 @import url('https://fonts.googleapis.com/css?family=Source+Code+Pro:500');
 
-.title {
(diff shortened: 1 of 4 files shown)


docs/src/components/code-block/styles.module.css
```diff
@@ -1,10 +1,12 @@
...
 
-.title {
+.header {
   display: flex;
   align-items: center;
-
-  & > h2:first-child {
+}
+.title {
+  &:not(.hack-to-make-more-specific) {
```
danethurber, line 8: 😬

jaketrent (approved):

PR #1186, opened 2020-09-19
verticaltabs: allow custom onClick on tier1 header

KaidenR:
### What You're Solving

- `onClick` functions set on `<VerticalTabs.Tier1.Header />` components are overridden if the tier is collapsible. This change calls the custom `onClick` if provided after toggling the tier.
- To get around this problem I'm currently using an `onClick` on the tier1 component itself but that gets called when the child nav items are clicked too. I only want it to be called when the tier is expanded/collapsed.

### How to Verify

- Add an onClick function to a collapsible tier1's header. Verify the fn gets called when clicking on the header.


jaketrent (approved):

jaketrent:
Looks great, Kaiden.  I'll get it merged and released.

PR #1191, opened 2020-09-21
fix(docs): bad dropdown docs core imports

EdwardIrby:
fixes core import


jaketrent (commented):
Now fixed in master earlier this morning. I think we can close this.

PR #1192, opened 2020-09-21
Code block enhancements

jaketrent:
- Provide example data to avoid prop drilling
- Support not rendering, even TS examples
- Collapse by default
- Expand for short examples by default
- Add fade out graphic to indicate expandability


docs/src/components/code-block/editor.tsx
```diff
@@ -1,49 +1,49 @@
...
 
   return (
     <Highlight
       {...defaultProps}
-      code={code}
-      language={language}
-      theme={codeTheme}
+      code={props.children}
+      language={context.language}
+      themeName={codeTheme}
```
danethurber, line 26:
```suggestion
      theme={codeTheme}
```

docs/src/components/code-block/editor.tsx
```diff
@@ -1,49 +1,49 @@
...
-import Highlight, { Language, defaultProps } from 'prism-react-renderer'
-import React, { HTMLAttributes } from 'react'
+import Highlight, { defaultProps } from 'prism-react-renderer'
+import React, { HTMLAttributes, useContext } from 'react'
 
+import { CodeBlockContext } from './index'
 import styles from './styles.module.css'
 import { darkTheme, lightTheme } from './theme'
 
 interface EditorProps extends HTMLAttributes<HTMLPreElement> {
```
danethurber, line 11: if we aren't going to spread the props on the child node then we can remove the `extends HTMLAttributes<HTMLPreElement>`

docs/src/components/code-block/editor.tsx
```diff
@@ -1,49 +1,49 @@
...
 
         const className = cx({
           [highlight.className]: true,
           [styles.editor]: true,
-          [styles.editorExpanded]: expanded
+          [styles.editorExpanded]: props.expanded
         })
 
         return (
-          <pre className={className} {...rest}>
```
danethurber, line 39: here was the spread that i mentioned

docs/src/components/code-block/editor.tsx
```diff
@@ -1,49 +1,49 @@
...
+          <pre className={className}>
             {tokens.map((line, i) => (
               <div key={i} {...getLineProps({ line, key: i })}>
                 {line.map((token, key) => (
                   <span key={key} {...getTokenProps({ token, key })} />
                 ))}
               </div>
             ))}
+            {!props.expanded && <div className={styles.editorFade} />}
```
danethurber, line 46: ❤️

docs/src/components/code-block/examples.tsx
```diff
@@ -30,25 +30,23 @@ export interface ExampleData {
...
-  language: Language
-  startExpanded: boolean
 }
 export const ExamplesSwitcher: React.FC<ExamplesSwitcherProps> = props => {
-  const { examples, language, className, startExpanded, ...rest } = props
-
-  const [selectedOption, setSelectedOption] = React.useState(examples[0].value)
+  const [selectedOption, setSelectedOption] = React.useState(
+    props.examples[0].value
+  )
```
danethurber, line 37:
```suggestion
  const [selectedOption, setSelectedOption] = useState(
    props.examples[0].value
  )
```

nitpick...just because we already import it and use it this way

docs/src/components/code-block/examples.tsx
```diff
@@ -30,25 +30,23 @@ export interface ExampleData {
 
 interface ExamplesSwitcherProps extends HTMLAttributes<HTMLDivElement> {
```
danethurber, line 31: similar comment about props and not need to extend the element attributes if we aren't going to pass them to the underlying dom node

docs/src/components/code-block/index.tsx
```diff
@@ -6,15 +6,24 @@ import { Language } from 'prism-react-renderer'
 import { ExamplesSwitcher, Example, parseCode } from './examples'
```
danethurber, line 6:
```suggestion
import React from 'react'
import { ExamplesSwitcher, Example, parseCode } from './examples'
```

docs/src/components/code-block/index.tsx
```diff
@@ -6,15 +6,24 @@ import { Language } from 'prism-react-renderer'
...
 import styles from './styles.module.css'
 
+interface CodeBlockContextValue {
+  language: Language
+  noRender: boolean
+  startExpanded: boolean
+}
+export const CodeBlockContext = React.createContext<CodeBlockContextValue>({
+  language: 'typescript',
```
danethurber, line 15: i'm on the fence about defaulting to typescript...thoughts on defaulting to undefined

danethurber (changes requested):
couple small things

docs/src/components/code-block/index.tsx
jaketrent: Yeah, I'm not sure what to default to either.  Why not typescript?

jaketrent: One reason not to do undefined is that prism-react-rendered#Highlight requires a Language, at least in typings. Have some default from the list would keep typings easy/correct.

```
  type Language =
    | "markup"
    | "bash"
    | "clike"
    | "c"
    | "cpp"
    | "css"
    | "javascript"
    | "jsx"
    | "coffeescript"
    | "actionscript"
    | "css-extr"
    | "diff"
    | "git"
    | "go"
    | "graphql"
    | "handlebars"
    | "json"
    | "less"
    | "makefile"
    | "markdown"
    | "objectivec"
    | "ocaml"
    | "python"
    | "reason"
    | "sass"
    | "scss"
    | "sql"
    | "stylus"
    | "tsx"
    | "typescript"
    | "wasm"
    | "yaml";
```

danethurber: ah....i see. well i don't know then. typescript seems fine

docs/src/components/code-block/index.tsx
danethurber: this is still needed i think

danethurber (approved):
just on react import needed - then merge away :shipit: 

also sorry for the back merge 😬

docs/src/components/code-block/index.tsx
jaketrent: React is still imported. It's on line 3.

PR #1200, opened 2020-09-21
add theme docs

danethurber:
theme docs

### DIFF
--- a/docs/src/pages/components/theme.mdx
+++ b/docs/src/pages/components/theme.mdx
@@ -1,0 +1,56 @@
+---
+name: Theme
+route: /components/theme
+---
+
+import Theme from '@pluralsight/ps-design-system-theme'
+
+import { Types, Usage } from '../../components/component'
+
+# Theme
+
+## Usage pattern
+
+To theme a component or part of the page, surround it in the `Theme` component.
+If you're on React < 16, you'll need to wrap a single child, so a child `div`
+may be required.
+
+```typescript noRender startExpanded
+import Theme from '@pluralsight/ps-design-system-theme'
+import React from 'react'
+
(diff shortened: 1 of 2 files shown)


docs/src/pages/components/theme.mdx
```diff
@@ -0,0 +1,56 @@
...
+route: /components/theme
+---
+
+import Theme from '@pluralsight/ps-design-system-theme'
+
+import { Types, Usage } from '../../components/component'
+
+# Theme
+
+## Usage pattern
```
jaketrent, line 12:
```suggestion
## Examples
```

Not sure this is necessarily required.  It's just the heading that we've used as a standard so far.

docs/src/pages/components/theme.mdx
```diff
@@ -0,0 +1,56 @@
...
+
+To theme a component or part of the page, surround it in the `Theme` component.
+If you're on React < 16, you'll need to wrap a single child, so a child `div`
+may be required.
```
jaketrent, line 16:
```suggestion
```

Does this apply to anyone any more?  Maybe remove?

docs/src/pages/components/theme.mdx
```diff
@@ -0,0 +1,56 @@
...
+
+```typescript noRender startExpanded
```
jaketrent, line 18: Why `startExpanded`?  It's long.  What makes this one special?

jaketrent (approved):

docs/src/pages/components/theme.mdx
danethurber: there isn't a preview and it's basically the only thing page. seems like this codeblock is the most important thing going on so i didn't want to hide it

Issue #1212, opened 2020-09-22
docs: support markdown in Intro


jaketrent:
maybe other places?

PR #1221, opened 2020-09-23
feat/dialogdocs

jaketrent:
Tried to use a portal from util to render the modal example. Didn't work for some import bug that I never solved.  If you're willing, please check out that commit and try to help me figure out what I'm missing.


docs/src/pages/components/dialog.mdx
```diff
@@ -0,0 +1,373 @@
...
+import * as Text from '@pluralsight/ps-design-system-text'
+
+import { Guideline, Intro, Types, Usage } from '../../components/component'
+import { ContentGridVisual, ModalGuidelineExample } from '../../components/dialog'
+
+# Dialog
+
+<Intro>
+  The purpose of a Dialog is to provide actionable messaging and may appear contextually or as a [modal](#modal). The Dialog adapts to various amounts and types of content. For contextual, non-actionable messaging, consider the [Tooltip](/components/tooltip) instead.
+</Intro>
```
danethurber, line 18: `Intro` doesn't use mdx and doesn't render the links

docs/src/pages/components/dialog.mdx
```diff
@@ -0,0 +1,373 @@
...
+```typescript 
+import Button from '@pluralsight/ps-design-system-button'
+import Dialog from '@pluralsight/ps-design-system-dialog'
+import { P } from '@pluralsight/ps-design-system-text'
+import React from 'react'
+
+function Example() {
+  return <div className="example-grid--col-2">
+    <Dialog disableFocusOnMount><ExampleContent /></Dialog>
+    <Dialog tailPosition={Dialog.tailPositions.topCenter} disableFocusOnMount><ExampleContent /></Dialog>
```
danethurber, line 100: you can't really see the tail on light theme...probably not a big deal

docs/src/pages/components/dialog.mdx
```diff
@@ -0,0 +1,373 @@
...
+
+## Guidelines
+
+In most cases, dialogs can assume the user to digest the content following a Z-Pattern. Readers will start in the top/left, move horizontally to the top/right and then diagonally to the bottom/left before finishing with another horizontal movement to the bottom/right. By that measure, place your actions at the bottom/right, with the primary action on the right.
+      
+<Guideline
+  do={
+    <Dialog disableFocusOnMount>
+      <Text.Heading>
+        <h3>Title</h3>
```
danethurber, line 199: all the `h3` tags are getting the `.h3` style from the mdx components and it's giving them a bunch of extra margin. no idea why

danethurber (changes requested):
couple comments. i'll look into the portal stuff

docs/src/pages/components/dialog.mdx
jaketrent: Correct.  I think we should make it do so.  Filed this yesterday. 

https://github.com/pluralsight/design-system/issues/1212

Since we're not live, I'm ok merging one before the other, as long as the work is captured.  You?

docs/src/pages/components/dialog.mdx
jaketrent: No, you can not.  I'm not sure what to do about that. Thought about locking dark theme. Decided against it, since that may mask a design issue in the Dialog instead of expose it.

danethurber (approved):

PR #1228, opened 2020-09-24
Fix dialog intro

jaketrent:
Went down the mdx-js/react road, but figured it wasn't worth it at the moment. Probably possible, but I wasn't patient enough.

### DIFF
--- a/docs/src/components/mdx/index.tsx
+++ b/docs/src/components/mdx/index.tsx
@@ -14,7 +14,7 @@ export const A: React.FC<HTMLAttributes<HTMLAnchorElement>> = props => {
   return (
     <DSLink>
       {isExternal ? (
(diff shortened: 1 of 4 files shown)


docs/src/components/mdx/index.tsx
```diff
@@ -1,18 +1,26 @@
...
-  <Link>
-    <a {...props} />
-  </Link>
-)
+export const A: React.FC<HTMLAttributes<HTMLAnchorElement>> = props => {
+  const isExternal = /http/.test(props.href)
+  return (
+    <DSLink>
+      {isExternal ? (
+        <a {...props} />
```
danethurber, line 17: rel noopener noreferrer ??

danethurber (approved):

PR #1239, opened 2020-09-25
Docs/styling tweaks

sweeting:


EdwardIrby (approved):

docs/src/components/types/prop.tsx
```diff
@@ -29,13 +29,13 @@ export const Prop: React.FC<PropProps> = props => {
         </Table.Cell>
 
         <Table.Cell>
-          <Text.Code>{props.type}</Text.Code>
```
EdwardIrby, line 32: Why the switch here?

sweeting: The styling applied by `Text.Code` looks clunky and this looks closer to the current docs prop table.
image

PR #1241, opened 2020-09-25
refactor(docs): props table change name, type, and default colors based on theme

sweeting:


docs/src/components/types/prop.tsx
```diff
@@ -13,11 +15,20 @@ interface PropProps {
...
+  const themeName = useTheme()
   return (
     <>
       <Table.Row>
         <Table.Cell>
-          <code className={styles.nameCode}>{props.name}</code>
+          <code
+            className={styles.nameCode}
+            style={{
+              '--color':
```
jaketrent, line 26: ahhh yeah, digging the cssvar approach.

jaketrent (approved):

docs/src/components/types/prop.tsx
sweeting: That was all @EdwardIrby's wizardry.

PR #1243, opened 2020-09-25
refactor(docs): programmatically query pages

jaketrent:
- This is the first (big) step in the TOC creation
- Just want to make sure ya'll are ok with this change first. And to minimize the time in a git branch with this big movement.
- Other thing this will allow that I'm excited about will be to co-locate the .mdx pages and the supporting .tsx pages.


docs/package.json
```diff
@@ -82,6 +82,9 @@
     "express": "^4.17.1",
     "gatsby-plugin-express": "^1.1.6",
     "gatsby-plugin-react-helmet": "^3.3.10",
+    "gatsby-remark-autolink-headers": "^2.3.13",
+    "gatsby-remark-table-of-contents": "^0.1.1",
+    "gatsby-transformer-remark": "^2.8.35",
     "helmet": "^4.1.0",
```
danethurber, line 88: i'm not seeing this packages being used in the gatsby config

danethurber (approved):
👍

docs/package.json
jaketrent: agh, they aren't.  relics of trying other things

PR #1244, opened 2020-09-25
Patch/doc build

danethurber:
there were a few issues with the docker build that i tracked down


docs/src/components/code-block/preview.tsx
```diff
@@ -1,6 +1,7 @@
 import { transform } from '@babel/core'
 import Theme from '@pluralsight/ps-design-system-theme'
 import cx from 'classnames'
+import { canUseDOM } from 'exenv'
```
jaketrent, line 4: this is now in the util pkg.

jaketrent (approved):

EdwardIrby (approved):

PR #1245, opened 2020-09-25
TOC on Friday, wat!

jaketrent:
- A little stinky to have to write it in every mdx file, but that's to allow for inline placement per the mobile design
- But inline is nice, because it allows custom placement
- Also learned that full component imports can be done with shortcodes


danethurber (commented):
i'm getting this error. i've done a gatsby clean and am still getting the error

```
 ERROR #98124  WEBPACK

Generating SSR bundle failed

Can't resolve '../../../../src/components/component' in '/Users/dane/Projects/design-system/docs/.cache/caches/gatsby-plugin-mdx/mdx-scopes-dir'

If you're trying to use a package make sure that '../../../../src/components/component' is installed. If you're trying to use a local file make sure that the path is correct.
```

docs/src/components/mdx/index.tsx
```diff
@@ -3,13 +3,18 @@ import DSLink from '@pluralsight/ps-design-system-link'
 import * as Text from '@pluralsight/ps-design-system-text'
 import cx from 'classnames'
 import { Link as GatsbyLink } from 'gatsby'
-import React, { HTMLAttributes } from 'react'
+import React, { AllHTMLAttributes, HTMLAttributes } from 'react'
```
danethurber, line 6: what's the diff between AllHTMLAttributes and HTMLAttributes

docs/src/components/table-of-contents.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+interface TableOfContentsProps {
+  headings: Heading[]
+}
+export const TableOfContents: React.FC<TableOfContentsProps> = props => {
+  const themeName = useTheme()
+  let slugger = new GithubSlugger()
+  React.useEffect(() => {
+    slugger = new GithubSlugger()
+  }, [slugger])
```
danethurber, line 26: you could use a ref for this but i'm not sure it matters

docs/src/components/table-of-contents.tsx
```diff
@@ -0,0 +1,70 @@
...
+  return (
+    <nav
+      className={styles.toc}
+      style={
+        {
+          '--bg':
+            themeName === Theme.names.dark
+              ? colorsBackgroundUtility[25]
+              : colorsBackgroundLight[2]
+        } as React.CSSProperties
```
danethurber, line 36: is this a new pattern we should follow. using css vars instead of toggling classes

jaketrent: Oh, I'm not sure.  I was just geeking out about it.  Had seen it in the wild but never in a project like ours until types/prop.tsx was updated.

I think I still prefer class toggling.  Keeps the core out of the code and the style info in a stylesheet.

Do you think we should consolidate approaches?

docs/src/components/table-of-contents.tsx
jaketrent: Seems to work as-is.  I avoid refs if I can.

docs/src/components/mdx/index.tsx
jaketrent: One screams about not having `href`, and the other one's ok.

docs/src/components/table-of-contents.tsx
danethurber: personally i'm not to worried about consolidating. let's see how it goes

danethurber (changes requested):
i'm having issues on the component pages - they aren't loading

also i just want to double check this is the TOC experience on mobile that we want...

Screen Shot 2020-09-28 at 11 34 22 AM

or do we have a mobile optimized design

jaketrent:
@danethurber I double-checked the mobile design and made one adjustment.

Regarding the pages not working, it seems as thought Types is not available through the MDXProvider#components shortcodes.  The amazing thing is that this has been working twice before but I've somehow bungled it twice since then. :) Anyway, I'll followup on the Types provider problem.

jaketrent:
Hey @danethurber , I think the 2 issues are addressed:

- Mobile design dialed
- Types is now available as an mdx shortcut. Had to change naming convention away from the dot naming.  Eg, Types.Table becomes TypesTable.

Please re-review.

danethurber (approved):
do it live

PR #1252, opened 2020-09-29
Docs home

jaketrent:
Welcome back, robot friend.


EdwardIrby (changes requested):
Resolve error appearing.

I delete `public` and `.cache` dirs and ran `yarn` and still see this.

Screen Shot 2020-09-29 at 11 22 16 AM

jaketrent:
@EdwardIrby Hey, yeah, switched back to branch, `gastby clean`, `yarn`, then `yarn start:dev`. I don't see the error, myself.

What is the error, exactly? "Object is not a function"? On line 96?  I don't quite understand what that means.

The only thing that it indicates to me is that maybe the only function that I was calling, `canUseDOM()`, isn't really a function.  I double checked, and it is.  Which means our code-block/preview.tsx file needs update to use it as a function. 

🤷 Not really sure where to go from here.

danethurber:
i can run it. @EdwardIrby i think your getting that error because the util package was changed but you probably haven't rebuilt it yet.....maybe?

docs/src/components/mdx/index.tsx
```diff
@@ -15,7 +15,7 @@ import { Usage } from '../usage'
 import * as styles from './index.module.css'
 
 export const A: React.FC<AllHTMLAttributes<HTMLAnchorElement>> = props => {
-  const isExternal = props.href.includes('http')
+  const isExternal = /[a-z]+:\/\//.test(props.href)
```
danethurber, line 18: i feel like this line keep evolving and getting stronger. it's gaining sentience...

danethurber (approved):

docs/src/components/mdx/index.tsx
jaketrent: Yes!  The next level is when I paste in something from the highest-voted StackOverflow answer.

PR #1256, opened 2020-09-29
Feat/typescript actionmenu

EdwardIrby:
- ActionMenu conversion to typescript
- test seem to still work without conversion. I'm not really looking to type forward ref in a test. That's just overkill. 1


packages/actionmenu/src/react/ellipsis.tsx
```diff
@@ -1,11 +1,11 @@
 import { css } from 'glamor'
-import React, { forwardRef } from 'react'
+import React, { HTMLAttributes, forwardRef } from 'react'
 import filterReactProps from '@pluralsight/ps-design-system-filter-react-props'
```
danethurber, line 3: i think we can get rid of filterReactProp in most places because the typings should tell us when unsupported props are passed. there might be a couple places that's not the case...but snapshots should alert us and we can remove the props without the filterReactProps helper

packages/actionmenu/src/react/icon.tsx
```diff
@@ -11,12 +10,13 @@ const styles = ({ marginLeft, marginRight }) =>
...
 
-export const Icon = ({ marginLeft, marginRight, ...props }) => (
+interface Props extends HTMLAttributes<HTMLSpanElement> {
+  marginRight: boolean;
+  marginLeft: boolean;
+}
+
+export const Icon: React.FC<Props> = ({ marginLeft, marginRight, ...props }) => (
   <span {...filterReactProps(props)} {...styles({ marginLeft, marginRight })} />
 )
```
danethurber, line 20:
```suggestion
export const Icon: React.FC<Props> = ({ marginLeft, marginRight, ...rest }) => (
  <span {...rest} {...styles({ marginLeft, marginRight })} />
)
```

danethurber: an example of using types to ensure only props that are supported are passed - and we filter out the props that are not supported on the el. we can use this pattern to remove our need for filterReactProps

packages/actionmenu/src/react/index.tsx
```diff
@@ -0,0 +1,102 @@
...
+  Item: typeof Item
+  Ellipsis: typeof Ellipsis
+  Icon: typeof Icon
+  origins: typeof vars.origins
+  tagName: typeof vars.tagName
+  useMenuRef: typeof useMenuRef
+}
+
+interface Props extends HTMLAttributes<HTMLUListElement> {
+  children: React.ReactNode | Array<React.ReactNode>;
```
danethurber, line 41: i don't think you need to define the children props. it should just be there with React.FC

packages/actionmenu/src/react/index.tsx
```diff
@@ -0,0 +1,102 @@
...
+  onClose: () => void;
+  origin: string;
```
danethurber, line 43: we can provide more helpful information to consumers in these types by limiting the strings to only the available vars

```suggestion
  origin: ValueOf<typeof string>;
```

`ValueOf` is in the utils package

danethurber (changes requested):
- carousel snapshots need updated
- couple linter errors...i believe they are all fixable with `yarn lint:fix` but they should have been cause in the precommit so i'm not sure

danethurber (approved):

PR #1257, opened 2020-09-30
Alternative approach to examples template selection

jaketrent:
Just an idea. I like it better because it's not wrapped up in frame and doesn't use the window.location. And the *template* switching is where templates are first chosen, in createPage().

### DIFF
--- a/docs/gatsby-node.js
+++ b/docs/gatsby-node.js
@@ -49,7 +49,7 @@ exports.createPages = async ({ graphql, actions }) => {
   result.data.allMdx.edges.forEach(({ node }) => {
     createPage({
       path: node.fields.slug,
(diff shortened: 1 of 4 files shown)


docs/gatsby-node.js
```diff
@@ -58,3 +58,9 @@ exports.createPages = async ({ graphql, actions }) => {
     })
   })
 }
+
+function chooseTemplate(slug) {
+  return slug.includes('examples')
+    ? path.resolve(`./src/templates/example-frame.tsx`)
+    : path.resolve(`./src/templates/page.tsx`)
+}
```
danethurber, line 66: yes much better

danethurber (commented):
would we still use an iframe to load the example?

PR #1260, opened 2020-09-30
viewtoggle: convert to typescript

danethurber:
viewtoggle: convert to typescript


packages/viewtoggle/package.json
```diff
@@ -9,21 +9,22 @@
...
   "scripts": {
-    "build": "run-s build:js build:cjs build:css",
-    "build:cjs": "babel --root-mode upward src --out-dir dist/cjs",
+    "build": "run-s build:svg build:cjs build:esm build:css copy:svg",
+    "build:cjs": "tsc --project tsconfig.build.json --module commonjs --target es5 --outDir dist/cjs",
     "build:css": "build-css --useGlamor -i dist/cjs/css/index.js",
-    "build:js": "cross-env ESM=true babel --root-mode upward src --out-dir dist/esm",
-    "build:watch": "npm run build:js -- --watch",
+    "build:esm": "tsc --project tsconfig.build.json --module es2015 --target es5 --outDir dist/esm",
+    "build:svg": "svgr --typescript --filename-case kebab --ext \"dist.tsx\" -d src/react/illustrations src/svg",
```
jaketrent, line 16: Looks like a copy-pasta mistakenelli to me. This package doesn't have svg.

packages/viewtoggle/src/react/__specs__/index.spec.tsx
```diff
@@ -1,7 +1,7 @@
 import React from 'react'
 import { render } from '@testing-library/react'
 
-import ViewToggle from '../index.js'
```
jaketrent, line 4: hehe, remember when we went through all the work to do the new/non-magic module resolution way?

packages/viewtoggle/src/react/__stories__/index.story.tsx
```diff
@@ -4,28 +4,28 @@ import { storiesOf } from '@storybook/react'
 import { css } from 'glamor'
 import React from 'react'
 
-import ViewToggle from '../index.js'
+import ViewToggle from '..'
 
 storiesOf('options count', module)
-  .add('one', _ => (
```
jaketrent, line 10: sniff, I always liked this style, but I too have done away with it.

packages/viewtoggle/src/react/index.tsx
```diff
@@ -0,0 +1,212 @@
+import Theme, { useTheme } from '@pluralsight/ps-design-system-theme'
```
jaketrent, line 1: Wish these were seen as renames in git with internal adjustments instead of net new.  We're going to lose a lot of first-level git blame history that's the easiest to investigate.

jaketrent: It's also more difficult for me to review.

jaketrent (changes requested):
Can't build because of the build copy-pasta issue, but I will re-test once that's updated.

packages/viewtoggle/package.json
danethurber: thanks. yes i copied from emptystate 😬

packages/viewtoggle/src/react/index.tsx
danethurber: totally agree. i'm going to do some digging to see if there is a better way. maybe rename the file and commit. then maybe the changes...

packages/viewtoggle/src/react/__specs__/__snapshots__/storyshots.spec.ts.snap
```diff
@@ -10,11 +10,11 @@ exports[`Storyshots active default first 1`] = `
     data-css-w8tgx1=""
     style={Object {}}
   >
-    <span
+    <div
```
jaketrent, line 13: This indicates that the render of this component was changed.  But I have a really bad diff, so I can't tell where or why easily.

jaketrent (approved):
tsc, build, test, storybook look good.  I'm yolo'ing on the changes in the react component though. I'm just not patient enough to do a manual visual diff against previous file verisons.

PR #1266, opened 2020-10-01
Docs/tweaks

sweeting:
A smorgasboard of doc tweaks. I'll try and encapsulate these better in the future.


docs/src/components/frame/frame.module.css
```diff
@@ -45,5 +45,17 @@
   flex: 0 1 auto;
   margin: 0 auto;
   max-width: 1200px;
-  padding: var(--psLayoutSpacingXLarge);
+  padding: var(--psLayoutSpacingMedium) var(--psLayoutSpacingXLarge);
+}
+
+.main h1 {
```
jaketrent, line 51: I'm feeling like we should do this through the MDXProvider components instead.

jaketrent: Just thinking that would make sure that everywhere we wrote mdx and used headings, we'd get those styles.

docs/src/content/index.module.css
```diff
@@ -37,7 +41,7 @@
   background-image: var(--robotImg);
   overflow: hidden;
   height: 0;
-  padding-top: calc(656/960*100%);
+  padding-top: calc(656 / 960 * 100%);
```
jaketrent, line 44: I'd consider using css vars to identify where these magic numbers came from.

docs/src/components/frame/frame.module.css
jaketrent: This kind of thing (other headings in the .main area) might be another reason to avoid this approach.

Screen Shot 2020-10-01 at 9 41 58 AM

docs/src/content/index.module.css
```diff
@@ -37,7 +42,7 @@
...
```
jaketrent, line 45: We made a change here pairing to add the vars which must not have made it into the commit.

jaketrent (approved):

docs/src/content/index.module.css
sweeting: File saved and committed. 😄

PR #1276, opened 2020-10-01
Replace packages-api

jaketrent:
- Supply the latest package version from npm
- At build time - In gatsby-node.js

### DIFF
--- a/docs/gatsby-node.js
+++ b/docs/gatsby-node.js
@@ -48,7 +48,6 @@ exports.createPages = async ({ graphql, actions }) => {
   `)
 
   const npmVersions = await findAllNpmPackages()
-  console.log({ npmVersions })
 
   result.data.allMdx.edges.forEach(({ node }) => {
     createPage({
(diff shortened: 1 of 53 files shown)


docs/gatsby-node.js
```diff
@@ -46,21 +47,84 @@ exports.createPages = async ({ graphql, actions }) => {
...
+
+async function findAllNpmPackages() {
+  try {
+    const list = await searchNpm()
+    return list.reduce((acc, searchResult) => {
+      acc[searchResult.name] = searchResult.version
+      return acc
+    }, {})
+  } catch (err) {
+    console.log('packages-api#findAll error', err)
```
danethurber, line 127: why not just throw the error?

danethurber (approved):
love it. nice work

docs/gatsby-node.js
jaketrent: Because I feel like this is "nice to have" data and also prone to not be super reliable, and I don't want to disrupt the build if it fails.

PR #1277, opened 2020-10-01
refactor(docs): yarn upgrade deps

jaketrent:
- Saw some secvuln alerts. Trying to clear those up.


docs/package.json
```diff
@@ -76,9 +76,22 @@
     "@pluralsight/ps-design-system-util": "^5.2.0",
     "@pluralsight/ps-design-system-verticaltabs": "^4.1.0",
     "@pluralsight/ps-design-system-viewtoggle": "^6.0.17",
+    "@storybook/addon-actions": "^6.0.21",
```
jaketrent, line 79: If I did a `cd docs/; yarn upgrade`, why would I have these lines *added*?  I'd think there'd just be modified lines.

danethurber: i think they were in the root package.json. i'm not sure either why it put them here

jaketrent: maybe I should run yarn upgrade in root instead/first?

jaketrent:
replaced with #1283.

PR #1280, opened 2020-10-02
feat: button link ts conversion

EdwardIrby:
ts conversion link and button


packages/button/src/react/index.tsx
```diff
@@ -0,0 +1,237 @@
...
+    }
+
+    const isDisabledLink = allProps.disabled && allProps.href
+    const filteredProps = filterReactProps(props, { tagName })
+    delete filteredProps.icon
+    if (isDisabledLink) {
+      delete filteredProps.onClick
+    }
+
+    return React.createElement(
```
danethurber, line 175: i wonder why we were using createElement....guessing this was one of the first components and maybe we didn't support jsx then. just thinking out load - don't change anything for this pr but i might follow up with a conversion to jsx...

packages/button/src/react/index.tsx
```diff
@@ -0,0 +1,237 @@
...
+  appearance: PropTypes.oneOf(Object.keys(vars.appearances)),
+  children: PropTypes.any,
+  disabled: PropTypes.bool,
+  href: PropTypes.string,
+  icon: PropTypes.element,
+  iconAlign: PropTypes.oneOf(Object.keys(vars.iconAligns)),
+  loading: PropTypes.bool,
+  size: PropTypes.oneOf(Object.keys(vars.sizes)),
+  style: PropTypes.object
+}
```
danethurber, line 221:
```suggestion
Button.propTypes = {
  appearance: PropTypes.oneOf(Object.keys(vars.appearances)),
  children: PropTypes.any,
  disabled: PropTypes.bool,
  href: PropTypes.string,
  icon: PropTypes.element,
  iconAlign: PropTypes.oneOf(Object.keys(vars.iconAligns)),
  loading: PropTypes.bool,
  size: PropTypes.oneOf(Object.keys(vars.sizes)),
  style: PropTypes.object
}
```

danethurber: i don't think we need to add proptypes anymore. if we want to support this in the future we should generate them from the ts typings

danethurber (approved):

PR #1284, opened 2020-10-02
docs: adds docusearch

EdwardIrby:
install docusearch


jaketrent (approved):
Well, I guess all that gatsby overengineer pays off here! :) Lol, nice plugin!

Not sure how much stylistic customization we can do, but it's looking a little odd in the popup:

Screen Shot 2020-10-02 at 1 39 08 PM

PR #1285, opened 2020-10-02
Daaaaark mode in typography

jaketrent:
:dracula:

### DIFF
--- a/docs/src/components/box.tsx
+++ b/docs/src/components/box.tsx
@@ -1,0 +1,25 @@
+import Theme, { useTheme } from '@pluralsight/ps-design-system-theme'
+import cx from 'classnames'
+import * as React from 'react'
+
+import * as styles from './box.module.css'
+
+export const Box: React.FC<HTMLDivElement> = props => {
+  const { className, ...rest } = props
+  const themeName = useTheme()
+  console.log({ themeName, styles })
+
+  return (
+    <div
+      className={cx(
+        {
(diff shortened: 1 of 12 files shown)


docs/src/components/box.tsx
```diff
@@ -0,0 +1,24 @@
+import Theme, { useTheme } from '@pluralsight/ps-design-system-theme'
+import cx from 'classnames'
+import * as React from 'react'
```
danethurber, line 3:
```suggestion
import React from 'react'
```

🤷‍♂️

docs/src/components/typography/monospace-font.module.css
```diff
@@ -13,13 +16,14 @@
   text-align: center;
 }
 .fontLetter {
+  composes: glyphs;
```
danethurber, line 19: 🍾

danethurber (approved):

docs/src/components/box.tsx
jaketrent: Apparently the `* as` is going to be the only supported import style in a later react version.

https://twitter.com/sebmarkbage/status/1250285076014747648

docs/src/components/typography/monospace-font.module.css
jaketrent: Just thought I'd introduce another class composition pattern into this codebase. :)

Lol, I'm not sure what your emoji means.

PR #1286, opened 2020-10-02
convert VerticalTabs to typescript

danethurber:


jaketrent (commented):
tsc and yarn build had errors in the package root dir.  From the looks of them, I thought maybe not all dependent packages were build (crying about `omit`). I went to DS root and ran yarn build, and still got an error on the types:

```
lerna ERR! yarn run build exited 1 in '@pluralsight/ps-design-system-verticaltabs'
lerna ERR! yarn run build stdout:
$ run-s build:cjs build:esm build:css
$ tsc --project tsconfig.build.json --module commonjs --target es5 --outDir dist/cjs
src/react/item.tsx(203,21): error TS2345: Argument of type 'PropsWithChildren<Tier2HeaderProps>' is not assignable to parameter of type 'Record<string, unknown>'.
  Index signature is missing in type 'Tier2HeaderProps & { children?: ReactNode; }'.
```

Do you get the same?

danethurber:
looks like ci is getting the error as well...i'll look into it

jaketrent (approved):

PR #1293, opened 2020-10-05
Badge tsc

jaketrent:
I started this branch trying to prove that we could rename files in this TS conversion.  Steps to accomplish:

1. mv, then commit
2. convter, then commit

I want to add the `strict: true` compiler option, but that whines about theme package not having typings, and I'd like to use those typings in the other strict errors (eg, `any` params that need `Theme.names`).

I think that all the packages should have `strict: true`.  We will have to circle back and get those that we missed.


danethurber (commented):
- build is currently not outputting anything and there are errors. if you look at one of the other packages(avatar?) you'll see that there are new build commands and i've dropped babel in those builds.
- i've been using two tsconfig files. one for general editor setup and another that use used during build. i noticed you've just got a single. i'd love for that approach to work but it's not currently

danethurber (commented):
`packages/badge/src/index.ts ` has some imports that need updated to remove the `.js`

packages/badge/src/react/index.tsx
```diff
@@ -1,13 +1,15 @@
-import { css } from 'glamor'
-import React from 'react'
-import PropTypes from 'prop-types'
-
 import filterReactProps from '@pluralsight/ps-design-system-filter-react-props'
```
danethurber, line 1: i've been removing filterReactProps hoping that with static types we no longer need to be defensive....i think that is safe for most needs and definitely should be ok for this simple component

jaketrent: Yes, great catch. Thanks!

danethurber (commented):
in the vars file you could use the keyMirror. it's basically the same thing but takes less code and is readonly which I like for vars

```
import { keyMirror } from '@pluralsight/ps-design-system-util'

export const appearances = keyMirror('default', 'subtle')

export const colors = keyMirror('gray', 'green', 'yellow', 'red', 'blue')
```

I think there were other typescript-y benefits but i forgot...it's a pattern i stole from gitprime...

danethurber:
i think `strict: true` is a great goal. i don't know what it will take but let's plan on it

jaketrent:
@danethurber thanks for the thorough review. I think I've addressed everything you mentioned.

danethurber (approved):

PR #1296, opened 2020-10-05
feat(docs): add css ssr plugins

### LINKED ISSUE #1272: docs: fix styles flash in ssr

danethurber:
there are a few issues with the initial rendering but looking at master i think they are caused by something else. 

resolves #1272


jaketrent (approved):
Looks great locally.

PR #1301, opened 2020-10-06
fix(docs): single line commits to fix transform in preview

jaketrent:
wheeeee!


danethurber (approved):
haha. forgot about this...

PR #1307, opened 2020-10-07
Move theming to app level

jaketrent:
- Should help us avoid doing .dark and .light classes on low level components for basic things
- Still exists where it doesn't make sense to create app-level variables
- Relies on css vars being available. Chosee frame.tsx to provide it since it also has the Theme.Provider

### DIFF
--- a/docs/src/components/frame/frame.module.css
+++ b/docs/src/components/frame/frame.module.css
@@ -34,6 +34,7 @@
 .dark {
   background: var(--psColorsBackgroundDark1);
 
+  --appColorsBackground2: var(--psColorsBackgroundDark2);
   --appColorsTextIconHigh: var(--psColorsTextIconHighOnDark);
   --appColorsTextIconLow: var(--psColorsTextIconLowOnDark);
 }
@@ -41,6 +42,7 @@
 .light {
   background: var(--psColorsWhite);
 
+  --appColorsBackground2: var(--psColorsBackgroundLight2);
   --appColorsTextIconHigh: var(--psColorsTextIconHighOnLight);
   --appColorsTextIconLow: var(--psColorsTextIconLowOnLight);
 }
--- a/docs/src/components/frame/frame.module.css
+++ b/docs/src/components/frame/frame.module.css
@@ -34,7 +34,11 @@
 .dark {
   background: var(--psColorsBackgroundDark1);
 
-  --appColorsBackground2: var(--psColorsBackgroundDark2);
+  --appColorsBackgroundLo: var(--psColorsBackgroundDark1);
+  --appColorsBackgroundMed: var(--psColorsBackgroundDark2);
+  --appColorsBackgroundHi: var(--psColorsBackgroundDark3);
+  --appColorsBorderLow: var(--psColorsBorderLowOnDark);
+  --appColorsBorderHigh: var(--psColorsBorderHighOnDark);
   --appColorsTextIconHigh: var(--psColorsTextIconHighOnDark);
   --appColorsTextIconLow: var(--psColorsTextIconLowOnDark);
 }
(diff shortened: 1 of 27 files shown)


docs/src/components/guidelines/guideline.module.css
```diff
@@ -8,15 +8,8 @@
...
   padding: var(--psLayoutSpacingXLarge);
   position: relative;
-
-  &.dark {
-    background: var(--psColorsBackgroundDark3);
-    color: var(--psColorsTextIconHighOnDark);
-  }
-
-  &.light {
-    background: var(--psColorsBackgroundLight2);
```
EdwardIrby, line 18: So this isn't a 1:1 switch it's both `appColorsBackgroundHi` and `appColorsBackgroundMed`

docs/src/components/motion/speeds.module.css
```diff
@@ -1,14 +1,14 @@
...
-  background: var(--psColorsBackgroundLight2);
+  background: var(--appColorsBackgroundMed);
   padding: var(--psLayoutSpacingMedium);
 }
 .speed {
   display: grid;
   grid-template-columns: 64px 1fr;
   grid-gap: var(--psLayoutSpacingMedium);
   grid-template-areas: 'icon vars' 'labels labels';
-  color: var(--psColorsBackgroundDark2);
```
EdwardIrby, line 11: Was this color wrong to begin with?

docs/src/components/frame/frame.module.css
```diff
@@ -33,10 +33,28 @@
...
+  --appColorsTextIconHigh: var(--psColorsTextIconHighOnDark);
+  --appColorsTextIconLow: var(--psColorsTextIconLowOnDark);
 }
 
 .light {
   background: var(--psColorsWhite);
+
+  --appColorsBackgroundLo: var(--psColorsBackgroundLight3);
+  --appColorsBackgroundMed: var(--psColorsBackgroundLight2);
+  --appColorsBackgroundHi: var(--psColorsBackgroundLight1);
```
EdwardIrby, line 52: Is this the right order? Seemed reversed but I'm not sure.

EdwardIrby (changes requested):
Recommend having @sweeting look at the areas I commented on. Totally unrelated to this pr we've got a broke import on http://localhost:8000/components/layout-page-heading

docs/src/components/frame/frame.module.css
jaketrent: That is a great observation.  I think it at first seems like it's in the wrong order because the dark and the light backgrounds are listed 1, 2, 3 in opposite order from each other. In dark, 1 is most dark. In light, 3 is most light. I think the "most" contrast should be at the same extreme, either 1 or 3, matched in dark and light. To try to be more clear, for the app-level vars, I used lo, med, hi to indicate contrast away from the dark/light, depending.

docs/src/components/motion/speeds.module.css
jaketrent: It's a background color. Semantically, it's just not what we use bg colors for.  It could have been that way in the mockup, probably was. Changed it so that we use textIcon colors for text instead.

jaketrent:
Mentioned broken layout example should be fixed in: https://github.com/pluralsight/design-system/pull/1301/files

docs/src/components/guidelines/guideline.module.css
jaketrent: True enough. I made the call for convenience. And I think it looks great. Happy to change it back if needed.

Screen Shot 2020-10-07 at 10 58 50 AM
Screen Shot 2020-10-07 at 10 58 44 AM

EdwardIrby (approved):

docs/src/components/frame/frame.module.css
sweeting: As far as naming conventions, the `-appColorsBackground` variables use abbreviations: Lo, Med, Hi, where the other variables use the full words: Low, High. Should we use the same convention?

PR #1309, opened 2020-10-07
ActionMenu: Typings

danethurber:
- added a babel config to ActionMenu so storybook can start
- update test deps
- added react testing lib typings
- some additional typings added to ActionMenu


package.json
```diff
@@ -99,7 +99,9 @@
     "@types/react": "^16.9.49",
     "@types/react-dom": "^16.9.8",
     "@types/react-helmet": "^6.1.0",
-    "@types/testing-library__jest-dom": "^5.9.2",
+    "@types/storybook__addon-storyshots": "^5.3.1",
```
danethurber, line 102: this is new. fixes some mis-typing i notices in the tests

packages/actionmenu/.babelrc
```diff
@@ -0,0 +1,3 @@
+{
+  "presets": ["@babel/env", "@babel/react", "@babel/typescript"],
+}
```
danethurber, line 3: needed by storybook - otherwise it can't start

packages/actionmenu/src/react/__specs__/index.spec.tsx
```diff
@@ -19,7 +19,7 @@ describe('ActionMenu', () => {
 
   it('fires onClick on option click', () => {
     return new Promise(resolve => {
-      function handleClick(evt, value) {
+      const handleClick = (_evt: unknown, value: unknown) => {
```
danethurber, line 22: these aren't really needed but it's better for the linter

packages/actionmenu/src/react/item.tsx
```diff
@@ -43,99 +47,107 @@ const styles = {
...
   className?: string
-  origin?: string
-  onClick?: (event: React.MouseEvent, value: string | number) => void
+  disabled?: boolean
+  href?: string
   nested?: React.ReactNode
+  onClick?: (event: React.MouseEvent, value: string | number) => void
+  origin?: string
+  tagName?: string
   value?: string | number
```
danethurber, line 58: sorry i can't help but alphabetize out of habit.

i did add `href` - it's something we missed the first time around but the stories helped find it

packages/actionmenu/src/react/item.tsx
```diff
@@ -43,99 +47,107 @@ const styles = {
...
       ...rest
-    },
-    forwardedRef
-  ) => {
-    const { onClickContext, onClose, originContext } = useContext<{
-      onClickContext: (e: Event, v: string | number) => void
-      onClose: (e: Event, v: string | number) => void
-      originContext: string
-    }>(ActionMenuContext)
+    } = props
```
danethurber, line 72: i moved the destructuring to inside the fn...only because my editor code highlighting can't figure it out otherwise....personal problem...

packages/actionmenu/src/react/item.tsx
```diff
@@ -43,99 +47,107 @@ const styles = {
...
-      tagName,
       ...rest
-    },
-    forwardedRef
-  ) => {
-    const { onClickContext, onClose, originContext } = useContext<{
-      onClickContext: (e: Event, v: string | number) => void
-      onClose: (e: Event, v: string | number) => void
-      originContext: string
-    }>(ActionMenuContext)
```
danethurber, line 75: typings are now on the contextCreation

packages/actionmenu/src/react/item.tsx
```diff
@@ -43,99 +47,107 @@ const styles = {
...
+export const Item = forwardRef<HTMLLIElement, ItemProps>(
+  (props, forwardedRef) => {
+    const {
+      children,
+      className,
       disabled,
       nested,
+      onClick,
       origin,
+      tagName: TagName = tagName.a,
```
danethurber, line 69: the react team has said they are going to get ride of defaultProps. i moved the default into this destructure statement. idk if this is something we should do everywhere...

packages/actionmenu/src/react/item.tsx
```diff
@@ -43,99 +47,107 @@ const styles = {
...
     useImperativeHandle(forwardedRef, () => ref.current)
-    const TagName = tagName
+
+    const subMenuRef = useRef()
+
     const [open, setOpen] = useState(false)
     const hasSubMenu = Boolean(nested)
-    const handleMouseOver = e => {
+
+    const handleMouseOver: MouseEventHandler<HTMLLIElement> = () => {
```
danethurber, line 86: i just provided some more elaborate typings to this events to match the props...also i used the term `evt` instead of `e`. `evt` makes the clean coder part of my brain happier...

packages/actionmenu/src/react/item.tsx
```diff
@@ -43,99 +47,107 @@ const styles = {
...
       }
     }
-    const handleClick = e => {
-      onClick && onClick(e, value)
-      onClickContext && onClickContext(e, value)
-      onClose && onClose(e, value) // : e.currentTarget.parentNode.focus()
+    const handleClick: MouseEventHandler<HTMLLIElement> = evt => {
+      onClick && onClick(evt, value)
+      onClickContext && onClickContext(evt, value)
+      onClose && onClose(evt, value) // : e.currentTarget.parentNode.focus()
```
danethurber, line 117: i noticed this comment....did we miss something that we need to fix?

packages/actionmenu/src/react/item.tsx
```diff
@@ -1,21 +1,25 @@
...
   useImperativeHandle,
-  useContext,
-  HTMLAttributes
+  useContext
 } from 'react'
 
-import filterReactProps from '@pluralsight/ps-design-system-filter-react-props'
-import stylesheet from '../css/index'
-import { tagName } from '../vars/index'
+import stylesheet from '../css'
```
jaketrent, line 14:
```suggestion
import * as stylesheet from '../css'
```

packages/actionmenu/src/react/item.tsx
```diff
@@ -1,21 +1,25 @@
...
+import { tagName } from '../vars'
+
 import { ActionMenuContext } from './context'
 import { Arrow } from './arrow'
 
-const styles = {
+type StyleFn = (obj?: unknown) => StyleAttribute
```
jaketrent, line 20:
```suggestion
type StyleFn = (props?: ItemProps) => StyleAttribute
```

Would this not allow you to avoid retyping all the objects (eg, line 23)?

packages/actionmenu/src/react/item.tsx
jaketrent: First I've heard of that going away.  Do you have a url where I can confirm and hear what their reasoning is?

If it is going away, I'm all for replacing React.defaultProps with an alternative default strategy.

packages/actionmenu/src/react/item.tsx
jaketrent: orly? That's surprising. This is very common in our code, including elsewhere in this file. What's special about this case? Also feels odd to make coding decisions based on code highlighting tools, but I can't be sure I haven't done the same thing. :)

packages/actionmenu/src/react/item.tsx
jaketrent: Yeah, I'm usually `evt` too. To really Uncle Bob this though, should probably be `mouseoverEvent` or something. :)

packages/actionmenu/src/react/item.tsx
jaketrent: Not sure. Existed before 574702192b, the TS conversion.

Line was added, including comment, in 0248ef32d4 by @EdwardIrby  in a large component rebuild.  Might want to ask him.

jaketrent (approved):
sweeto burrito

packages/actionmenu/src/react/item.tsx
danethurber: https://github.com/facebook/react/pull/16210

packages/actionmenu/src/react/item.tsx
danethurber: yeah...probably not the best approach from me

packages/actionmenu/src/react/item.tsx
danethurber: well that's a bridge to far... 😄

packages/actionmenu/src/react/item.tsx
jaketrent: If it helps, I think we destructure way too much in our code. I think in some cases it harms readability, requiring code tracing more often than not.  

Reducing destructuring to required cases might be another way to help.

packages/actionmenu/src/react/item.tsx
EdwardIrby: I modded this a bunch in Dropdown conversion so I'll just have to resolve the conflict later. I prever going with `evt: ReactMouiseEvt, value: React.ReactText`

packages/actionmenu/src/react/__specs__/index.spec.tsx
EdwardIrby, line 22: `(evt: React.MouseEvent, value: React.ReactText) => void`

PR #1311, opened 2020-10-08
build: fix issues with running multiple builds

danethurber:
i was getting issues when publishing. during the publish - lerna was running the root level `build` script once for each package. not sure why...so i changed up the process a bit

i'm hopeful this will work...but it's hard to tell with lerna because there isn't a way to verify until we actually run it

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,5 +1,13 @@
 {
   "packages": ["docs", "./packages/*"],
+  "command": {
+    "version": {
+      "allowBranch": "master",
+      "conventionalCommits": true,
+      "ignore": ["docs"],
+      "message": "build: publish"
+    }
+  },
   "version": "independent",
   "npmClient": "yarn",
   "useWorkspaces": true
--- a/package.json
+++ b/package.json
@@ -35,8 +35,8 @@
     "octopus:deploy": "octopus-deploy-create-release-and-deploy",
     "octopus:publish": "gulp octopus-publish",
     "postinstall": "link-parent-bin",
-    "publish:all": "lerna publish --conventional-commits --message \"build: publish\"",
-    "publish:all:beta": "yarn publish -- --npm-tag=beta",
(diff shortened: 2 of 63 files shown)


lerna.json
```diff
@@ -1,5 +1,13 @@
 {
   "packages": ["docs", "./packages/*"],
+  "command": {
+    "version": {
+      "allowBranch": "master",
+      "conventionalCommits": true,
+      "ignore": ["docs"],
+      "message": "build: publish"
+    }
+  },
```
danethurber, line 10: move the config here that i could. running through a yarn script or through the lerna cli should give us similar results

package.json
```diff
@@ -35,8 +35,8 @@
     "octopus:deploy": "octopus-deploy-create-release-and-deploy",
     "octopus:publish": "gulp octopus-publish",
     "postinstall": "link-parent-bin",
-    "publish:all": "lerna publish --conventional-commits --message \"build: publish\"",
-    "publish:all:beta": "yarn publish -- --npm-tag=beta",
+    "publish:all": "yarn build && yarn publish:all --ignore-scripts --ignore-prepublish",
```
danethurber, line 38: this is the main change. we no longer are running `prepublish` in each package. now we run a single build from the root for all packages.

we're also using the `ignore-prepublish` option to ignore any prepublish scripts in the packages

packages/actionmenu/package.json
```diff
@@ -14,7 +14,6 @@
     "build:css": "build-css --useGlamor -i dist/cjs/css/index.js",
     "build:esm": "tsc --project tsconfig.build.json --module es2015 --target es5 --outDir dist/esm",
     "build:watch": "yarn build:esm -- --watch",
-    "prepublish": "yarn build",
```
danethurber, line 17: `prepublish` is deprecate. i think we're supposed to use `prepare` or `prepublishOnly`. i just decided to remove these and build once at the root

jaketrent (approved):
yolo!

PR #1315, opened 2020-10-08
fix(position): update on ref change

danethurber:
i should have just looked at the eslint warnings....spent time debugging when the linter was telling me exactly what was wrong 🤦


EdwardIrby (approved):

jaketrent:
Lint is worth something?!  What was the rule that it flagged?

PR #1318, opened 2020-10-08
refactor(docs): collapse sidenav links to start

jaketrent:


docs/src/components/side-nav/index.tsx
```diff
@@ -25,38 +25,35 @@ export const SideNav: React.FC<Props> = () => {
...
-              })}
+              {items.map(section => (
+                <VerticalTabs.Tier1
+                  header={
+                    <VerticalTabs.Tier1.Header>
+                      {section.header.title}
+                    </VerticalTabs.Tier1.Header>
+                  }
+                  collapsible
+                  collapsed
```
jaketrent, line 36: this is the line added ^

sweeting: @jaketrent This takes care of collapsing by default. The full behavior should be that the expanded/collapsed state of the side nav items should be preserved through page reloads, and if I open a url directly to a page, the section of the navigation it's in should be expanded and the nav item should display the selected state. Would you like me to create a new card for that or address in this PR?

jaketrent: Definitely a new card.  That was not in the original issue.

sweeting (approved):

PR #1334, opened 2020-10-13
CircularProgress TS Convert

jaketrent:
The yarn-script for test has a nice addition for tsc that Justin had put in, that I kept here, and I think would be good to have generally.

Another diff on this one is that `strict: true` in tsconfig.  I think we should do this everywhere as well.


packages/circularprogress/package.json
```diff
@@ -9,20 +9,20 @@
...
+    "build:cjs": "tsc --project tsconfig.build.json --module commonjs --target es5 --outDir dist/cjs",
     "build:css": "build-css --useGlamor -i dist/cjs/css/index.js",
-    "build:js": "cross-env ESM=true babel --root-mode upward src --out-dir dist/esm && cp src/index.d.ts dist/esm/index.d.ts",
+    "build:esm": "tsc --project tsconfig.build.json --module es2015 --target es5 --outDir dist/esm",
     "build:watch": "yarn build:js -- --watch",
     "storybook": "start-storybook -p 6006",
     "test": "tsc --noEmit --project ./tsconfig.json && jest",
-    "test:watch": "yarn test -- --watchAll"
+    "test:watch": "yarn test --watchAll"
   },
```
danethurber, line 20: please add the reference for types

packages/circularprogress/src/react/index.tsx
```diff
@@ -31,34 +37,54 @@ const styles = {
...
+>((props, ref) => {
+  const defaultIndeterminateValue = 25
+  const { size, value = defaultIndeterminateValue, ...rest } = props
   const themeName = useTheme()
 
-  const isIndeterminate = typeof props.value === 'undefined'
-  const value = isIndeterminate ? 25 : props.value
+  /* if (typeof props.value !== 'undefined') { */
+  /*   value = props.value */
+  /* } */
```
danethurber, line 79:
```suggestion

```

packages/circularprogress/package.json
jaketrent: ^ I don't know what you mean.

> please add the reference for types

Please elaborate.  What reference? What types? I'm not intuiting that from the line reference here.

danethurber: we've been adding these two lines to each package. one for types and one for styles just because

```
  "style": "dist/index.css",
  "types": "dist/esm/index.d.ts",
```

EdwardIrby (approved):

PR #1335, opened 2020-10-13
Halo typescript convert

danethurber:
Halo typescript convert


jaketrent (approved):
Looks good.  Made extract adjustments for tsc strict mode in last couple commits.

packages/halo/src/react/index.tsx
```diff
@@ -16,19 +16,12 @@ import * as vars from '../vars'
 
 if (canUseDOM()) polyfillFocusWithin(document)
 
-type StyleFn = (
+export type StyleFn<P> = (
```
jaketrent, line 19: It'd be good to put this in util or somewhere common.

Would allow reuse.

Would also help us to be consistent in the format of these things.

But, maybe we don't have consistent need.  For example, some StyleFn's include state as a param. This one obviously doesn't need that.

packages/halo/src/react/__specs__/storyshots.spec.tsx
```diff
@@ -4,12 +4,12 @@ import initStoryshots, {
...
+  '../__stories__/focusable',
+  () => (props: React.HTMLAttributes<HTMLDivElement>) => <div {...props} />
+)
 
 initStoryshots({
   configPath: path.resolve(__dirname, '../../../.storybook'),
-  framework: 'react',
-  test: snapshotWithOptions({ createNodeMock })
+  framework: 'react'
 })
```
danethurber, line 15: @jaketrent thanks i never know how to get this to work.

also we we not need the node mock anymore?

packages/halo/src/react/index.tsx
danethurber: yeah i'm torn. i'm not sure all our styles fns can be the same - like when we have multiple components in a single file that have different props

packages/halo/src/react/index.tsx
```diff
@@ -0,0 +1,98 @@
...
+import React, { HTMLAttributes, forwardRef } from 'react'
+
+import stylesheet, { BASE_CLASSNAME, themeClasses } from '../css'
+import * as vars from '../vars'
+
+if (canUseDOM()) polyfillFocusWithin(document)
+
+export type StyleFn<P> = (
+  themeName: ValueOf<typeof themeNames>,
+  props: P
```
danethurber, line 21: so previously this was a little different then the component props because the values were required(the components are optional)

packages/halo/src/react/__specs__/storyshots.spec.tsx
jaketrent: I guess not! 🤷   I looked at the type of `snapshotWithOptions`, and it said it didn't even take that option, `createNodeMock`.  I removed it, tests ran and snapshots were the same.

packages/halo/src/react/index.tsx
jaketrent: oh, snap, lemme take another look...

PR #1340, opened 2020-10-14
Feat/text ts conversion

EdwardIrby:
convert to TS


packages/text/src/react/__specs__/__snapshots__/storyshots.spec.ts.snap
```diff
@@ -150,7 +150,6 @@ exports[`Storyshots Heading className override 1`] = `
       }
     }
     data-css-1uk9rqa=""
-    size="large"
```
danethurber, line 153: 👍

packages/text/src/css/index.ts
```diff
@@ -47,7 +47,7 @@ export default {
     textTransform: 'uppercase',
     fontSize: type.fontSizeXSmall,
     letterSpacing: type.letterSpacingXSmall,
-    lineHeight: type.lineHeighTight,
+    lineHeight: type.lineHeightTight,
```
danethurber, line 50: 👍

packages/text/src/react/__stories__/index.story.tsx
```diff
@@ -18,27 +18,27 @@ const PaddingDecorator = storyFn => (
...
 const heading = storiesOf('Heading', module).addDecorator(PaddingDecorator)
 Object.keys(Heading.sizes).forEach(size =>
-  heading.add(`size: ${size}`, _ => (
-    <Heading size={size}>
+  heading.add(`size: ${size}`, () => (
+    <Heading size={size as keyof typeof Heading.sizes}>
       <h1>{size}</h1>
     </Heading>
   ))
 )
```
danethurber, line 26: i think this work if you do `Object.values` instead....maybe...and then you don't have to cast `size`

packages/text/src/react/__stories__/index.story.tsx
```diff
@@ -18,27 +18,27 @@ const PaddingDecorator = storyFn => (
...
-heading.add('style override', _ => (
+heading.add('style override', () => (
   <Heading style={style}>
     <h2>pink</h2>
   </Heading>
 ))
-heading.add('className override', _ => (
-  <Heading className={className}>
+heading.add('className override', () => (
+  <Heading className={(className as unknown) as string}>
```
danethurber, line 33:
```suggestion
  <Heading className={...className}>
```

maybe? not sure

packages/text/src/react/code.tsx
```diff
@@ -0,0 +1,23 @@
+import { compose, css } from 'glamor'
+import React, {HTMLAttributes} from 'react'
+import { useTheme, names } from '@pluralsight/ps-design-system-theme'
+
+import stylesheet from '../css'
+
+const style = ({ themeName } : { themeName: keyof typeof names; }) =>
```
danethurber, line 7:
```suggestion
const style = ({ themeName } : { themeName: ValueOf<typeof names> }) =>
```

packages/text/src/react/code.tsx
```diff
@@ -0,0 +1,23 @@
...
+  compose(
+    css(stylesheet[`.psds-text__code`]),
+    css(stylesheet[`.psds-text__code.psds-theme--${themeName}`])
+  )
+
+const Code: React.FC<HTMLAttributes<HTMLElement>> = props => {
```
danethurber, line 13: 👍 

i guess `code` isn't cool enough to get it's own typings

packages/text/src/react/heading.tsx
```diff
@@ -0,0 +1,47 @@
+import { compose, css } from 'glamor'
+import React, { HTMLAttributes } from 'react'
+import { useTheme, names } from '@pluralsight/ps-design-system-theme'
+
+import stylesheet from '../css'
+import * as vars from '../vars'
+
+const style = ({ themeName, size }: { themeName?: keyof typeof names; size?: keyof typeof vars.headingSizes }) =>
```
danethurber, line 8: these should use ValueOf i think

packages/text/src/react/heading.tsx
```diff
@@ -0,0 +1,47 @@
...
+      stylesheet[`.psds-text__heading--size-${size}.psds-theme--${themeName}`]
+    )
+  )
+
+interface HeadingStatics {
+  sizes: typeof vars.headingSizes
+}
+
+interface HeadingProps extends HTMLAttributes<HTMLHeadingElement> {
+  size?:  keyof typeof vars.headingSizes
```
danethurber, line 23: ValueOf here as well

packages/text/src/react/heading.tsx
```diff
@@ -0,0 +1,47 @@
...
+}
+
+const Heading: React.FC<HeadingProps> & HeadingStatics = ({children, size, ...props}) => {
+  const themeName = useTheme()
+
+  return React.cloneElement(React.Children.only(children as React.ReactElement), {
```
danethurber, line 29:
```suggestion
  if(!React.isValidElement(children)) return
  
  return React.cloneElement(React.Children.only(children), {
```

i think that should work....

packages/text/src/react/list/index.tsx
```diff
@@ -0,0 +1,43 @@
...
+import React, {HTMLAttributes} from 'react'
+
+import { useTheme, names } from '@pluralsight/ps-design-system-theme'
+
+import ListItem from './list-item'
+
+import stylesheet from '../../css'
+import * as vars from '../../vars'
+
+const styles = ({ themeName, type}: { themeName: keyof typeof names; type: keyof typeof vars.listTypes}) =>
```
danethurber, line 11: same ValueOf

packages/text/src/react/list/index.tsx
```diff
@@ -0,0 +1,43 @@
...
+      css(stylesheet[`.psds-text__list--type-${type}`])
+    )
+
+interface ListStatics {
+  types: typeof vars.listTypes
+  Item: typeof ListItem
+}
+    
+interface ListProps extends HTMLAttributes<HTMLUListElement | HTMLOListElement> {
+  type?:  keyof typeof vars.listTypes
```
danethurber, line 24: same ValueOf

packages/text/tsconfig.json
```diff
@@ -3,6 +3,7 @@
   "compilerOptions": {
     "allowJs": true,
     "rootDir": null,
-    "jsx": "react"
+    "jsx": "react",
+    "strict": true
```
danethurber, line 7: 💥

danethurber (changes requested):

packages/text/src/react/list/index.tsx
EdwardIrby: Just realized this is exactly the same when I went back to look at the type. It also does not shorten it but I'll use it.

packages/text/src/react/__stories__/index.story.tsx
EdwardIrby: Nope looks like this breaks it

danethurber (approved):

PR #1343, opened 2020-10-15
Breadcrumb TS convert

jaketrent:
A couple interesting challenges here, surprisingly. The one that had me going for a while was related to how we split props. Some go onto a button, others onto a div.  Getting the HTMLAttributes of props and ref types right was "fun". See the "Omit" line for the light bulb.


packages/breadcrumb/src/react/index.tsx
```diff
@@ -0,0 +1,58 @@
...
+    BreadcrumbStatics
+  > {}
+
+const Breadcrumb = React.forwardRef<
+  HTMLAnchorElement | HTMLButtonElement,
+  BreadcrumbProps
+>((props, ref) => {
+  const { disabled, href, loading, onClick, ...rest } = props
+  return (
+    <div {...rest} {...styles.breadcrumb()}>
```
EdwardIrby, line 41: I know rest was probably here to begin with but I can't help but feel it should be on button. It's the interaction trigger not the div.  We should drop the padding on the wrapping div and add margins to button instead of padding on the wrapping div. We could even move the line-height to the button.

packages/breadcrumb/src/react/index.tsx
```diff
@@ -0,0 +1,58 @@
...
+import stylesheet from '../css'
+
+const styles = {
+  breadcrumb: () => css(stylesheet['.psds-breadcrumb'])
+}
+
+interface BreadcrumbProps
+  extends Omit<
+    React.HTMLAttributes<HTMLDivElement>,
+    'disabled' | 'href' | 'onClick'
```
EdwardIrby, line 16: Also if we dropped the wrapping div we wouldn't have to omit these

EdwardIrby (approved):
Please consider suggested changes

packages/breadcrumb/src/react/index.tsx
jaketrent: Hey, I think the point on the surrounding div is a fair point.  I felt that myself as I was realizing what we had done there.  But for now, I'm just wanting to type what is there with minimal collateral damage.

PR #1346, opened 2020-10-19
Card TS Convert

jaketrent:
Blast, forgot strict: true again.  Will work on that...


packages/card/src/react/__specs__/storyshots.spec.tsx
```diff
@@ -8,5 +8,6 @@ const createNodeMock = () => document.createElement('div')
 initStoryshots({
   configPath: path.resolve(__dirname, '../../../.storybook'),
   framework: 'react',
+  // @ts-ignore
```
jaketrent, line 11: This is the thing I feel the worst about.

The answer in other packages has been to remove this line. Here, it broke Shiitake in the storyshots.

Putting it back, the typings for the package don't even indicate that this is available. 

Perhaps there are new typings?  I haven't looked yet.  Perhaps we could patch/extend the typings locally. Haven't tried that yet either.

packages/card/src/react/index.tsx
```diff
@@ -0,0 +1,360 @@
+import polyfillFocusWithin from 'focus-within'
```
jaketrent, line 1: Lots o changes to this file around the sub components.  In short, there were a lot. Now there are a lot fewer.

packages/card/src/js/index.ts
```diff
@@ -0,0 +1,7 @@
+export const toPercentageString = (num: number): string => {
+  try {
+    return Math.min(parseFloat(num.toFixed()), 100).toString() + '%'
+  } catch (_) {
```
danethurber, line 4:
```suggestion
  } catch (_err) {
```

🤷  idk - seems like you'd need to do this to make ts happy

packages/card/src/react/__specs__/storyshots.spec.tsx
danethurber: let's make an issue and follow up in a focused pr

packages/card/src/react/__stories__/index.story.tsx
```diff
@@ -18,25 +19,24 @@ const longStringsMetaData = [
...
 
 const sizes = {
   [Card.sizes.small]: { width: '140px' },
   [Card.sizes.medium]: { width: '320px' },
   [Card.sizes.large]: { width: '540px' }
 }
-const ConstrainSizeDecorator = size => storyFn => (
-  <div style={sizes[size]}>{storyFn()}</div>
-)
+const ConstrainSizeDecorator = (size: keyof typeof sizes) => (
```
danethurber, line 33:
```suggestion
const ConstrainSizeDecorator = (size: ValueOf<typeof sizes>) => (
```

i've been checking the value instead of the key. that has made the most sense to me

packages/card/src/react/__stories__/index.story.tsx
```diff
@@ -286,8 +286,8 @@ storiesOf('fullOverlay', module)
 
 const sizeStory = storiesOf('size', module)
 Object.values(Card.sizes).forEach(size =>
-  sizeStory.add(size, _ =>
-    ConstrainSizeDecorator(size)(_ => (
+  sizeStory.add(size, () =>
+    ConstrainSizeDecorator(size)(() => (
```
danethurber, line 290: fyi there are newer ways to wrap decorators around stories. not really a big deal because all our stories use the old syntax anyway...i see some more refactoring in our future

packages/card/src/react/index.tsx
```diff
@@ -0,0 +1,359 @@
...
+  useTheme
+} from '@pluralsight/ps-design-system-theme'
+import {
+  canUseDOM,
+  omit,
+  RefForwardingComponent,
+  ValueOf
+} from '@pluralsight/ps-design-system-util'
+import { compose, css, StyleAttribute } from 'glamor'
+import * as React from 'react'
```
danethurber, line 15:
```suggestion
import React from 'react'
```

to match the more common pattern we use

packages/card/src/react/index.tsx
```diff
@@ -0,0 +1,359 @@
...
+    )
+  }
+}
+
+interface ActionBarActionProps extends React.HTMLAttributes<HTMLButtonElement> {
+  disabled?: boolean
+  icon?: React.ReactElement // TODO: retype as Icon when it's TS
+  title: string
+}
+interface ActionBarActionStatics {}
```
danethurber, line 139: if there are no statics you can type this without the extra overhead of `RefForwardingComponent`. not requesting a change - just letting you know. happy to show the difference

packages/card/src/react/index.tsx
```diff
@@ -0,0 +1,359 @@
...
+    return compose(
+      css(stylesheet[`.${label}`]),
+      css(stylesheet[`.${label}--size-${size}`])
+    )
+  }
+}
+
+interface ActionBarActionProps extends React.HTMLAttributes<HTMLButtonElement> {
+  disabled?: boolean
+  icon?: React.ReactElement // TODO: retype as Icon when it's TS
```
danethurber, line 136: Icon is now merged. type away

packages/card/src/react/index.tsx
```diff
@@ -0,0 +1,359 @@
...
+}
+Image.displayName = 'Card.Image'
+
+const ImageLink: React.FC<React.HTMLAttributes<HTMLSpanElement>> = props => (
+  <span {...styles.imageLink()} {...props} />
+)
+ImageLink.displayName = 'Card.ImageLink'
+
+interface TagProps extends React.HTMLAttributes<HTMLDivElement> {
+  icon?: React.ReactElement // TODO: Icon when Icon is TS
```
danethurber, line 190: Icon is now merged. type away

packages/card/src/react/index.tsx
```diff
@@ -0,0 +1,359 @@
...
+    'fullOverlay',
+    'fullOverlayVisible',
+    'image',
+    'metadata1',
+    'metadata2',
+    'progress',
+    'size',
+    'tag',
+    'title'
+  ])
```
danethurber, line 308: 😭  i need to figure out a better way to omit without all this overhead

packages/card/src/react/index.tsx
```diff
@@ -0,0 +1,359 @@
...
+  const size = props.size || sizes.medium
+  return (
+    <div {...styles.card()} {...htmlProps}>
+      <div {...styles.overlays({ size })}>
+        {props.image ? props.image : null}
```
danethurber, line 313: is the null in the ternary because of strict mode...or could you do

`{props.image && props.image}`

or even just

`{props.image}`

packages/card/src/react/index.tsx
```diff
@@ -0,0 +1,359 @@
...
+          <div {...styles.fullOverlay(props)}>{props.fullOverlay}</div>
+        ) : null}
+
+        {Array.isArray(props.actionBar) && props.actionBar.length > 0 ? (
+          <div {...styles.actionBar(props)}>{props.actionBar}</div>
+        ) : null}
+
+        {props.bonusBar ? (
+          <div {...styles.bonusBar()}>{props.bonusBar}</div>
+        ) : null}
```
danethurber, line 325: same question about null in the ternary

packages/card/src/react/index.tsx
```diff
@@ -0,0 +1,359 @@
...
+        {props.tag}
+
+        {props.progress ? (
+          <div {...styles.progress()}>
+            <div
+              {...styles.progressBar(props)}
+              aria-label={`${toPercentageString(props.progress)} complete`}
+            />
+          </div>
+        ) : null}
```
danethurber, line 336: same question about null in the ternary

danethurber (changes requested):

packages/card/src/react/index.tsx
jaketrent: Not sure why we wouldn't keep the new pattern. We're going to go in that direction eventually. I'm not stressed about going there piecemeal instead of en masse at a future point.  You ok with piecemeal, per-package?

packages/card/src/react/index.tsx
jaketrent: Great, yeah, I'd love to see an example.

packages/card/src/react/index.tsx
jaketrent: yeah, def seems overkill

packages/card/src/react/index.tsx
jaketrent: for tsc, yes. I can't remember if it was strict or unstrict that exposed this.

packages/card/src/react/index.tsx
jaketrent: ditto, yes, tsc.

packages/card/src/react/index.tsx
jaketrent: ditto, yes, tsc.

packages/card/src/react/index.tsx
danethurber: are you saying `* as React` is the new pattern. i don't really know what is the correct/current approach

with React 17 we don't need to import React at all for jsx...so we'd just pull in named exports? maybe? not sure

packages/card/src/react/__stories__/index.story.tsx
jaketrent: Yes, I think that is the common use case.  In this case, you can see on line 35 that `size` is actually a key in this case.

packages/card/src/react/__specs__/storyshots.spec.tsx
jaketrent: carded https://github.com/pluralsight/design-system/issues/1355

packages/card/src/react/index.tsx
jaketrent: Yes, I understand `* as React` to be the eventually-only supported import pattern.  Currently, you can go either way. 

See:
https://twitter.com/sebmarkbage/status/1250285076014747648

React 17 removes the need to use the react import as the jsx pragma.  But, we'll still need it for things like hooks or typings.

packages/card/src/react/index.tsx
jaketrent: done.

jaketrent:
Thanks for the review.  I have responded to or made all the changes that I think were requested.  Please have another look.

danethurber (approved):

PR #1351, opened 2020-10-21
Collapsible: typescript

danethurber:
diff is messier than most....sorry

i updated the directory structure to match our other components and the hook that was in `rc` is now used by the main component and exported


packages/verticaltabs/src/react/list.tsx
```diff
@@ -24,8 +24,9 @@ interface CollapsibleListProps extends ListProps {
   collapsed?: boolean
 }
 export const CollapsibleList: React.FC<CollapsibleListProps> = props => {
+  const { collapsed, ...rest } = props
```
EdwardIrby, line 27: nit picky but why not just destructure above?

EdwardIrby (approved):

EdwardIrby (changes requested):
👇🏿

packages/collapsible/tsconfig.build.json
```diff
@@ -0,0 +1,16 @@
...
+    "node_modules",
+    "dist",
+    "**/*.spec.*",
+    "**/*.story.*"
+  ],
+  "compilerOptions": {
+    "declaration": true,
+    "outDir": "dist",
+    "rootDir": "src"
+  }
```
EdwardIrby, line 15: Add strict

packages/verticaltabs/src/react/list.tsx
danethurber: personal preference. i don't like it when the line with the component definition breaks onto multiple lines - i think it makes it harder to read. also my editor highlighting can't always parse it. 🤷

danethurber:
the ts build file extends from the other ts file with has strict so it's only needed in the one place

PR #1352, opened 2020-10-21
Feat/dialog ts conversion

EdwardIrby:
focusmanager & dialog ts conversion


packages/dialog/src/react/__specs__/__snapshots__/storyshots.spec.ts.snap
```diff
@@ -44,12 +44,15 @@ exports[`Storyshots modal default 1`] = `
   <div
     aria-label="Storybook Modal"
     data-css-hdres9=""
+    disableCloseOnOverlayClick={false}
```
jaketrent, line 47: This seems like a bad addition.

packages/dialog/src/react/__specs__/__snapshots__/storyshots.spec.ts.snap
```diff
@@ -1798,12 +1824,23 @@ exports[`Storyshots stylesFor dialog_content 1`] = `
...
     id="psds-dialog__overlay"
     onClick={[Function]}
+    onClose={[Function]}
     role="region"
   >
     <div
+      UNSAFE_stylesFor={
```
jaketrent, line 1834: Shouldn't be in DOM

packages/dialog/src/react/__specs__/__snapshots__/storyshots.spec.ts.snap
```diff
@@ -1905,6 +1942,7 @@ exports[`Storyshots stylesFor dialog_content 1`] = `
 exports[`Storyshots tailPosition bottomCenter 1`] = `
 <div
   data-css-1pr6kyo=""
+  tailPosition="bottomCenter"
```
jaketrent, line 1945: Shouldn't be in DOM

packages/dialog/src/react/index.tsx
```diff
@@ -0,0 +1,169 @@
...
+  disableCloseButton?: boolean
+  disableCloseOnEscape?: boolean
+  disableCloseOnOverlayClick?: boolean
+  disableFocusOnMount?: boolean
+  modal?: boolean
+  onClose?: (evt: React.KeyboardEvent | React.MouseEvent) => void
+  tailPosition?: ValueOf<typeof vars.tailPositions>
+  returnFocus?: boolean
+  // eslint-disable-next-line camelcase
+  UNSAFE_stylesFor?: Record<string, unknown>
```
jaketrent, line 87: I think this is ok for now.  

I'm imagining how difficult this may be for devs to use.  Each one of these styles objects is totally custom (though we have a small convention).  So, it could be nice to have some typings at at least those first level object keys so that developers can have some help constructing the theme objects.

packages/focusmanager/src/react/use-focus-manager.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-unnecessary-type-assertion */
```
jaketrent, line 1: This is a weird line to have to disable eslint for this rule.  Where *is* the type assertion here?

packages/focusmanager/src/react/use-focus-manager.ts
```diff
@@ -13,35 +14,43 @@ const FOCUSABLE_SELECTORS = [
...
+  }: {
+    autofocus: boolean
+    returnFocus: boolean
+    trapped: boolean
+  }
+) {
+  const prevActiveElRef = React.useRef<HTMLElement>(null)
   React.useEffect(() => {
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
```
jaketrent, line 32: Why the ts-ignore here?  I'd like to avoid in all except the most extreme needs.

packages/focusmanager/src/react/use-focus-manager.ts
```diff
@@ -13,35 +14,43 @@ const FOCUSABLE_SELECTORS = [
...
+    trapped
+  }: {
+    autofocus: boolean
+    returnFocus: boolean
+    trapped: boolean
+  }
+) {
+  const prevActiveElRef = React.useRef<HTMLElement>(null)
   React.useEffect(() => {
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
```
jaketrent, line 31: just merged an eslintrc update that allows ts-comments, but it requires a commented reason.  Eg,

```
@ts-ignore: this is the reason why
```

packages/focusmanager/src/react/use-focus-manager.ts
```diff
@@ -13,35 +14,43 @@ const FOCUSABLE_SELECTORS = [
...
+    // @ts-ignore
     prevActiveElRef.current = document.activeElement
   }, [])
 
-  const [node, setNode] = React.useState(null)
+  const [node, setNode] = React.useState(
+    (null as unknown) as HTMLDivElement | null
+  )
```
jaketrent, line 38: Instead of casting, would it work to type useState?

```suggestion
  const [node, setNode] = React.useState<HTMLDivElement>(
    null
  )
```

packages/focusmanager/src/react/use-focus-manager.ts
```diff
@@ -13,35 +14,43 @@ const FOCUSABLE_SELECTORS = [
...
   React.useEffect(() => {
     setNode(ref.current)
   }, [ref])
 
-  const [focusableNodes, setFocusableEls] = React.useState([])
+  const [focusableNodes, setFocusableEls] = React.useState([] as HTMLElement[])
```
jaketrent, line 43: ditto?
```suggestion
  const [focusableNodes, setFocusableEls] = React.useState<HTMLElement[]>([])
```

packages/util/src/styles.ts
```diff
@@ -8,7 +8,7 @@ export function stylesFor(selectorKey: string, props: Props) {
   )
 }
 
-interface Props {
+export interface Props {
```
jaketrent, line 11: I'd like this to be a more specific name, since `util` pkg is a junk drawer and we are exporting tons of named stuff in the same namespace.

```suggestion
export interface StylesForProps {
```

jaketrent (changes requested):
storybook looks good.  tsc local looks good.

packages/dialog/src/react/index.tsx
EdwardIrby: I don't think a typing would matter here...it's an eslint error not a TS one. Typescript doesn't care. Does typescript override eslint if we make a type?

packages/focusmanager/src/react/use-focus-manager.ts
EdwardIrby: We're overriding a read only property and React types don't allow this but our component's original code does this for functionality.

packages/focusmanager/src/react/use-focus-manager.ts
EdwardIrby: needed for refs that need to be typed like focus

packages/focusmanager/src/react/use-focus-manager.ts
EdwardIrby: This throws a type error
```
Argument of type 'null' is not assignable to parameter of type 'HTMLDivElement | (() => HTMLDivElement)'.
```

packages/dialog/src/react/index.tsx
jaketrent: I agree with what you said: A typing, as I suggested, wouldn't affect eslint.  I wasn't thinking about eslint.  

Also to clarify: this isn't something I want necessarily included in the "request for change" on the PR.  Was me more pondering the future.

I was just thinking about how a developer might benefit from a type more specific than `Record<string, unknown>`.

packages/focusmanager/src/react/use-focus-manager.ts
jaketrent: oh, I see.  It's a disable for the whole file. I was reading it as a next-line like most of the others. 👍

packages/focusmanager/src/react/use-focus-manager.ts
jaketrent: Interesting that the react typings don't allow it but it's one of the "advanced use cases" for refs. Too bad.

packages/focusmanager/src/react/use-focus-manager.ts
jaketrent: `useState<HTMLDivElement | null>`?

I'm just attempting to use the typings on the generic function first as opposed to the casting.

EdwardIrby: Ahh I'll try that

jaketrent (approved):
Thanks for the updates. Tsc and storybook look good.  I'm going to merge since you're out for the week.

PR #1353, opened 2020-10-22
Row/Drawer ts conversion

danethurber:


packages/drawer/src/react/__stories__/index.story.tsx
```diff
@@ -0,0 +1,217 @@
...
+            <DrawerContent />
+          </Drawer>
+        </div>
+      )
+    }
+
+    return <PreventToggleDrawerStory />
+  })
+
+// TODO: move row stories to Row package to fix circular deps
```
danethurber, line 126: oops...i still need to do this

danethurber: so i removed since there are already related stories in `row`

packages/row/src/react/__specs__/storyshots.spec.tsx
```diff
@@ -8,5 +8,8 @@ const createNodeMock = () => document.createElement('div')
 initStoryshots({
   configPath: path.resolve(__dirname, '../../../.storybook'),
   framework: 'react',
+
+  // eslint-disable-next-line
+  // @ts-ignore
```
jaketrent, line 13:
```suggestion
  // @ts-ignore: storyshots tests require it, but it's not in storyshots typings
```

jaketrent (approved):

PR #1354, opened 2020-10-22
Carousel TS Convert

jaketrent:
yeeehaa!


packages/carousel/src/js/__specs__/utils.spec.ts
```diff
@@ -56,43 +56,6 @@ describe('utils', () => {
     })
   })
 
-  describe('#combineFns', () => {
```
jaketrent, line 59: using util pkg instead

packages/carousel/src/react/__specs__/index.spec.tsx
```diff
@@ -30,16 +30,22 @@ describe('Carousel', () => {
...
     it('should hide the prev control', () => {
-      const control = container
-        .querySelector('[aria-label="get previous carousel page"]')
-        .closest('[data-testid="carousel control"]')
+      const page = container.querySelector(
+        '[aria-label="get previous carousel page"]'
+      )
+      expect(page).toBeDefined()
+      // eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion
+      const control = page!.closest('[data-testid="carousel control"]')
```
jaketrent, line 38: I like the assertion here.  I don't like that I have to disable eslint.  The alternative is to put an `if` block around this.

packages/carousel/src/react/context.ts
```diff
@@ -0,0 +1,26 @@
...
+  offset: number
+  prev: () => void
+  id: string
+  pageCount: number
+  perPage: number
+  transitioning: boolean
+  setTransitioning: React.Dispatch<React.SetStateAction<boolean>>
+}
+const context = React.createContext<CarouselContextValue>({
+  activePage: 0,
```
jaketrent, line 15: Some of these defaults are new.  Simplified the typing process. Double-check my logic that having these are ok as defaults.

packages/carousel/src/react/__specs__/index.spec.tsx
danethurber: maybe disable this rule for .spec files 🤷

packages/carousel/src/react/control.tsx
```diff
@@ -34,33 +34,49 @@ const styles = {
...
 
-export function Control(props) {
+interface ControlProps
+  extends Omit<React.HTMLAttributes<HTMLButtonElement>, 'onClick'> {
+  direction: ValueOf<typeof vars.controlDirections>
+  onClick?: (evt?: React.MouseEvent) => void
+}
+interface ControlStatics {
+  directions: typeof vars.controlDirections
+}
```
danethurber, line 44: did you need to replace the `onClick` fn. seems like the types are the same

packages/carousel/src/react/index.tsx
```diff
@@ -0,0 +1,341 @@
...
+                key={pageIndex}
+                isActivePage={isActivePage}
+                paged={pager.paged}
+              >
+                {items.map(
+                  (item: React.ReactElement | null, itemIndex: number) => {
+                    const wrapped = !!item && isOfComponentType(item, Item)
+                    if (!wrapped) item = <Item key={itemIndex}>{item}</Item>
+
+                    // eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion
```
danethurber, line 114: one way to do this is to `{React.isActiveElement(item) && ...}`

packages/carousel/src/react/use-swipe.ts
```diff
@@ -23,10 +37,10 @@ export default function useSwipe(ref, opts = {}) {
     timeThreshold
   } = { ...defaultOpts, ...opts }
 
-  const timeout = useRef()
+  const timeout = useRef<NodeJS.Timeout>()
 
-  const [swipping, setSwipping] = useState(false)
```
danethurber, line 28: 🤦

danethurber (approved):
awesome work!

packages/carousel/src/react/control.tsx
jaketrent: yeah, great point.  Not sure why I did that. Probably just a too-literal conversion from proptypes.

packages/carousel/src/react/index.tsx
jaketrent: seems like I tried that option.  Not sure what the outcome was or why I chose this one.  There's a weekend in between in my mind.  

Is there an approach you think we should prefer?

packages/carousel/src/react/control.tsx
jaketrent: Well, doesn't make sense in isolation why I defined line 40.  But when I remove, I get down to combineFns, and I can't harangue Params to be typed correctly.

PR #1356, opened 2020-10-23
Feat/tooltip error ts conversion

EdwardIrby:
tooltip conversion && errors conversion


packages/errors/src/react/error-page.tsx
```diff
@@ -37,28 +40,28 @@ const styles = {
 
     return media(
       '(min-width: 769px)',
-      stylesheet['@media (min-width: 769px)'][`.${label}`]
+      stylesheet['@media (min-width: 769px)']['.psds-error-page__text']
```
jaketrent, line 43: If not using label, which I agree seems silly, should probably remove the label var.

packages/errors/src/react/error-page.tsx
```diff
@@ -67,39 +70,51 @@ const styles = {
...
-  <form action={props.action} method="get" {...styles.search(props)}>
-    <div {...styles.searchIcon(props)}>{icons.search()}</div>
+interface SearchFormProps extends HTMLAttributes<HTMLFormElement> {
+  action?: string
+  themeName: ValueOf<typeof themeNames>
+}
+const SearchForm: React.FC<SearchFormProps> = ({
+  action,
+  themeName,
+  ...props
```
jaketrent, line 80: I'm all for our other pattern of `rest`-named vars in order to help disambiguate from original/full `props`

packages/errors/src/react/forbidden-error-page.tsx
```diff
@@ -0,0 +1,19 @@
...
+import ErrorPage from './error-page'
+
+interface ForbiddenErrorPageProps extends HTMLAttributes<HTMLAnchorElement> {
+  href?: string
+}
+const ForbiddenErrorPage: React.FC<ForbiddenErrorPageProps> = ({
+  href = 'https://help.pluralsight.com/help/contact-us'
+}) => (
+  <ErrorPage
+    iconId="eyeball"
```
jaketrent, line 12: Not sure it applies to this PR, but perhaps these should be enumed/exposed in vars so that they're code completable?

packages/errors/src/react/heading.tsx
```diff
@@ -0,0 +1,58 @@
+import { compose, css } from 'glamor'
+import React, { HTMLAttributes } from 'react'
+
+import {
+  css as textStylesheet,
```
jaketrent, line 5: rename plain `stylesheet`?  Look for name collisions first.

packages/errors/src/react/icon-loader.ts
```diff
@@ -0,0 +1,28 @@
+/* eslint-disable @typescript-eslint/ban-ts-comment */
+// @ts-ignore
```
jaketrent, line 2: Would really prefer to define a typings file for these packages, even if empty.

packages/errors/src/react/icon-loader.ts
```diff
@@ -0,0 +1,28 @@
...
+import camelize from 'camelize'
+// @ts-ignore
+import importAll from 'import-all.macro'
+
+const imported = importAll.sync('./icons/*.dist.js')
+
+const convertToIconId = (filePath: string) => {
+  const trimmed = filePath.replace('./icons/', '').replace('-icon.dist.js', '')
```
jaketrent, line 10: Do we generate .dist.js anymore?  Or are they .dist.ts files?

packages/tooltip/src/css/index.ts
```diff
@@ -191,9 +191,9 @@ function getDiamondHeight(height, width) {
   )
 }
 
-function getDiamondWidth(height, width) {
-  height = parseInt(height, 10)
-  width = parseInt(width, 10)
+function getDiamondWidth(height: string | number, width: string | number) {
```
jaketrent, line 194: Verify that number | string are required and can't be consolidated.

packages/tooltip/src/react/index.tsx
```diff
@@ -0,0 +1,126 @@
...
+} from '@pluralsight/ps-design-system-util'
+import stylesheet from '../css'
+import * as vars from '../vars'
+
+const fade = keyframes(stylesheet[`@keyframes psds-tooltip__keyframes__fade`])
+const styles = {
+  tail: ({
+    appearance,
+    tailPosition
+  }: {
```
jaketrent, line 16: Attempt to avoid re-defining TooltipProps on all these.

jaketrent (changes requested):
I have a few things I suggest changing.  They're mostly nits.  Since Edward is out, I will address them.

packages/errors/src/react/forbidden-error-page.tsx
jaketrent: Already typed.  We're good here.

iconId?: keyof typeof icons

packages/errors/src/react/icon-loader.ts
jaketrent: We use dist.js here.  I'm not sure it matters.  They are svgr-generated files.

jaketrent:
@danethurber I have made the changes I wanted on this.  Would you mind adding your review when you're ready, please?

packages/errors/src/react/icon-loader.ts
danethurber: svgr has a typescript output option...but since we don't export the icons i'm not sure it matters

danethurber (approved):

PR #1359, opened 2020-10-26
TS convert datawell

jaketrent:
- refactor(datawell): mv .js to .ts files
- refactor(datawell): tsc round one
- refactor(datawell): rm prop-types
- refactor(datawell): remove .js exts on imports
- refactor(datawell): remove filterReactProps
- test(datawell): fix up tests
- refactor(datawell): tsc strict


packages/datawell/src/react/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+      stylesheet[`.psds-datawell__sublabel.psds-theme--${themeName}`]
+    )
+}
+
+interface DataWellProps extends React.HTMLAttributes<HTMLDivElement> {
+  children: React.ReactNode
+  label: string
+  subLabel?: React.ReactNode
+}
+
```
danethurber, line 43:
```suggestion
interface DataWellProps extends React.HTMLAttributes<HTMLDivElement> {
  label: string
  subLabel?: React.ReactNode
}

```

danethurber (approved):

packages/datawell/src/react/index.tsx
jaketrent: child node was marked as required on the old proptypes.  Given that, would you still make this change (because React.HTMLAttributes<HTMLDivElement> is going to make it optional)?

danethurber: hmmm.....i probably would make the change - but i get your point. i like not needing to add props that are already covered. i leave it up to you to make an exec decision based on component usage

PR #1360, opened 2020-10-27
Patch/checkbox a11y

danethurber:


packages/checkbox/src/css/index.ts
```diff
@@ -114,11 +114,6 @@ export default {
     fill: colorsTextIcon.highOnDark
   },
 
-  // __input
-  '.psds-checkbox__input': {
-    display: 'none'
```
danethurber, line 119: this completely removed the input from the accessibility tree

packages/checkbox/src/react/index.tsx
```diff
@@ -108,6 +97,7 @@ const Checkbox = React.forwardRef<HTMLInputElement, CheckboxProps>(
 
     return (
       <label
+        htmlFor={id}
```
danethurber, line 100: use `htmlFor` and `id` to link label/checkbox to input

packages/checkbox/src/react/__specs__/index.spec.tsx
```diff
@@ -23,4 +23,15 @@ describe('Checkbox', () => {
...
+
+  it('is accessible by role', () => {
+    const { getByRole } = render(
+      <>
+        <Checkbox label="First" name="first" value="value" />
+        <Checkbox label="Second" name="second" value="value" />
+      </>
+    )
+
+    expect(getByRole('checkbox', { name: /first/i })).toBeInTheDocument()
```
danethurber, line 35: @willsoto i needed to use regex to query the accessibility tree. couldn't do it just by string ie `{ name: 'first' }`

packages/util/src/index.ts
```diff
@@ -18,14 +18,17 @@ export { whitelistProps } from './props'
...
 export { useMatchMedia } from './use-match-media'
 export { usePortal } from './use-portal'
 export { usePrevious } from './use-previous'
 export { useResizeObserver } from './use-resize-observer'
 export { useToggle } from './use-toggle'
-export {
-  useCloseOnDocumentEvents,
-  onGlobalEventsClose
-} from './use-close-on-document-events'
+export { useUniqueId } from './use-unique-id'
```
jaketrent, line 34: thanks for putting this in util

packages/util/src/use-unique-id.ts
```diff
@@ -0,0 +1,13 @@
+import { useLayoutEffect, useState } from 'react'
+
+import { uniqueId } from './unique-id'
+
+export function useUniqueId(prefix = '') {
+  const [id, setId] = useState('')
+
+  useLayoutEffect(() => {
```
jaketrent, line 8: Why does this have to be useLayoutEffect instead of useEffect?

packages/checkbox/src/css/index.ts
jaketrent: why did we have it, then?

jaketrent (approved):

packages/checkbox/src/css/index.ts
danethurber: we want to hide the input - but i used the screen ready only styles instead

packages/util/src/use-unique-id.ts
danethurber: i actually don't remember...i just copied this over

PR #1363, opened 2020-10-28
Appframe convert to typescript

danethurber:


packages/appframe/src/react/index.tsx
```diff
@@ -202,28 +240,34 @@ function SkipBanner(props) {
...
+interface SideNavStatics {
+  variants: typeof vars.sidenavVariants
+}
+const SideNav: React.FC<SideNavProps> & SideNavStatics = props => {
   const { sidenavVariants: variants } = vars
 
-  const { children, ...rest } = props
+  const { children: _children, ...rest } = props
+  const children = props.children as SideNavProps['children']
+
```
danethurber, line 270: @jaketrent so the issue we couldn't fix yesterday is because React.FC appends `{children: ReactNode}` onto the end of whatever props you declare...so i had to get hacky

packages/appframe/src/react/index.tsx
```diff
@@ -65,19 +70,62 @@ const styles = {
...
+})
+
+type RenderProp<P extends Record<string, unknown>> = (
+  props: P
+) => React.ReactNode
+
+interface AppFrameProps extends HTMLAttributes<HTMLDivElement> {
+  onRequestSideNavClose?: () => void
+  onRequestSideNavOpen?: () => void
+  sidenav?: ReactChild | RenderProp<{ visible: boolean }>
```
danethurber, line 93: i should probably use `ReactNode` instead of `ReactChild` throughout this file. will update

packages/appframe/src/react/index.tsx
```diff
@@ -172,21 +219,15 @@ const AppFrame = React.forwardRef((props, forwardedRef) => {
...
-  children: PropTypes.node,
-  onRequestSideNavClose: PropTypes.func,
-  onRequestSideNavOpen: PropTypes.func,
-  sidenav: PropTypes.oneOfType([PropTypes.func, PropTypes.node]),
-  sidenavOpen: PropTypes.bool,
-  topnav: PropTypes.oneOfType([PropTypes.func, PropTypes.node])
-}
 
-function SkipBanner(props) {
+// TODO: why is there a an href here?
```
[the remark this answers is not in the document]
jaketrent: I tested the div#href link in storybook, and it doesn't seem to work.  May be a pre-existing condition.

jaketrent (approved):
Rebuilt, tsc is good, storybook looks good.  Tests still fail, but looks like an env thing, so I'm waiting to see what CI says.

PR #1364, opened 2020-10-28
update peer deps

danethurber:
because lerna doesn't do this for us 🤦


packages/appframe/package.json
```diff
@@ -32,8 +32,8 @@
     "prop-types": "^15.7.2"
   },
   "peerDependencies": {
-    "@pluralsight/ps-design-system-normalize": "^3.1.4",
-    "@pluralsight/ps-design-system-theme": "^3.0.2",
+    "@pluralsight/ps-design-system-normalize": "^4.0.0",
+    "@pluralsight/ps-design-system-theme": "^6.0.0",
```
jaketrent, line 36: All this could create a lot of new noise for people as they build and see new requirements.  Are we sure that all these packages *need* these late versions?

Seems like we may not.

But then it might quiet noise that people like Will were seeing by now using the latest theme package version, for instance.

Seems like we might get the best of both worlds with something more flexible with a range, like theme@4-6??  Not sure it matters.

jaketrent (approved):

packages/appframe/package.json
danethurber: we want to make sure all versions of theme are on the same range - so that need to match. and i think we'll need that range to be the latest from packages that use theme@6 as a direct dependency

PR #1367, opened 2020-10-28
TS Convert Datepicker

jaketrent:
agh


packages/datepicker/package.json
```diff
@@ -9,21 +9,21 @@
...
     "build:css": "build-css --useGlamor -i dist/cjs/css/index.js",
-    "build:js": "cross-env ESM=true babel --root-mode upward src --out-dir dist/esm",
-    "build:watch": "yarn build:js -- --watch",
+    "build:esm": "tsc --project tsconfig.build.json --module es2015 --target es5 --outDir dist/esm",
+    "build:watch": "yarn build:esm -- --watch",
     "storybook": "start-storybook -p 6006",
-    "test": "jest",
+    "test": "tsc --noEmit --project ./tsconfig.json && jest",
     "test:watch": "yarn test -- --watchAll"
   },
```
danethurber, line 20: need references to the type declarations..and might as well to the styles as well

packages/datepicker/src/js/index.ts
```diff
@@ -0,0 +1,136 @@
...
+
+  const maxDays = parseInt(getDaysInMonth(date), 10)
+  const day = parseInt(date.dd, 10)
+  return isNaN(day)
+    ? ''
+    : day > maxDays
+    ? maxDays.toString()
+    : day < 1
+    ? '1'
+    : day.toString()
```
danethurber, line 27: not new...but wanted to complain that this is hard to read 😛

danethurber (changes requested):

packages/datepicker/src/js/index.ts
jaketrent: I will hear your complaint and add another TS-approved if/else statement.

jaketrent:
Should be good to go again, Dane.  Please have a look when you're ready.

danethurber (approved):

PR #1369, opened 2020-10-29
Refactor/checkbox a11y

danethurber:
our checkboxes were not accessible from the screen reader or from testing-library it seems. i took a new approach that is a bit simpler and that i could get working with both VoiceOver and in tests

notable improvements
- input label is now read by voiceover
- no need for a `role` 
- no need for a unique id
- added jest-axe...it will catch ~30% of a11y errors...so not the best safety net
- it's noise...but i did a quick refactor of the typing and style....it's never complete

cc @willsoto


packages/checkbox/src/react/index.tsx
```diff
@@ -70,108 +69,93 @@ const Checkbox = React.forwardRef<HTMLInputElement, CheckboxProps>(
...
-      }
+      toggle(evt)
+    }
 
-      if (square.current) square.current.focus()
+    const handleKeyDown: React.KeyboardEventHandler<HTMLLabelElement> = evt => {
+      if (disabled) return
+
+      evt.persist()
+      if (evt.key === 'Enter' || evt.key === ' ') toggle(evt)
```
danethurber, line 100: native checkboxes only response to space....we also support `Enter`...i just left it

jaketrent (changes requested):
Build and test and code look great.

Tried using voiceover, it seems great.

Clicking and using enter work great.

Space does not work great.  When I check, it immediately unchecks.  Does it do that for you?

packages/checkbox/src/react/index.tsx
```diff
@@ -65,113 +64,101 @@ const Checkbox = React.forwardRef<HTMLInputElement, CheckboxProps>(
...
+    }
 
-      const { onCheck } = props
-      if (typeof onCheck === 'function') {
-        onCheck(evt, !props.checked, props.value, props.name)
-      }
+    const handleKeyDown: React.KeyboardEventHandler<HTMLLabelElement> = evt => {
+      if (disabled) return
 
-      if (square.current) square.current.focus()
```
danethurber, line 95: i removed this feature to be more inline with the native browser interaction. 🤷

jaketrent: What *was* this feature?  Look like it's for keeping focus on the checkbox after clicking.  That seems like a browser-native thing to do.  I'll checkout and try out...

jaketrent (approved):
Seems to work just fine.  

So, I guess I'm still interested in the reason for the square.current.focus() removal

packages/checkbox/src/react/index.tsx
danethurber: it shows the *halo* focus indicator when a checkbox is clicked with a mouse. this was different that the native implementation so i changed it. gifs attached

native behaviour
Nov-04-2020 15-10-37

new behavior that matches native

Nov-04-2020 15-13-14

PR #1372, opened 2020-10-30
polymorphic button types

danethurber:
currently the button doesn't support undeclared properties that should be allow. example - the Button should allow a `type` props. 

Screen Shot 2020-10-29 at 3 45 34 PM

i've attempted to update the prop typings to allow button or anchor props....but i can't figure it out. anybody else feel like taking a shot?


jaketrent:
I'd love to give it a shot.

packages/button/src/react/index.tsx
```diff
@@ -126,89 +129,115 @@ interface ButtonStatics {
...
+  iconAlign?: ValueOf<typeof vars.iconAligns>
   loading?: boolean
-  size?: string
+  size?: ValueOf<typeof vars.sizes>
+}
+
+type AnchorElProps = ButtonProps<HTMLAnchorElement> & { href: string }
+type ButtonElProps = ButtonProps<HTMLButtonElement> & {
+  name?: string
+  type?: string
```
danethurber, line 145: needed to redeclare these really defeats the purpose....

jaketrent:
@danethurber what do you think of the latest commit.  I used Will's approach.  Seems to tsc ok, test ok, and get code completion ok.  I'm probably missing something. 🤷

danethurber (commented):
let's try it. i honestly don't know what is correct at this point

jaketrent:
+1

jaketrent (approved):

PR #1382, opened 2020-11-02
ts conversion: NavBar and NavBrand

danethurber:


packages/navbar/package.json
```diff
@@ -10,22 +10,23 @@
...
   "gitHead": "7418883d96a76c59a99f86c2b16516ec752bf913",
+  "style": "dist/index.css",
+  "types": "dist/esm/index.d.ts",
   "dependencies": {
     "@pluralsight/ps-design-system-core": "^6.5.1",
-    "@pluralsight/ps-design-system-filter-react-props": "^1.2.1",
     "@pluralsight/ps-design-system-icon": "^19.0.5",
     "@pluralsight/ps-design-system-navitem": "^1.4.7",
-    "prop-types": "^15.7.2"
+    "@pluralsight/ps-design-system-util": "*"
```
jaketrent, line 29: Are we going to '*' on internal deps?  Doesn't seem like we have entirely.  Why here?

packages/navbar/src/react/index.tsx
```diff
@@ -17,31 +15,36 @@ const styles = {
   utility: () => css(stylesheet['.psds-navbar__utility'])
 }
 
-const NavBar = React.forwardRef((props, forwardedRef) => {
-  const ref = React.useRef()
-  React.useImperativeHandle(forwardedRef, () => ref.current)
+interface NavBarProps extends React.HTMLAttributes<HTMLDivElement> {
+  brand?: React.ReactNode
+  items?: React.ReactNode
+  onMobileMenuClick?: React.MouseEventHandler
```
jaketrent, line 21: ooh, nice type.  I haven't used that. Maybe I could have.  Nice shorthand

packages/navbrand/src/react/__specs__/index.spec.tsx
```diff
@@ -1,20 +1,51 @@
...
+  it('should render a hyperlink when an `href` is present', () => {
+    const ref = React.createRef<HTMLAnchorElement>()
+    const { getByTestId } = render(
+      <NavBrand data-testid="undertest" href="/" ref={ref}>
+        test
+      </NavBrand>
+    )
+    const el = getByTestId('undertest')
+
+    expect(el.tagName.toLowerCase()).toEqual('a')
```
jaketrent, line 29: Why not rely on the stories that seem to test for the renderings of `<a>`s and `<button>`s?

packages/navbrand/src/react/index.tsx
```diff
@@ -1,25 +1,29 @@
...
+import { css, media } from 'glamor'
+import React, { MouseEventHandler, ReactNode, RefObject } from 'react'
 
-import filterReactProps from '@pluralsight/ps-design-system-filter-react-props'
+import stylesheet from '../css'
 
-import stylesheet from '../css/index.js'
+// TODO: move to utils
+type AnyTag = keyof JSX.IntrinsicElements
+type PropsOf<Tag extends AnyTag> = JSX.IntrinsicElements[Tag]
```
jaketrent: This seems to be the "union of 173 keys" that is mentioned here:

https://blog.andrewbran.ch/polymorphic-react-components/#distributivity-schmistributivity-im-gonna-make-it-work

tsc passes in this code.  But I have some trepidation about this warning:

> TypeScript 3.5 does handle this without crashing by deferring a lot of the work that was happening to simplify conditional types, but do you really want to write components that are just waiting for the right moment to explode?

I'm not sure how this might be manifest, but it seems like a problem:

> Long story short, you’ve just invented a button that cannot be reasoned about within Node’s default heap size. 

Long build times?  Non-performant tooling?  Crashing builds in which we are a dependency?  I'm not sure.  But I feel forewarned.

packages/navbrand/src/react/index.tsx
```diff
@@ -34,52 +38,79 @@ const styles = {
...
+          {...wrapperProps}
+          {...(rest as PropsOf<'div'>)}
+        />
+      )
+
+    return (
+      <Halo inline gapSize={Halo.gapSizes.small}>
+        <Wrapper {...styles.navBrand(props)}>
+          <Logo>{logo}</Logo>
+          <Wordmark>{wordmark}</Wordmark>
```
jaketrent, line 106: What was your reasoning for changing these internal apis?

jaketrent (commented):
Great work for a complex problem.  

I have a major concern (see inline comment) in the implementation based on it seeming to teeter on a pitfall that the most conclusive/deep blog on the Internet about React polymorphic components seems to have warned about.  I'm not sure what the answer is.  Maybe it's just go forward, but I'm not sure how to back out of it once we encounter the forewarned problems.  

As an alternative, would the Button polymorphic solution be applied with success here?  If not, what limitation did you run into?

packages/navbar/package.json
danethurber: i just `*` is and lerna will update it with the version range during publish. the works for deps and devDeps. lerna does not do it for peerDeps

packages/navbrand/src/react/__specs__/index.spec.tsx
danethurber: mostly just for TDD purposes. I like that they are focused and the test name is explicit about what is tested

packages/navbrand/src/react/index.tsx
danethurber: two fold.
  - limit what is passed to them as props. just children instead of all nav brand props
  - use children because it's standard and doesn't need additional typings

packages/navbrand/src/react/index.tsx
danethurber: yes totally that is my main concern as well. where i think this is different and ok....is that we only allow a button, a, or div as the element to the NavBrand component. for something that would allow any element, things like the ScreenReaderOnly component or where we allow the `tagName`/`as` props i think we need to explore different solutions like the renderContainer prop

danethurber: now i'm reading the blog article again(i think i'm at 10 times). i still think there could be edge cases that we probably haven't accounted for - mostly in typescript perf. without changing the components api i do think this pr demonstrates the best approach i have. the alternative is to explore new ways to conditionally render tagNames for these polymorphic components

I vote we try this out and see what happens. i'm also going to start another thread in ts-guild

packages/navbar/package.json
jaketrent: So, if lerna takes care of setting dependencies on publish, should we just * most all the internal psds deps?  Sounds like it to me.  Maybe there's a corner case where we have more specific version requirements, and we'd set those specifically.

packages/navbrand/src/react/index.tsx
jaketrent: Re the `keyof JSX.IntrinsicElements` issue: If we wanted to make the union smaller, could we do `keyof Pick<keyof JSX.IntrinsicElements, 'a'|'button'|'div'>`?

Re the best approach we know of: I still don't know what's better about this approach than the Button component approach we also have.  Will you expound on or demonstrate that, please?  If not better in a substantial way, maybe we don't have to take the downsides here too?

packages/navbar/package.json
danethurber: lerna replaces the `*` during publish. so i just set it initially and expect lerna to do the right thing

packages/navbrand/src/react/index.tsx
danethurber: gotcha. let me step back real quick and see if it works the same with the attrs types

danethurber: @jaketrent tried using types like in button. ie

```
interface AnchorProps
  extends BaseNavBrandProps,
    React.AnchorHTMLAttributes<HTMLAnchorElement> {
  href: string
  onClick?: MouseEventHandler<HTMLAnchorElement>
}
interface ButtonProps
  extends BaseNavBrandProps,
    React.ButtonHTMLAttributes<HTMLButtonElement> {
  href?: undefined
  onClick: MouseEventHandler<HTMLButtonElement>
}
```

but the compiler is unable to parse the overloads or infer the attrs to use. I think `JSX.IntrinsicElements` is the way to go

jaketrent: okey dokey.  Did you ever try decreasing the union size with `Pick<keyof JSX.IntrinsicElements, 'a'|'button'|'div'>`?

jaketrent (approved):

packages/navbrand/src/react/index.tsx
danethurber: yeah i did. it doesn't seem to have any effect on the compile time but makes the generic util less useful. at this point i don't think the PropsOf utils is of any concern since it's just a lookup map and doesn't return a large union type - it should be no different than what the react typings are already doing....i think

I will remove the AnyTag type just so it's not accidentally abused

jaketrent: If in order to be useful, PropsOf includes `keyof JSX.IntrinsicElement` and then creating a potential perf problem, maybe it shouldn't exist.  

One could write:

```
interface AnchorProps extends BaseNavBrandProps, JSX.IntrinsicElements['a'] {}
```

directly into the NavBrand component file.  This doesn't seem to be overly burdensome.

Am I seeing this right?

danethurber: yeah that makes sense to me. i'll update the republish

danethurber: well....i have no idea why but that doesn't work. i still do think it's a problem because we're only ever allowing up to 3 of the elements. we avoid the problem from the blog post where they are allowing all elements

> Ok, fine. Let’s say you work out a way around that assignability error, and you’re ready to replace 'a' | 'button' with all keyof JSX.IntrinsicElements.
> …and, congratulations, you’ve crashed TypeScript 3.4!

we haven't gone that far...yet 😉

jaketrent: Huh, weird.  Ok, well, it is what it it is for now, I suppose.  

Thanks for talking through it.

PR #1386, opened 2020-11-03
fix(docs): set one cookie for all paths and time

jaketrent:
Not 100% this will fix it, but it does seem to address the symptom of multiple session cookies.


danethurber (approved):
let's try it out

PR #1400, opened 2020-11-04
Detect and render theme on first render

jaketrent:
- Accomplished via inline script that sets body for flipping css vars
- Things that rely on react context for theming, are also getting set on first render, but with the imperfection of a blink of unstyled content in the case of light theme
- In the future, it's recommend to default to dark theme in the case of users have not preferred or selected a theme.  This because the DS components all default to dark theme when no theme is given, so the blink doesn't happen.
- Another downside: the blink is now happening between pages.  This may be a non-starter, I'm not sure. I'm inclined to take this as an improvement and deal with the per-page blink asap when we figure it out.


danethurber (approved):
i'm willing to merge it and see what happens

jaketrent:
In prod now.  Looks pretty good.  

@sweeting Seems like you may have said it's ok in standup, but will you take a look at the behavior in prod, then weigh in on whether it's ok to set dark theme as the default/fallback in the case of no preference or selection, please?

PR #1401, opened 2020-11-04
fix(docs): rm main body scrollbar

### LINKED ISSUE #1399: docs: rm odd scrollbar from voice and tone page
Screen Shot 2020-11-04 at 1 01 21 PM

jaketrent:
Resolves #1399

Not sure why the overflow: auto was there.  Since I couldn't think of a good reason doesn't mean that there isn't one.  Anyone have some more context on this?

### DIFF
--- a/docs/src/components/frame/frame.module.css
+++ b/docs/src/components/frame/frame.module.css
@@ -55,6 +55,7 @@
 }
 
 .main {
+  overflow-x: auto;
   flex: 0 1 auto;
   margin: 0 auto;
   max-width: 1000px;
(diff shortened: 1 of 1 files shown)


danethurber:
looking at the commits i think this is it

https://github.com/pluralsight/design-system/commit/b9dde7515e7262244d1a7f3c857329115db2baf1

codeblock with large hoz width were causing an overflow. maybe `overflow-x: none` would work better

jaketrent:
@danethurber good call, thanks for the pointer.

danethurber (approved):

danethurber:
this should resolve #1399

PR #1405, opened 2020-11-04
Default docs theme to dark

jaketrent:
Double checked with Sweeting.  He sees this is as a ux-neutral/lil-positive change, so we're good on that front.

### DIFF
--- a/docs/gatsby-ssr.js
+++ b/docs/gatsby-ssr.js
@@ -46,7 +46,7 @@ const prefersDark = window.matchMedia('(prefers-color-scheme: dark)').matches
 const initialThemeName =
   cookieThemeName || (prefersDark && '${Theme.names.dark}') || '${Theme.names.light}'
 
-document.body.setAttribute('data-theme', initialThemeName)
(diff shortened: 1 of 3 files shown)


docs/gatsby-ssr.js
```diff
@@ -41,12 +41,11 @@ function getCookie(name) {
...
 
-const prefersDark = window.matchMedia('(prefers-color-scheme: dark)').matches
+const prefersLight = window.matchMedia('(prefers-color-scheme: light)').matches
 
-const initialThemeName =
-  cookieThemeName || (prefersDark && '${Theme.names.dark}') || '${Theme.names.light}'
+const initialThemeName = cookieThemeName || (prefersLight && '${Theme.names.light}') || '${Theme.names.dark}'
 
-document.body.setAttribute('data-theme', initialThemeName)
+document.body.setAttribute('data-psds-theme', initialThemeName)
```
jaketrent, line 48: just a bit more specific about where this doc-level attr is coming from.

danethurber: good idea

danethurber (approved):

Issue #57, opened 2017-08-11
site: add font applied to page usage example


jaketrent:

jaketrent:
did this with a normalize usage example

PR #1407, opened 2020-11-05
TS Convert Note

jaketrent:
## Instructions

- fill in title: format of "<package name>: short description"
- label: attach bug or enhancement request label

### What You're Solving

- Any links to #57 related issues
- Explanation of the purpose of the code

### Design Decisions

- Why you made certain design decisions
- Anything to watch out for
- Any input you're seeking related to the design
- Tradeoffs you conciously made

### How to Verify

- Steps needed to independently verify additions


packages/note/package.json
```diff
@@ -10,25 +10,24 @@
...
-    "build:js": "cross-env ESM=true babel --root-mode upward src --out-dir dist/esm",
-    "build:watch": "yarn build:js -- --watch",
+    "build:esm": "tsc --project tsconfig.build.json --module es2015 --target es5 --outDir dist/esm",
+    "build:watch": "yarn build:esm -- --watch",
     "storybook": "start-storybook -p 6006",
-    "test": "jest",
+    "test": "tsc --noEmit --project ./tsconfig.json && jest",
     "test:watch": "yarn test -- --watchAll"
   },
   "gitHead": "7418883d96a76c59a99f86c2b16516ec752bf913",
```
danethurber, line 22: add reference to generated typings file

packages/note/package.json
```diff
@@ -10,25 +10,24 @@
...
     "@pluralsight/ps-design-system-avatar": "^6.1.7",
-    "@pluralsight/ps-design-system-core": "^6.5.1",
-    "@pluralsight/ps-design-system-filter-react-props": "^1.2.1",
-    "@pluralsight/ps-design-system-prop-types": "^2.1.1"
+    "@pluralsight/ps-design-system-core": "^6.5.1"
   },
   "peerDependencies": {
     "@pluralsight/ps-design-system-normalize": "^4.0.0",
     "@pluralsight/ps-design-system-theme": "^6.0.0",
+    "@pluralsight/ps-design-system-util": "*",
```
danethurber, line 30: should this be a dep and not a peerDep?

if peerDeps is the right approach....we should move all util to peerDeps and mark version range

packages/note/src/react/index.tsx
```diff
@@ -0,0 +1,235 @@
...
+            {renderActionBar}
+          </div>
+        )}
+
+        {message}
+
+        <div {...styles.footer()}>
+          {hasMetadata && (
+            <div {...styles.metadata(themeName)}>
+              {(metadata || []).map((datum: React.ReactNode, i) => {
```
danethurber, line 160: maybe move the default array to the props destructing...your call

packages/note/src/react/__specs__/__snapshots__/storyshots.spec.tsx.snap
```diff
@@ -552,7 +552,6 @@ exports[`Storyshots Note/actions no author - long metadata 1`] = `
         >
           <button
             data-css-td0jeq=""
-            icon={<BookmarkIcon />}
```
danethurber, line 555: 👍

danethurber (approved):
approach with one change - add typing references to the package.json file

packages/note/package.json
jaketrent: oops, yeah

packages/note/src/react/index.tsx
jaketrent: I don't want a default array, I think.  This is just because TS isn't smart enough to see the safety of `hasMetadata` above.

PR #1413, opened 2020-11-06
Feat/table ts conversion

EdwardIrby:
ts conversion

Also the inDrawer prop isn't actually used in the current code https://github.com/pluralsight/design-system/blob/cd0b624596912658db979abc93bf7cfdcc1ce066/packages/table/src/react/index.js#L216-L239


packages/table/src/react/__stories__/index.story.tsx
```diff
@@ -435,7 +435,7 @@ storiesOf('drawer', module)
     return (
       <Table>
         <Drawer base={baseRow} startOpen>
-          <Table inDrawer>
```
danethurber, line 438: we should update the docs if this isn't a thing anymore

https://github.com/pluralsight/design-system/blob/0b544456b4325e5d5c509afde0197fa4e8142508/docs/src/content/components/table.mdx

danethurber (approved):

packages/table/src/react/index.tsx
```diff
@@ -0,0 +1,265 @@
...
+  row: (themeName: ValueOf<typeof themeNames>, _tableHasDrawers: boolean) => {
+    const label = 'psds-table__row'
+
+    return compose(
+      css(stylesheet[`.${label}`]),
+      css(stylesheet[`.${label}.psds-theme--${themeName}`]),
+      _tableHasDrawers && css(stylesheet[`.${label}--drawers`])
+    )
+  },
+  table: (themeName: ValueOf<typeof themeNames>, inDrawer: boolean) => {
```
jaketrent, line 70: inDrawer looks like it's still a thing.

PR #1414, opened 2020-11-06
TS Position Convert

jaketrent:
- [x] tsc first pass is happy!
- [x] storybook weirdness - problem was spacing around story's <Box />
- [x] tsc strict


packages/position/src/react/index.tsx
```diff
@@ -0,0 +1,174 @@
...
+
+interface ShowProps {
+  ref?: React.RefObject<HTMLElement>
+  style?: React.CSSProperties
+}
+interface PositionProps {
+  children: React.FunctionComponentElement<{}>
+  inNode?: HTMLElement
+  position: positionFns.PositionFunction
+  show: React.FunctionComponentElement<ShowProps>
```
jaketrent, line 84: Made required.  I don't know why we wouldn't want this.

packages/position/src/react/index.tsx
```diff
@@ -0,0 +1,174 @@
...
+
+    const ref = React.useRef<HTMLElement>()
+    React.useImperativeHandle(forwardedRef, () => ref.current as HTMLElement)
+
+    const child = React.Children.only(
+      props.children as React.FunctionComponentElement<{}>
+    )
+    const innerRef = React.useRef<HTMLElement>()
+    const showRef = props.show.ref || innerRef
+    const showEl = cloneElementWithRef(
```
jaketrent, line 109: Used util#cloneElementWithRef instead of React.cloneElement here.  Was happier with typing.  And it seemed like we were being smarter in this util fn.

packages/util/src/clone-element-with-ref.ts
```diff
@@ -13,7 +13,7 @@ export const cloneElementWithRef = <
   T extends HTMLElement
 >(
   el: ReactHTMLElement<T>,
-  outerRef: MutableRefObject<T>,
+  outerRef: MutableRefObject<T | undefined>,
```
jaketrent, line 16: Most of the code we're righting has refs that match this type.  We rarely declare useRef with an initial value of type T, which is the only way we avoid undefined | null.

EdwardIrby (approved):

packages/position/src/react/__stories__/index.story.tsx
```diff
@@ -1,20 +1,45 @@
-import { storiesOf } from '@storybook/react'
-
-import { css } from 'glamor'
-import PropTypes from 'prop-types'
-import React from 'react'
 import { InfoIcon } from '@pluralsight/ps-design-system-icon'
 import * as core from '@pluralsight/ps-design-system-core'
 import Tooltip from '@pluralsight/ps-design-system-tooltip'
+import { usePortal, ValueOf } from '@pluralsight/ps-design-system-util'
```
danethurber, line 4:
```suggestion
import { usePortal } from '@pluralsight/ps-design-system-util'
```

packages/position/src/react/index.tsx
```diff
@@ -0,0 +1,170 @@
...
+          ? target.current
+          : ref.current
+      if (!showRef.current || !targetNode) return
+
+      const nextStyle = positionFn(targetNode).styleFor(showRef.current)
+      setStyle(nextStyle)
+      setShownOnce(true)
+    }, [positionFn, target, showRef.current])
+
+    useOnWindowResize(evt => updateStyle())
```
danethurber, line 131:
```suggestion
    useOnWindowResize(_evt => updateStyle())
```

packages/position/src/react/index.tsx
```diff
@@ -0,0 +1,170 @@
...
+  return <Position position={positionFns.leftOf} ref={ref} {...props} />
+})
+LeftOf.displayName = 'LeftOf'
+
+interface ShowProps {
+  ref?: React.RefObject<HTMLElement>
+  style?: React.CSSProperties
+}
+interface PositionProps {
+  children: React.FunctionComponentElement<{}>
```
danethurber, line 81: what is `FunctionComponentElement` and is that better to use here than `ReactNode` or `ReactElement`

danethurber (approved):
there a good amount of stuff going on here. looks good in storybook. i left a couple small questions/comments

packages/position/src/react/index.tsx
jaketrent: Franky, I'm not 100% sure and still learning.  First timed used.  Though I think I also use it in Tab.

From the typings, it's a ReactElement with a ref:

```
    interface FunctionComponentElement<P> extends ReactElement<P, FunctionComponent<P>> {
        ref?: 'ref' extends keyof P ? P extends { ref?: infer R } ? R : never : never;
    }
```

Which is what I wanted. It's used here because it allows me to see the ref on the child and the props on the child. 

This was a major challenge in this refactor.

PR #1416, opened 2020-11-09
final nav component ts conversions

danethurber:
TODO:
- [x] `NavCookieBanner` converted
- [x] `NavItem` converted
- [x] `NavUser` converted
- [ ] update docs

Note: Breaking Change in `NavItem`. trying to type the polymorphic(`a`|`button`|`div`) interface was unwieldy so I adopted a renderProp to allow changing the underlying dom. This change looks like the following:

Before:
```ts
<NavItem href="/" icon={<Icon />} onClick={...} ref={ref} >
  Label
</NavItem>
```

After:
```ts
<NavItem
  icon={<Icon />}
  renderContent={props => (
    <a href="/" onClick={...} ref={ref} {...props} />
  )}
>
  Label
</NavItem>
```

feedback and prop naming suggestions appreciated


packages/navitem/src/react/context.tsx
```diff
@@ -0,0 +1,28 @@
...
+}
+
+export const initialValue: ContextValue = {
+  alignment: alignments.horizontal,
+  menu: false,
+  selected: false,
+  renderContainer: injected => <button {...injected} />
+}
+
+const Context = createContext<ContextValue>(initialValue)
```
danethurber, line 26: adopted a context for the NavItem to simplify typings and avoid prop drilling

packages/navitem/src/react/common.tsx
```diff
@@ -0,0 +1,47 @@
...
+    )
+  },
+  button: (props: { selected: boolean }) =>
+    compose(
+      css(stylesheet['.psds-navitem__button']),
+      props.selected && css(stylesheet['.psds-navitem__button--selected'])
+    )
+}
+
+export const Bar: React.FC<PropsOf<'div'>> = props => {
```
danethurber, line 34: I believe that using the `PropsOf` util (which is doing `JSX.IntrinsicElements[TagName]`) is more complete than `HTMLAttributes<Element>`. here is an example of the difference -> https://codesandbox.io/s/wonderful-night-c38uw?file=/src/Button.test.tsx

you'll see that when using `HTMLAttributes` - typescript can't completely type all button attrs - `type` is not typed

Screen Shot 2020-11-09 at 12 32 32 PM

packages/navitem/src/react/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+import Context, { ContextValue, initialValue } from './context'
+import { HorzLayout } from './horz'
+import { VertLayout } from './vert'
+
+type AllowedSelectors = 'navitem__bar' | 'navitem__bar--selected'
+
+interface NavItemProps
+  extends PropsWithStylesFor<AllowedSelectors>,
+    Partial<ContextValue> {}
```
danethurber, line 14: there is a new util(`PropsWithStylesFor`) to simplify `stylesFor` and add type checking

packages/navuser/src/react/index.tsx
```diff
@@ -0,0 +1,113 @@
...
+  onClick?: MouseEventHandler<HTMLDivElement>
+}
+
+type NavUserElement = HTMLAnchorElement | HTMLButtonElement | HTMLDivElement
+type NavUserProps = AnchorProps | ButtonProps | DivProps
+type NavUserComponent = React.ForwardRefExoticComponent<NavUserProps> & {
+  (props: AnchorProps, ref?: RefFor<'a'>): JSX.Element
+  (props: ButtonProps, ref?: RefFor<'button'>): JSX.Element
+  (props: DivProps, ref?: RefFor<'div'>): JSX.Element
+}
```
danethurber, line 47: we could adopt the render prop polymorphic pattern here as well. I would change the component prop to be slightly more complicated but would made the component internals simpler

packages/navitem/src/react/common.tsx
EdwardIrby: Would the union type issue pop up here?

packages/navitem/src/react/horz.tsx
```diff
@@ -0,0 +1,74 @@
...
+        </span>
+      </Button>
+
+      <Bar />
+    </HorzContainer>
+  )
+}
+HorzLayout.displayName = 'NavItem.HorzLayout'
+
+const HorzContainer: React.FC<PropsOf<'span'>> = props => {
```
jaketrent, line 51: Maybe I'm scared of a boogie man who doesn't exist, but I really don't want to use PropsOf except where it's deemed necessary.  

At first glance, PropsOf usage here seems much wider than necessary, for a couple unrelated reasons:

1. React.HTMLAttributes has gotten us a long way -- almost through the TS conversion of the entire DS codebase.  It feels uneven to change the strategy on these last couple packages.

2. `HorzContainer` simply doesn't need this.  It doesn't take props, so it doesn't need to be typed to anticipate HTML attributes.

Props are able to be sent:

```
51:   const HorzContainer: React.FC<PropsOf<'span'>> = props => {
```

, but they are never sent:

```
36:   <HorzContainer>
```

I don't think we need this prop spreading or even the props param.

packages/navitem/src/react/index.tsx
jaketrent: This is a great add, thank you!

packages/navbar/src/react/__stories__/index.story.tsx
```diff
@@ -69,14 +69,21 @@ storiesOf('Navbar', module).add('desktop', () => {
...
-                    onClick={() => setBrowseMenuOpen(!isBrowseMenuOpen)}
-                  >
-                    Browse
-                  </NavItem>
+                  <div>
+                    <NavItem
+                      icon={<BrowseIcon />}
+                      selected
+                      menu
+                      renderContainer={containerProps => (
```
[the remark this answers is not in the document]
jaketrent: I found the Context that for NavItem below.  Looks like button is the default tagName used in the renderContainer initialValue. Great.

But then why is there what looks like an override here?  

Looking at common.ts, Button below, must be for the onClick.  Why would a `NavItem#onClick` not be passed to the default renderContainer function?

Without this, I don't think the default `renderContainer` is very useful at all.  Ie, What good is a button without a click handler?

packages/navuser/src/react/index.tsx
```diff
@@ -0,0 +1,113 @@
...
+    const ref = React.useRef<NavUserElement>(null)
+    React.useImperativeHandle(
+      forwardedRef,
+      () => (ref.current as unknown) as NavUserElement
+    )
+
+    const isAnchor = 'href' in props
+    const isButton = !isAnchor && 'onClick' in props
+
+    const Wrapper: React.FC = wrapperProps =>
```
jaketrent, line 62: nice!

jaketrent (commented):
Lots of great work here.

I have some typing strategy questions and comments that I think warrant some conversation before a merge.

packages/navcookiebanner/src/react/index.tsx
```diff
@@ -0,0 +1,106 @@
...
+  message: () => css(stylesheet['.psds-navcookiebanner__message']),
+  dismiss: () => css(stylesheet['.psds-navcookiebanner__dismiss'])
+}
+
+interface NavCookieBannerProps extends React.HTMLAttributes<HTMLDivElement> {
+  cookieOptions?: CookieSetOptions
+  message?: React.ReactNode
+  privacyUpdatedDate?: typeof Date
+}
+interface CookieSetOptions {
```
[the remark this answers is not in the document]
danethurber: They aren't exported so i copy/pasted them from `react-cookie`. if you have a way to get that out of `react-cookie` i would prefer that

packages/navitem/src/react/__specs__/index.spec.tsx
```diff
@@ -0,0 +1,53 @@
...
+        UNSAFE_stylesFor={{
+          navitem__bar: {},
+          'navitem__bar--selected': {}
+        }}
+      >
+        test
+      </NavItem>
+    )
+
+    expect(true).toBe(true)
```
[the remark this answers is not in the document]
danethurber: totally. must have added to ignore a linting error at the time. i'll move to a story/snapshot or something better

packages/navitem/src/react/common.tsx
danethurber: it's not necessary for the Bar component since it's a `div` - just the first part of the review where i hit the typing so i'm calling it out. i'm trying to show that the difference between `HTMLAttributes` and `JSX.IntrinsicElements` is a important one and that `JSX.IntrinsicElements` provides a higher level of introspection. Since it's the more complete typing i'd venture we'd want to use it everywhere.

`AllHTMLAttributes` would include the `type` prop but accept any `string`. `JSX.IntrinsicElements` properly types the `type` attribute to only include values accepted on a `button` which provides greater introspection. `AllHTMLAttributes` also includes all possible attributes for **any** element - which seem way to wide and error prone.

packages/navbar/src/react/__stories__/index.story.tsx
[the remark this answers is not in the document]
danethurber: - yes button is the default. the old implementation accepted an anchor or a button. drilling the props and types through the components to get to the Button component was just crazy
- the story was supposed to show how to change from using the default element to something else...so i should be a story showing an `a` being use. i'll update
- the story also shows adding things like ref and event handlers. we could add those to the `NavItem` props but then we hit the issues of drilling the types/props and typing polymorphism
- i doubt we need to support div at all....but...with the render prop the consumer can actually replace the element with whatever they want 🤷

packages/navitem/src/react/horz.tsx
danethurber: 1: i've addressed above around PropsOf/JSX.IntrinsicElements providing a more complete set of typings and more accurate introspection. at this point i'm reasonably sure(75% 😉 ) that JSX.IntrinsicElements doesn't provide any complier issues when used to type a single element or a small set of elements for polymorphism. It is also used in the react types everything we reach for `ComponentProps` or `ElementType` which we have used with not apparent downsides. I agree that it's a pivot from our previous strategy of using `HTMLAttributes` but at this point i think it's the write pivot to make - especially for the public apis of the components

2: i see where you're coming from since the props aren't used and the components are internal. I personally like spreading props for all valid attributes - and with typescript we benefit from the type checking to ensure the props we pass are valid. but it's a personal preference and not something i'm all that attached to. i'd be happy to add typings as we need them to internal components

as always i think the discussion is important and happy to keep churning on it

packages/navbar/src/react/__stories__/index.story.tsx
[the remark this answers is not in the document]
danethurber: this is a pretty new pattern so i'm happy to work through it before we merge. the `renderContainer` name isn't something i'm all that happy with. also the way refs and button/link props is no longer part of the component api is both good and bad for various reasons

packages/navitem/src/react/common.tsx
jaketrent: ^ Thanks for your additional observations on `AllHTMLAttributes`.  The latter point, esp, I had missed "AllHTMLAttributes also includes all possible attributes for any element".  I agree: too wide and unuseful.

packages/navitem/src/react/horz.tsx
jaketrent: 1: Thanks for your reasoning around using JSX.IntrinisicElements. I agree with this being a pivot.  Let me play with it a little today to further convince myself (hopefully approach a happy 75% with you) that this is a good thing.

2: Not a huge deal.  The most basic thing I'm saying is that if the props aren't used, don't type them and don't spread them.  This preference is in favor of simplicity and noise reduction.

packages/navitem/src/react/__specs__/index.spec.tsx
```diff
@@ -28,26 +26,11 @@ describe('NavItem', () => {
     const ref = React.createRef<HTMLButtonElement>()
 
     render(
-      <NavItem renderContainer={props => <button ref={ref} {...props} />}>
+      <NavItem renderContent={props => <button ref={ref} {...props} />}>
```
jaketrent, line 29: Almost seems like this test case can go away.  Now it's reading as "`<button />` can render with refs", which must, of course be true.

Well, I guess if we didn't use renderContent() or messed with it somehow, _maybe_ that's less-obviously true.

packages/navitem/src/react/context.tsx
```diff
@@ -12,15 +14,17 @@ export interface ContextValue {
   bar?: ReactNode
   icon?: ReactNode
   menu: boolean
-  renderContainer: (props: ContainerProps) => JSX.Element
+  renderContent: (props: ContentProps) => JSX.Element
```
jaketrent, line 17: Why the switch from renderContainer?

I know you said you "didn't like" the naming.  

I think I prefer "container" generally.  Because generally we're replacing some outermost tag, in this case the `button` with an `a`, etc.  This also seems a common term for this problem outside this codebase.

In scrollable, I went with `renderContent` because it was the _inner_ tag (the... content 🤷 , I dunno, it's a pretty lame generic term too) that I was changing the tagName of.

packages/navitem/src/react/index.tsx
```diff
@@ -24,12 +26,22 @@ const NavItem: React.FC<NavItemProps> & NavItemStatics = props => {
     children,
     icon,
     menu = initialValue.menu,
-    renderContainer = initialValue.renderContainer,
-    selected = initialValue.selected
+    renderContent = initialValue.renderContent,
```
jaketrent, line 29: There's a weirdness in setting prop defaults with context defaults and then setting the context value with the result.

But I think I know why you're doing it: probably so you have fewer null checks to do on menu, renderContent and selected w/in the codebase.

Good news is that menu and selected are booleans, so those can easily go w/o a default.

Alignment is similar because it's "either-or" vert/horz value and could have a default/fallback.

This might mean you don't have to ensure non-nulls after all.

packages/navitem/src/react/common.tsx
```diff
@@ -12,6 +12,8 @@ const styles = {
   bar: (props: {
     alignment: ValueOf<typeof alignments>
     selected: boolean
+    // eslint-disable-next-line camelcase
+    UNSAFE_stylesFor: unknown
   }) => {
     const isHorz = (props.alignment = alignments.horizontal)
     const isVert = (props.alignment = alignments.vertical)
```
jaketrent: These look fishy to me.  Seems like this should be a test (`==`) instead of assignment (`=`).

packages/navitem/src/react/common.tsx
```diff
@@ -31,17 +34,26 @@ const styles = {
     )
 }
 
-export const Bar: React.FC<PropsOf<'div'>> = props => {
-  const { alignment, selected } = useContext(Context)
-  return <div {...styles.bar({ alignment, selected })} {...props} />
+type DivElProps = JSX.IntrinsicElements['div']
```
jaketrent, line 37: I guess we're to HTMLPropsFor at the moment.

jaketrent (approved):

packages/navitem/src/react/common.tsx
danethurber: ha. good catch. 🤦

packages/navitem/src/react/common.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+const styles = {
+  bar: (props: {
+    alignment: ValueOf<typeof alignments>
+    selected: boolean
+    // eslint-disable-next-line camelcase
+    UNSAFE_stylesFor: unknown
+  }) => {
+    const isHorz = (props.alignment = alignments.horizontal)
+    const isVert = (props.alignment = alignments.vertical)
```
jaketrent, line 19: These look fishy to me. Seems like this should be a test (==) instead of assignment (=).

packages/navitem/src/react/__specs__/index.spec.tsx
danethurber: yeah makes sense. let me noodle on that a bit. i'll probably remove the test

packages/navitem/src/react/context.tsx
danethurber: yeah i just matched the name in scrollable

packages/navitem/src/react/common.tsx
danethurber: ha. good catch 🤦

PR #1417, opened 2020-11-09
TS Convert Scrollable

jaketrent:
- [ ] Update docs

- refactor(scrollable): mv .js to .ts
- build(checkbox): add missing types and script packagejson
- build(scrollable): update npm-scripts
- refactor(scrollable): tsc convert; use renderProp for content render
- test(scrollable): rm createNodeMock
- test(scrollable): adjust accidentally-altered test; rm trailing index.js
- test(scrollable): update test config for tsx
- refactor(util): allow useResizeObserve to take MutableRefObject
- refactor(scrollable): tsc strict


packages/scrollable/src/react/index.tsx
```diff
@@ -0,0 +1,204 @@
...
+function areEqualProps(
+  prevProps: Record<string, unknown>,
+  nextProps: Record<string, unknown>
+) {
+  const changed = shallowCompare(prevProps, nextProps)
+  return !changed
+}
+
+interface ScrollableProps extends React.HTMLAttributes<HTMLDivElement> {
+  renderContent?: (props: renderContentProps) => React.ReactNode
```
jaketrent, line 51: This is a breaking change.

For context: https://pluralsight.slack.com/archives/C5CMEKH5Z/p1604933253002100

packages/scrollable/src/react/index.tsx
```diff
@@ -0,0 +1,204 @@
...
+    )
+  }
+)
+Scrollable.displayName = 'Scrollable'
+
+interface renderContentProps {
+  children: React.ReactNode
+  cssSelectors: StyleAttribute
+  onScroll: React.UIEventHandler
+  setRef: React.RefCallback<HTMLElement>
```
jaketrent, line 150: Dunno if this is "the best" name for this.  Need to forward ref.  Only allowed through callback ref w/o using forwardRef, afaik.

packages/util/src/use-resize-observer.ts
```diff
@@ -1,12 +1,12 @@
...
-import { RefObject, useEffect, useState } from 'react'
+import { MutableRefObject, RefObject, useEffect, useState } from 'react'
 import ResizeObserver from 'resize-observer-polyfill'
 
 import { combineFns } from './combine-fns'
 import { debounce } from './debounce'
 
 export function useResizeObserver(
-  ref: RefObject<HTMLElement>,
+  ref: RefObject<HTMLElement> | MutableRefObject<HTMLElement | undefined>,
```
jaketrent, line 9: add MutableRefObject since that's the type returned from forwardRef.  Needed in this case.  And probably a common case.

packages/scrollable/src/react/index.tsx
danethurber: that's good point. i don't know the best way to handle the ref. maybe a second arg to mimic the `forwardRef` arity

packages/scrollable/src/react/index.tsx
```diff
@@ -0,0 +1,204 @@
...
+}
+const defaultRenderContent = (props: renderContentProps) => {
+  return (
+    <div {...props.cssSelectors} onScroll={props.onScroll} ref={props.setRef}>
+      {props.children}
+    </div>
+  )
+}
```
danethurber, line 158: maybe there is a way to do the css different? i like the idea of just spreading all the props like

`(contentProps, contentRef) => <div ref={ref} {...contentProps} />`

i think that change would help usage and upgrades

jaketrent: Yeah, I'll play with this.  Thanks!

packages/scrollable/src/react/index.tsx
```diff
@@ -0,0 +1,206 @@
...
+          {...styleOverride}
+        />
+      </Outer>
+    )
+  }
+)
+Scrollable.displayName = 'Scrollable'
+
+export interface renderContentProps {
+  /* [cssSelector: string]: string */
```
danethurber, line 153: remove?

packages/scrollable/src/react/index.tsx
```diff
@@ -0,0 +1,206 @@
...
+  nextProps: Record<string, unknown>
+) {
+  const changed = shallowCompare(prevProps, nextProps)
+  return !changed
+}
+
+interface ScrollableProps extends React.HTMLAttributes<HTMLDivElement> {
+  renderContent?: (
+    props: renderContentProps,
+    ref: React.RefCallback<HTMLElement>
```
danethurber, line 53: should this be `React.Ref`

>     type Ref<T> = RefCallback<T> | RefObject<T> | null;

I honestly still don't know what to use for Refs

packages/scrollable/src/react/index.tsx
```diff
@@ -0,0 +1,206 @@
...
+      <Outer data-scrollable {...rest}>
+        <Inner key="scrollable-wrapper">
+          {renderContent(
+            {
+              children: props.children,
+              onScroll: updateDimensions,
+              ...styles.content()
+            },
+            (node: HTMLElement) => (ref.current = node)
+          )}
```
danethurber, line 136: 👍

danethurber (approved):

packages/scrollable/src/react/index.tsx
jaketrent: Yeah, probably.

I really wanted this line to be there, but I couldn't get it to jive well with tsc.  

I wanted it here because it indicates that the cssSelector is unknown in the value, but it will be present and  usable by anyone overriding `renderContent`.

But that overriding user won't know the value either, so I'm not sure just the instructions to "always spread the incoming props onto the content markup" aren't sufficient anyway.

packages/scrollable/src/react/index.tsx
jaketrent: Yes, that still works as Ref, which is the supertype, so I'll use it.

Still think a user will only be able to pass a RefCallback in practice (no RefObject via forwardRef, only LegacyRef left available), but that could just be a misunderstanding.

PR #1419, opened 2020-11-09
TS Convert Tab

jaketrent:
- [x] Refactor into overloaded types


packages/tab/src/react/list-item.tsx
```diff
@@ -0,0 +1,63 @@
...
+export interface BaseListItemProps {
+  id: string | number
+  active?: boolean
+  onClick?: (i: number, event: React.MouseEvent) => void
+}
+export type ListItemButtonTagProps = BaseListItemProps &
+  Omit<React.ButtonHTMLAttributes<HTMLButtonElement>, 'id' | 'onClick'>
+export type ListItemAnchorTagProps = BaseListItemProps &
+  Omit<React.AnchorHTMLAttributes<HTMLAnchorElement>, 'id' | 'onClick'>
+type ListItemProps = ListItemButtonTagProps | ListItemAnchorTagProps
```
danethurber, line 36: by using a union and not an overload we lose some type safety. i setup an example here

https://codesandbox.io/s/angry-robinson-woy3s?file=/src/Item.test.tsx

in the example you'll see that we are expecting an anchor tag but the props are not properly typ checked as such

danethurber (approved):
left a comment about how we can provide better prop typings

packages/tab/src/react/list-item.tsx
jaketrent: Thank you for looking critically at this.  

Please be more specific with this so I can see where we're not getting proper type checking.  Which part is not getting typechecked in the example but is wrong?

I do see this, fwiw:
Screen Shot 2020-11-10 at 6 00 05 AM

danethurber: yes of course. the typing error in the example is actually correct and i'm showing how an `a` tag should be typed. 

the `ListItem` in the example should be an `a` since it has an `href` but it's wrongly accepting an eventHandler with a `HTMLButtonElement` type

jaketrent: ah, I finally see it.  The typings do not prevent mixing/matching anchor/button stuff.  And this shows that in practice.  Thanks for the codesandbox and the explanation.  I will play with type overloads.

packages/tab/src/react/list-item.tsx
```diff
@@ -0,0 +1,72 @@
...
+    Omit<JSX.IntrinsicElements['button'], 'id' | 'onClick'> {
+  href?: undefined
+  onClick?: (i: number, event: React.MouseEvent<HTMLButtonElement>) => void
+}
+type ListItemElement = HTMLButtonElement | HTMLAnchorElement
+type ListItemProps = ListItemAnchorProps | ListItemButtonProps
+type ListItemComponent = ForwardRefExoticComponent<ListItemProps> & {
+  (props: ListItemAnchorProps, ref?: RefObject<HTMLAnchorElement>): JSX.Element
+  (props: ListItemButtonProps, ref?: RefObject<HTMLButtonElement>): JSX.Element
+}
```
jaketrent, line 48: Here are the new typings.

1. I did not use the util helpers PropOf and RefFor.

Instead of:

```
PropsOf<'a'>
```

I did:

```
JSX.IntrinsicElements['a']
```

Instead of:

```
RefFor<'a'>
```

I did:

```
RefObject<HTMLAnchorButton>
```

I prefer this approach, because I feel that while the utils may be useful markers for an approach difference, the abstraction isn't super high value.  Removing the abstraction removes a mental translation -- this, every time it's used and every time you inspect the abstraction, which has non-trivial internal typings, to try to figure out what it really means. Once removed, the code is of similar length. And at least in the case of the RefObject, the syntax is more common and familiar.

Thoughts on removing the PropsOf and RefFor abstractions?

danethurber: I can get behind removing the abstractions. I like that they look the same(take a tagName) and they are short. but that's not enough reason to keep them :)

danethurber (approved):

PR #1420, opened 2020-11-10
Feat/tag starrating ts conversion

EdwardIrby:
ts conversion


packages/starrating/src/react/__stories__/index.story.tsx
```diff
@@ -16,14 +16,20 @@ const rounding = storiesOf(
...
 })
 
 ROUNDING_VALUES.forEach(value => {
-  rounding.add(`${value}`, _ => (
-    <StarRating onChange={action('value changed')} value={value} />
+  rounding.add(`${value}`, () => (
+    <StarRating
+      onChange={(
+        val: React.Key,
+        evt: React.ChangeEvent<Element> | React.MouseEvent<Element, MouseEvent>
```
jaketrent, line 30:
```suggestion
        _val: React.Key,
        _evt: React.ChangeEvent<Element> | React.MouseEvent<Element, MouseEvent>
```

It's too bad that tsc apparently requires typing this.  Storybook's `action` was always so nice and terse.

packages/starrating/src/react/star.tsx
```diff
@@ -65,38 +73,65 @@ function HalfStarIcon(props) {
   )
 }
 
-function Star(props) {
+interface StarProp
+  extends Omit<
+    HTMLAttributes<HTMLSpanElement & HTMLButtonElement>,
```
jaketrent, line 78: I love this simplicity.  But I wonder if it's sufficient.  Because as written this means that every Star instance can have the attributes of spans AND buttons, allowing something like:

```
<Star interactive={false} disabled name="star" value="3" />
```

*Practically*, however, this is probably ok, because *we* control all usage of Star, because it's not a public interface.  For that reason, I could go either way -- keep as-is or make polymorphic.

packages/tag/src/react/index.tsx
```diff
@@ -0,0 +1,143 @@
...
+    )
+  }
+}
+
+export interface TagStatics {
+  sizes: typeof vars.sizes
+}
+
+export interface TagProps
+  extends HTMLAttributes<HTMLAnchorElement & HTMLDivElement> {
```
jaketrent, line 61: Unlike Star, I think this is insufficient and should be updated to be polymorphic.

This is because one of the main goals of types is to give proper introspection into our components so devs will use them correctly.  

`<Tag download />` shouldn't be legal.

`<Tag href="file.zip" download />` should be.

But both are with this typing.

packages/tag/src/react/index.tsx
```diff
@@ -0,0 +1,143 @@
...
+
+  const style = {
+    cursor: icon.props.onClick ? 'pointer' : 'default'
+  }
+  const onClick = (evt: React.MouseEvent) => {
+    evt.stopPropagation()
+    if (icon.props.onClick) icon.props.onClick(evt)
+  }
+
+  return React.cloneElement(icon, {
```
jaketrent, line 134: way to get these cloneElement types to work cleanly w/o casting!

packages/tag/tsconfig.build.json
```diff
@@ -0,0 +1,17 @@
...
+    "node_modules",
+    "dist",
+    "**/*.spec.*",
+    "**/*.story.*"
+  ],
+  "compilerOptions": {
+    "declaration": true,
+    "outDir": "dist",
+    "rootDir": "src",
+    "strict": true
```
jaketrent, line 15: I have confirmed, btw, that this duplicate is not needed.

jaketrent (changes requested):
Great work!

I would look at Tag especially and try to make the HTMLAttributes polymorphic.

packages/starrating/src/react/star.tsx
EdwardIrby: will leave as is

packages/tag/src/react/index.tsx
EdwardIrby: sorta fixed I really think we need to figure this one out

packages/tag/src/react/index.tsx
```diff
@@ -56,24 +56,27 @@ const styles = {
 export interface TagStatics {
   sizes: typeof vars.sizes
 }
-
-export interface TagProps
-  extends HTMLAttributes<HTMLAnchorElement & HTMLDivElement> {
+export interface BaseProps {
```
jaketrent, line 59: I'd suggest `BaseTagProps` so that it's more specific and matching most components' approach.

jaketrent (approved):

PR #1421, opened 2020-11-10
Refactor/ts typeahead

danethurber:


packages/typeahead/src/js/index.ts
```diff
@@ -0,0 +1,23 @@
+export type FilterFn = () => boolean
```
jaketrent, line 1: Looks unused.

packages/typeahead/src/react/use-on-document-click.ts
```diff
@@ -1,10 +1,12 @@
...
-import React from 'react'
+import { useEffect } from 'react'
 
 const isBrowser = canUseDOM()
 
-export default function useOnDocumentClick(handler) {
-  React.useEffect(() => {
+type Callback = (evt: Event | MouseEvent | UIEvent) => void
+
+export default function useOnDocumentClick(handler: Callback) {
```
jaketrent, line 8: util's useCloseOnDocumentEvents also exists.  Maybe replaceable 🤷 ?

jaketrent (approved):
Not this PR, but I did find that I couldn't press keyboard Enter to select a result from the menu.  I'll card it.

PR #1423, opened 2020-11-11
Rm unused/deprecated prop-types utils

jaketrent:
These were deprecated when we started moving to TS. Now we are not
referencing them. Removing.

BREAKING CHANGE: Remove props and prop-type utils; unused since TS
conversion


danethurber (approved):
🎆

PR #1426, opened 2020-11-11
convert ts button types to overloads

jaketrent:
Better HTML attr checking. Lame "El" aliases.


packages/button/src/react/index.tsx
```diff
@@ -128,29 +128,33 @@ interface ButtonStatics {
   iconAligns: typeof vars.iconAligns
   sizes: typeof vars.sizes
 }
-
 interface BaseButtonProps {
   appearance?: ValueOf<typeof vars.appearances>
+  // TODO: rm? dup of html attr
   disabled?: boolean
```
danethurber, line 134: i think only button has `disabled` - so you'd want to leave it so the anchors can also be disabled

danethurber (approved):

packages/button/src/react/index.tsx
```diff
@@ -128,29 +128,33 @@ interface ButtonStatics {
...
+type ButtonProps = ButtonAnchorProps | ButtonButtonProps
+
+type ButtonComponent = RefForwardingComponent<
+  ButtonProps,
+  HTMLAnchorElement | HTMLButtonElement,
+  ButtonStatics
+> & {
+  (props: ButtonAnchorProps, ref?: RefObject<HTMLAnchorElement>): JSX.Element
+  (props: ButtonButtonProps, ref?: RefObject<HTMLButtonElement>): JSX.Element
+}
```
danethurber, line 157: the linter doesn't seem to like this style of overriding

packages/button/src/react/index.tsx
jaketrent: good call

packages/button/src/react/index.tsx
jaketrent: Huh, seems to have made it through the last ci lint ok.  What is flagging for you?

danethurber: ci was erroring out. if that passed i'm not concerned

jaketrent: Maybe I checked too fast.  We'll see this time around.  CI looked like a test failure to me, which was in Banner, tsc-related.  Maybe there was something else in there too...

PR #1429, opened 2020-11-11
update plop for tsc and latest practices

jaketrent:
PLOP it!


plop-templates/component/src/react/index.tsx.hbs
```diff
@@ -1,40 +1,36 @@
...
+import { compose, css } from 'glamor'
+import React, { forwardRef } from 'react'
 
 import stylesheet from '../css'
 
-interface {{pascalCase name}}Props extends extends HTMLAttributes<HTMLDivElement> {}
-
-type StyleFn = (themeName:string, props: {{pascalCase name}}Props) => StyleAttribute
+type DivEl = JSX.IntrinsicElements['div']
+interface {{pascalCase name}}Props extends DivEl {}
```
danethurber, line 9: have we settled on using an alias?

danethurber (approved):

plop-templates/component/src/react/index.tsx.hbs
jaketrent: Dunno if we've settled on that, but it is the latest practice.  Seems like we've settled on moving away from HTMLAttributes.

Should we settle?

If so, on PropsOf or alias?

PR #1432, opened 2020-11-11
Fix broken calendar on datepicker

jaketrent:
Well this is embarrassing.  Somehow got removed in TS conversion.  

Wrongly removed in 3751cee52e

### DIFF
--- a/packages/datepicker/src/react/__specs__/index.spec.tsx
+++ b/packages/datepicker/src/react/__specs__/index.spec.tsx
@@ -28,7 +28,7 @@ describe('DatePicker', () => {
     const { getByText, getByLabelText, container, rerender } = render(
       <DatePicker value="1/20/1993" />
     )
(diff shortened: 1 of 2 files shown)


packages/datepicker/src/react/__specs__/index.spec.tsx
```diff
@@ -24,6 +24,19 @@ describe('DatePicker', () => {
     expect(ref.current).not.toBeNull()
   })
 
+  it('opens calendar on icon click', () => {
+    const { getByText, getByLabelText, container, rerender } = render(
+      <DatePicker value="1/20/1993" />
+    )
+    const icon = getByLabelText('calendar icon', { selector: 'svg' }) // container.querySelector<HTMLButtonElement>('button')
```
danethurber, line 31:
```suggestion
    const icon = getByLabelText('calendar icon', { selector: 'svg' })
```

danethurber (approved):

Issue #1344, opened 2020-10-16
datepicker: find onSubBlur bug


jaketrent:
### Expected behavior

Datepicker picks and reports proper picking.

### Actual behavior

A few related things:

- Will's screencast and thread: https://pluralsight.slack.com/archives/C70HPSJPP/p1602861794095000
- In Will's experience, `null` always returned, even when date is valid/full.
- In Jake's experience, `null` always returned except for the blur on the `mm` field, then the value is returned
- In test, the onSubBlur tests pass, but upon console logging, the internal behavior is odd (see below) (ie, why would you ever set nextDate as anything *less* than a full date.
- The predictability of what happens on blur just feels fragile.  Some thing is not quite right.
- There is an uncontrolled-to-controlled error from react on blur (see below)

TEST
```
~  13     it.only('should call onBlur when onBlur is called for a valid date', () => {
   12       const onBlurMock = jest.fn()
   11       const { getByDisplayValue } = render(
   10         <DatePicker value="8/14/2001" onSubBlur={onBlurMock} />
    9       )
~   8       fireEvent.blur(getByDisplayValue('14'))
    7       expect(onBlurMock).toHaveBeenCalledTimes(1)
    6       expect(onBlurMock).toHaveBeenCalledWith('8/14/2001')
+   5       fireEvent.blur(getByDisplayValue('14'))
+   4       expect(onBlurMock).toHaveBeenCalledTimes(2)
+   3       expect(onBlurMock).toHaveBeenCalledWith('8/14/2001')
+   2       fireEvent.blur(getByDisplayValue('14'))
+   1       expect(onBlurMock).toHaveBeenCalledTimes(3)
+ 41        expect(onBlurMock).toHaveBeenCalledWith('8/14/2001')

OUTPUTS:
    console.log
      { currentDateOverwrittenByEventValue: { dd: '14', mm: 8, yyyy: 2001 },
        alwaysReValidateDay: 14,
        nextDate: { dd: 14 },
        onBlurDate: { dd: 14, mm: 8, yyyy: 2001 } }

      at handleSubFieldBlur (packages/datepicker/src/react/index.js:164:13)

    console.log
      { currentDateOverwrittenByEventValue: { dd: '14' },
        alwaysReValidateDay: 14,
        nextDate: { dd: 14 },
        onBlurDate: { dd: 14 } }

      at handleSubFieldBlur (packages/datepicker/src/react/index.js:164:13)

    console.log
[...]

PR #1437, opened 2020-11-12
DatePicker Date type refactor

jaketrent:
Resolves: #1344

- onSubBlur should eventually return valid date
- date value changes in and out of component seem to propogate correctly into and out of
- got rid of date wannabe for native Date
- cleaned up on handleSubBlur quite a bit, but still more complicated that I wanted
- got rid of uncontrolled to controlled react warning. is now always controlled

BREAKING CHANGE: Must be a controlled component via supplied value prop
BREAKING CHANGE: onSelect and onSubBlur signatures change
BREAKING CHANGE: input and output values are of Date type


packages/datepicker/src/react/__stories__/index.story.tsx
```diff
@@ -26,77 +29,144 @@ function StateDemo(props: { value?: string }) {
...
+    <DatePicker subLabel="Some sublabel" value={undefined} />
+  ))
   .add('label and subLabel', () => (
-    <DatePicker label="Some label" subLabel="Some sublabel" />
+    <DatePicker label="Some label" subLabel="Some sublabel" value={undefined} />
   ))
 
 storiesOf('value', module)
-  .add('single slash-separated value', () => <DatePicker value="12/07/1941" />)
+  .add('date', () => <DatePicker value={new Date(1941, 11, 7)} />)
```
EdwardIrby, line 58: I like this new format

packages/datepicker/src/react/index.tsx
```diff
@@ -65,43 +66,29 @@ const styles = {
...
-    newDateString: string | undefined,
-    newDateParts?: DateParts
-  ) => void
-  onSubBlur?: (date: string | undefined, evt: React.FocusEvent) => void
-  subLabel?: React.ReactNode
-  value?: string
+  onSelect?: (evt: React.FocusEvent | React.MouseEvent, newDate: Date) => void
+  onSubBlur?: (evt: React.FocusEvent, date: Date | undefined) => void
+  subLabel?: ReactNode
+  value: Date | undefined
```
EdwardIrby, line 91: If value can be undefined why not make it optional. Optional resolves too undefined I believe.

packages/datepicker/src/react/sub-field.tsx
```diff
@@ -0,0 +1,62 @@
...
+interface SubFieldProps
+  extends Omit<
+    React.HTMLAttributes<HTMLInputElement>,
+    'onBlur' | 'onChange' | 'onFocus'
+  > {
+  appearance: ValueOf<typeof vars.appearances>
+  disabled?: boolean
+  name: string
+  onBlur: (evt: React.FocusEvent) => void
+  onFocus: (evt: React.FocusEvent) => void
```
EdwardIrby, line 34: I think you can use React.FocusEventHandler here

packages/datepicker/src/react/sub-field.tsx
```diff
@@ -0,0 +1,62 @@
...
+  value: number | undefined
```
EdwardIrby, line 35: This can be optional as well `value?: number`

packages/datepicker/src/react/calendar.tsx
```diff
@@ -46,43 +42,28 @@ const styles = {
 }
 
 interface CalendarProps {
-  date: DateParts
-  onSelect: (value: string | undefined) => void
+  value: Date | undefined
```
EdwardIrby, line 45: This can be optional as well `value?: Date`

packages/datepicker/src/react/__stories__/index.story.tsx
jaketrent: yesssss! :D

packages/datepicker/src/react/index.tsx
jaketrent: In order to ensure that it's a controlled component.  In master now, there are warnings of switching from uncontrolled to controlled.  Didn't go down the path of why enough to get to root cause, but I thought this would solve the problem well enough.  It seems to have.

These typings are different:

The value prop must exists.  One value can be undefined:
`value: Date | undefined`

The value prop does not have to be passed at all.
`value?: Date`

packages/datepicker/src/react/sub-field.tsx
jaketrent: if the explanation above for the DatePickerProps passes muster, I think it'll apply here too.  Let me know.

packages/datepicker/src/react/calendar.tsx
jaketrent: Ditto.

packages/datepicker/src/react/index.tsx
jaketrent: Related:
https://stackoverflow.com/questions/50490773/why-doesnt-typescript-undefined-type-behave-same-as-optional
https://rules.sonarsource.com/typescript/RSPEC-4782

jaketrent: I thought about changing it, because I do like the `value?: Date` syntax better.  But I think `value: Date | undefined` expresses my intent better: that this is a controlled component, always plan to send a value by coding in a value prop.  

The valid case of undefined will probably end up being coded in an app something like this:

```
const App: React.FC = () => {
  const [selected, setSelected] = useState<Date>()

  return <DatePicker value={selected} onSelect={(_evt, date) => setSelected(date)} />
}
```

This also seems to match the current typing well.

EdwardIrby (approved):

packages/datepicker/src/react/index.tsx
EdwardIrby: While I don't necessarily agree with that. Do what you think is best.

I feel eventually someone more familiar with typescript will file a bug against this when they are forced to do this ```<DatePicker
      onSelect={action('onselect')}
      onSubBlur={action('onblur')}
      value={undefined}
    />``` 

In the above snippet there is no reason to pass in undefined for value if it was just optional. It is not necessary for the functioning of the component and creates more boilerplate and seems like an anti-pattern.

jaketrent: ^ This is a great counter point (component can work without it with callbacks and the undefined value can look wrong).  I'll let this sit over the weekend and stew on it.  Thank you.

PR #1442, opened 2020-11-12
test(datepicker): timezone mock, lint and type fixes

EdwardIrby:
- Mock timezone for consistent test snapshots
- fix linting errors `'==' => '==='` in **packages/datepicker/src/js/index.ts**
- change type | undefined to key?: as it resolves to type | undefined

### DIFF
--- a/jest/setup/react-dom.js
+++ b/jest/setup/react-dom.js
@@ -1,9 +1,12 @@
 import ReactDOM from 'react-dom'
+import MockDate from 'mockdate'
 
 beforeAll(() => {
   ReactDOM.createPortal = jest.fn(el => el)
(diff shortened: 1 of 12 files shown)


EdwardIrby:
Seems my test fix does not work on the CI. Back to drawing board.

jaketrent:
In my branch, I changed approaches to use the ISO8601 date format in the form input, per MDN:

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date

This gets rid of the TZ issue entirely.  Yay!

I've lifted the linting adjustments that you pinpointed here out of your diff.  Thank you!

Closing.

PR #1446, opened 2020-11-13
HTMLPropsFor conversion 2/3

jaketrent:
For the middle third of components plus a few typings conversion-related fixes along the way.

- refactor(emptystate): use htmlpropsfor
- refactor(errors): use htmlpropsfor
- refactor(focusmanager): use htmlpropsfor and reffor
- refactor(form): use htmlpropsfor
- refactor(halo): use htmlpropsfor and reffor
- refactor(icon): use htmlpropsfor
- refactor(layout): use htmlpropsfor
- refactor(linearprogress): use htmlpropsfor and update npm-scripts
- refactor(navbar): use htmlpropsfor
- refactor(navcookiebanner): use htmlpropsfor
- refactor(note): use htmlpropsfor
- refactor(position): use htmlpropsfor
- refactor(radio): use htmlpropsfor and fix spec typings


packages/radio/src/react/__specs__/index.spec.tsx
```diff
@@ -29,7 +29,7 @@ describe('Radio', () => {
       </Radio.Group>
     )
     const radio = container.querySelector('input')
```
danethurber, line 31:
```suggestion
    const radio = container.querySelector('input')!
```

danethurber: don't ask me why...but you're supposed to do the bang on this line instead of the next

danethurber (approved):
💥

PR #1450, opened 2020-11-13
ActionMenu strict typings

danethurber:
and various final touches


packages/actionmenu/src/react/index.tsx
```diff
@@ -1,62 +1,80 @@
...
+      children,
+      ...rest
+    },
+    forwardedRef
+  ) => {
+    const ref = React.useRef<HTMLUListElement | null>(null)
+    useImperativeHandle<HTMLUListElement | null, HTMLUListElement | null>(
+      forwardedRef,
+      () => ref.current
+    )
```
danethurber, line 76: found a way to avoid casting the ref...not sure it's any better 🤷

EdwardIrby: looks a bit better

EdwardIrby (approved):

packages/actionmenu/src/react/item.tsx
```diff
@@ -47,15 +52,29 @@ const styles: { [name: string]: StyleFn } = {
...
+  nested?: ReactNode
+  origin?: ValueOf<typeof origins>
+  value?: ReactText
+  onClick?: (event: MouseEvent, value?: ReactText) => void
+}
+
+interface AnchorProps
+  extends BaseItemProps,
+    Omit<HTMLPropsFor<'a'>, 'onClick'> {
+  tagName?: 'a'
```
jaketrent, line 66: Why introduce another patten for this and not use the existence of href or the renderProp patterns that we have so far?

packages/actionmenu/tsconfig.json
```diff
@@ -3,6 +3,7 @@
   "compilerOptions": {
     "allowJs": true,
     "rootDir": null,
-    "jsx": "react"
+    "jsx": "react",
+    "strict": true
```
jaketrent, line 7: nice catch!

packages/actionmenu/src/react/item.tsx
danethurber: to avoid a breaking change. this is the pattern that was used in ActionMenu - i just added the typings

PR #1452, opened 2020-11-16
fix(typeahead): allow selection with tab, enter and space keys

jaketrent:
I wanted to drive with a test something like this, but I never got it working.  I couldn't figure out why input's value was never updated on Enter.  All the elements were found, just seemed like that final step didn't show up for some reason.  Kind of funny there are no other keyboarding tests in the suite either.  My current best guess is there's some sort of timing problem with the Position wrapper. 

``` 
  import { logDOM } from '@testing-library/dom'
  import userEvent from '@testing-library/user-event'
  it('selects the focused element upon pressing enter', async () => {
    const { getByLabelText } = render(
      <Typeahead label="inputlabel">
        <Typeahead.Suggestion>primo</Typeahead.Suggestion>
        <Typeahead.Suggestion>second</Typeahead.Suggestion>
        <Typeahead.Suggestion>secondino</Typeahead.Suggestion>
      </Typeahead>
    )
    const input = getByLabelText('inputlabel')

    input.focus()
    fireEvent.change(input, { target: { value: 'sec' } })

    logDOM(document.body)
    userEvent.tab()
    fireEvent.keyDown(document.activeElement, { key: 'Enter', code: 13 })

    expect(input).toHaveValue('second')
  })
```

### DIFF
--- a/packages/typeahead/src/react/menu.tsx
+++ b/packages/typeahead/src/react/menu.tsx
@@ -20,7 +20,12 @@ const styles = {
 
 interface SuggestionsMenuProps extends Omit<HTMLPropsFor<'div'>, 'onChange'> {
   activeValue?: string
-  onChange: (evt: React.MouseEvent<HTMLButtonElement>, value: string) => void
+  onChange: (
+    evt:
+      | React.MouseEvent<HTMLButtonElement>
+      | React.KeyboardEvent<HTMLDivElement>,
+    value: string
+  ) => void
   onFocus?: React.FocusEventHandler<HTMLDivElement>
   onKeyDown?: React.KeyboardEventHandler<HTMLDivElement>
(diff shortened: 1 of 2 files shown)


packages/typeahead/src/react/menu.tsx
```diff
@@ -87,13 +92,17 @@ const SuggestionsMenu = forwardRef<HTMLDivElement, SuggestionsMenuProps>(
     }, props.onFocus)
 
     const handleKeyDown = combineFns(evt => {
-      const tab = evt.key === 'Tab'
-      if (tab) return
-
-      if (evt.key === 'ArrowDown') next()
+      console.log({ key: evt.key })
```
danethurber, line 95:
```suggestion
    
```

danethurber (approved):

PR #1456, opened 2020-11-16
Remove unused deps after TS convert

jaketrent:
- prop-types
- filter-prop-types


danethurber (approved):
looks good. better merge it before i merge the ActionMenu changes ...

jaketrent:
^ For real?  Is there an order you'd prefer for actionmenu-something?

danethurber:
no. you merge first. i'm happy to do the conflicts

jaketrent:
k, I'll wait til the pipeline is green, then merge.

packages/actionmenu/src/react/item.tsx
```diff
@@ -12,7 +12,7 @@ import React, {
 } from 'react'
 
 import stylesheet from '../css'
-import { tagName } from '../vars'
+import { tagName as tagNames } from '../vars'
```
EdwardIrby, line 15: Should we just rename this tagNames

EdwardIrby (approved):

packages/actionmenu/src/react/item.tsx
jaketrent: nbd, but imo, yes. I avoided just so that I didn't have to bump the major version. :)

PR #1464, opened 2020-11-18
refactor(datepicker): make subfields available to accessibility tree

jaketrent:
- Inputs already textbox in accessibility tree
  https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/textbox_role
- But needed better, now individual labelling
- Following WAI subfield labelling suggestion
  https://www.w3.org/TR/WCAG20-TECHS/ARIA14.html
- Using getByRole prioritized testing-library query
  https://testing-library.com/docs/guide-which-query/
- getByRole(roleName, { name }) is non-intuitive: name is a label, not
  an html attr
  https://testing-library.com/docs/dom-testing-library/api-queries/#byrole


danethurber (approved):
👍

PR #1465, opened 2020-11-18
<Table /> refactor

### LINKED ISSUE #1362: table: make copy-paste experience better for excel users
Source: Ben in plan analytics
https://pluralsight.slack.com/archives/C70HPSJPP/p1603836664143200

Before using the DS table, users could copy the on-screen table's content by selection and paste nicely into Excel.  Now, they cannot after switching to the DS table.  This is presumably because the DS table uses divs instead of table tags.

Explore the options for making copy-paste better. 

Consider whether the original reasons for not using table on the DS table are valid or attainable in other ways.
, the reasons being:
- CSS styling freedom
- mobile reflow flexibility

danethurber:
Built from stratch, this new Table component uses the latest and greatest in accessible tabular
markup and supports advanced features like native browser copy/paste and auto table layout.

BREAKING CHANGE: a new set of components that should be used like native html tables

fix #1362

---

TODO:
  - [x] new styles
  - [x] expandable drawer
  - [x] add forwardRefs
  - [x] ie11 supported styling(drop css vars 😞  )
  - [x] horz scroll for responsive
  - [ ] ~sticky header that works with horz scoll....~ 🚫 
  - [x] kitchen sink example with react-table
    - pagination
    - filtering
    - sorting
    - bulk actions and row selection
  - [x] a11y audit
  - [x] moar browser testing
  - [x] react-beautiful-dnd examples
  - [ ] ~write test for copy/paste~ 🚫 
  - [x] documentation


packages/storybook-addon-center/src/index.tsx
```diff
@@ -8,6 +8,8 @@ export const withCenter = makeDecorator({
   name: ADDON_ID,
   parameterName: PARAM_KEY,
   wrapper: (storyFn, context, { parameters }) => {
+    if (parameters?.disabled) return storyFn(context)
```
jaketrent, line 11: Why add a disabled param? Why not just not use Centered in your storybook setup?

packages/table-next/src/react/__stories__/react-table.story.tsx
```diff
@@ -0,0 +1,393 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
jaketrent, line 1: Should we disable eslint on spec files entirely?

packages/table-next/src/react/index.tsx
```diff
@@ -0,0 +1,158 @@
...
+import { compose, css } from 'glamor'
+import React from 'react'
+
+import stylesheet from '../css'
+import { aligns, sorts } from '../vars'
+
+const styles = {
+  container: (
+    themeName: ValueOf<typeof themeNames>,
+    opts: { scrollable?: boolean }
```
jaketrent, line 20: Looks like object not needed. Just pass bool?

packages/table-next/src/react/index.tsx
```diff
@@ -0,0 +1,158 @@
...
+import {
+  names as themeNames,
+  useTheme
+} from '@pluralsight/ps-design-system-theme'
+import { HTMLPropsFor, ValueOf } from '@pluralsight/ps-design-system-util'
+import { compose, css } from 'glamor'
+import React from 'react'
+
+import stylesheet from '../css'
+import { aligns, sorts } from '../vars'
```
jaketrent, line 15: `alignments` might be more standard naming here?

packages/table-next/src/react/index.tsx
```diff
@@ -0,0 +1,158 @@
...
+  Row: typeof TableRow
+}
+
+interface TableProps extends HTMLPropsFor<'table'> {
+  renderContainer?: React.FC<React.HTMLAttributes<HTMLElement>>
+  scrollable?: boolean
+}
+const Table: React.FC<TableProps> & TableStatics = props => {
+  const {
+    renderContainer: Container = defaultRenderContainer,
```
jaketrent, line 80: What are you anticipating by the ability here to change from a div container?

packages/table-next/typings/react-table-config.d.ts
```diff
@@ -0,0 +1,133 @@
+//
+// NOTE: https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/react-table
```
jaketrent, line 2: If this is in DefinitelyTyped, why copy it here?

packages/storybook-addon-center/src/index.tsx
danethurber: i wanted centered on some stories but wanted to disable the addon on only some stories. this is the general approach that i really should have added in the beginning

packages/table-next/src/react/__stories__/react-table.story.tsx
danethurber: maybe certain rules. we'd still want eslint for typescript linting since the spec/stories are the best way we current have to validate the typings/props/etc

packages/table-next/src/react/index.tsx
danethurber: no reason. just habit of passing objects to the style functions like we've done in other components

packages/table-next/src/react/index.tsx
danethurber: mostly exposing the div so consumers can add props(styles, aria, testids, etc) and refs when needed

packages/table-next/typings/react-table-config.d.ts
danethurber: it's a weird package. the instructions and reasonings are outlines in that link

https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/react-table#configuration-using-declaration-merging

packages/table/src/react/index.tsx
```diff
@@ -1,270 +1,196 @@
...
 
-interface RowProps extends HTMLPropsFor<'div'> {
-  _tableHasDrawers?: boolean
+interface TableCellProps extends HTMLPropsFor<'td'> {
+  align?: ValueOf<typeof alignments>
 }
+const TableCell = forwardRef<HTMLTableCellElement, TableCellProps>(
+  (props, ref) => {
+    const { align = alignments.left, ...rest } = props
+    return <td ref={ref} {...styles.cell({ align })} {...rest} />
```
EdwardIrby, line 123: We'll need to submit a new pr for the [docs search config](https://github.com/algolia/docsearch-configs/blob/master/configs/pluralsight_design-system.json) or add role="cell" here

EdwardIrby:
Does this no longer work with drawer? I didn't see an example of this using Drawer in the story or I missed it?

packages/table/src/react/index.tsx
danethurber: the new role is `gridcell`.

TODO: update docs search config

danethurber:
@EdwardIrby we dropped the drawer because table rows can't be wrapped in a div. there are example of expand/collapse in the react-table story - but without an animating height

danethurber:
PR is ready for review. docs are still in progress but i'd like to open the react component for review

EdwardIrby:
> @EdwardIrby we dropped the drawer because table rows can't be wrapped in a div. there are example of expand/collapse in the react-table story - but without an animating height

Do we know if this will affect any teams adversely?

danethurber:
@EdwardIrby great question. this is a breaking change so there is a fair amount that teams will need to explore with the new structure. worst case scenario - they give us negative feedback and we reignite the need for a `DataGrid` component?

jaketrent (approved):
react-beautiful-table -- ship it!

Very nice, Dane.  Way to make this one shine.

PR #1466, opened 2020-11-18
Refactor/html props for edwards set

EdwardIrby:
HTMLProps for refactors


danethurber (approved):
👍 

sorry i forgot i reviewed this before. looks like i forgot to click the submit button 😬

PR #1472, opened 2020-11-23
Feat/drawer rc conversion

### LINKED ISSUE #1462: Drawer: update to use rc
In the drawer component switch to using rc vairant
- Convert RC to typescript"
- Check with design for sub-component naming preference
- Is a breaking change
- verify Row and Table still work with new drawer.

EdwardIrby:
resolve #1462


docs/src/content/components/drawer.mdx
```diff
@@ -7,13 +7,17 @@ import Drawer from '@pluralsight/ps-design-system-drawer'
...
+  The Drawer can be used as an uncontrolled or controlled component. It consist
+  of two subcomponents Summary and Details. The top-level component Drawer is headless.
+  It's puporse is to supply context and cordinate the subcomponents.
+</Intro>
 
-## Examples
+<TableOfContents {...props} />
 
-### Adaptive size
+## Examples: uncontrolled
```
jaketrent, line 18: Instead of broad, page-level categories, I think I could understand these better if they were individually marked as un/controlled, if that's helpful information, pertinent to the specific example.

docs/src/content/components/drawer.mdx
```diff
@@ -189,37 +256,16 @@ const PinkBox = props => (
   version={props.version}
 />
 
+### Drawer
```
jaketrent, line 259: I'd add the prop tables for the other subcomponents too.

docs/src/content/components/drawer.mdx
```diff
@@ -7,13 +7,17 @@ import Drawer from '@pluralsight/ps-design-system-drawer'
 
 # Drawer
 
-<TableOfContents {...props} />
+<Intro>
```
jaketrent, line 10: I would refrain from technical usage comments in the Intro. I think we want to train our users to see the intro as a description of what purpose a component serves and how it is differentiated from others.

These comments are still useful. I would put them in a paragraph with its own heading as-is.  Or, I would think about how to put these comments on technical options closer to the examples that show them off.

docs/src/content/components/drawer.mdx
```diff
@@ -96,9 +100,13 @@ const ExampleDrawerPanel = () => (
 )
 ```
 
-### Drawer base
-
-The DrawerBase may expand to fit any elements, and includes the mechanism to expose the DrawerPanel.
+### Drawer Subcomponents
```
jaketrent, line 103:
```suggestion
### Drawer subcomponents
```

packages/drawer/src/react/__specs__/index.spec.tsx
```diff
@@ -1,46 +1,52 @@
...
-  describe('when the arrow icons is clicked', () => {
-    it('toggles open/close', () => {
-      const { container } = render(
-        <Drawer base={<div />}>
-          <div />
+  describe('when uncontrolled', () => {
+    it('toggles open/close', async () => {
+      const { getByTestId } = render(
+        <Drawer>
+          <Drawer.Summary data-testid="head">
```
jaketrent, line 11: Not a huge deal in this context, but perhaps helpful generally:  I find it helpful to not create new names for things, creating the need to mentally track synonyms. In this case, if the testid is on the Drawer.Summary, it could be called "summary" instead of "head", and on line 20, one would know more easily what's being referenced on line 11. One might make the same argument for "btn" on line 20.

packages/drawer/src/react/__stories__/index.story.tsx
```diff
@@ -1,144 +1,271 @@
...
+          </DrawerSummaryContent>
+        </Drawer.Summary>
+        <Drawer.Details>
+          <DrawerBodyContent>Drawer Content here</DrawerBodyContent>
+        </Drawer.Details>
+      </Drawer>
+    </>
+  )
+}
+storiesOf('drawer:rc', module)
```
jaketrent, line 88:
```suggestion
storiesOf('drawer', module)
```

jaketrent (approved):
Looks great!  Thanks for the consolidation work here.

Made a few comments inline about things that might be good to tweak.

docs/src/content/components/drawer.mdx
EdwardIrby: They don't take anything but whatever div can take and children?

packages/drawer/src/react/__specs__/index.spec.tsx
EdwardIrby: forgot to refactor that

PR #1473, opened 2020-11-23
Dropdown A11y and Testingpalooza

### LINKED ISSUE #1395: Dropdown: label not selectable using testing-library
using `getByLabelText` in a testing library test throws an error. testing-library is the preferred approach because it mimics user behaviour and the a11y tree. we should investigate, and update the component, to insure we are a11y compliant.

jaketrent:
Resolves #1395 

- Makes Dropdown accessible by label - wrap only label text
- Bunch of new tests for *in-DOM functionality* on the Dropdown
- Feature for ActionMenu - style items as "focused" without :focus event
- A couple tweaks to ActionMenu - span as child of a/button
- Stop supporting nested menus in Dropdown - simplify logicz
- Breaking changes: markup, onClick interface, Item only takes ReactText
- Followed [WAI-ARIA-1.2 recommendation for combobox](https://www.w3.org/TR/wai-aria-practices-1.2/examples/combobox/combobox-autocomplete-list.html), including markup, roles, focus strategy, keyboard support
  - Added backing input field
  - Button still opens menu
  - After menu is opened, focus doesn't leave input; uses aria-activedescendant
- Escape keyboard support was different than the rest. Would love it not to be.
- Learned a ton about testing-library. Will doc some key findings.
- Resolved bug where selected item wasn't focused on menu open

Dropdown still needs:
- Wondering if useDropdown really belongs in js/ dir or should be in react/ dir because of React api use
- Realized tsc strict is false; will card it


packages/actionmenu/src/react/item.tsx
```diff
@@ -178,10 +179,10 @@ export const Item = forwardRef<HTMLLIElement, ItemProps>(
         tabIndex={!disabled ? -1 : undefined}
       >
         <Wrapper {...styles.item({ hasSubMenu })} aria-haspopup={!!nested}>
-          <div className={className} {...styles.inner()}>
+          <span className={className} {...styles.inner()}>
```
EdwardIrby, line 182: Why the tag change?

EdwardIrby (commented):
I think the changes made too ActionMenu should be reverted. We should copy over the ActionMenu Component and Styles to dropdown and make those changes there, perhaps? We discussed at planning doing this. Most of these changes seems Dropdown specific. I'll take some more time to dig through the code today and finish my review later.

packages/actionmenu/src/react/item.tsx
jaketrent: So that it's valid markup.

Divs are invalid here https://validator.w3.org/nu/#textarea

Apparently related section of the spec is 3.2.5.1.5 "phrasing content" https://www.w3.org/TR/2012/WD-html5-20120329/content-models.html#phrasing-content

jaketrent:
Changes to ActionMenu made here are:
- Remove unused css code
- Add `active` prop for focus styling without :focus
- Adjust markup to html spec

These seem like generally-good and helpful changes that any user of ActionMenu would want.

The feature add of `active` here is going to be helpful to anyone who uses ActionMenu in a custom compound component where they want to follow a similar WAI recommendation on focus strategy for compound widgets.

I would recommend not rolling these back for these reasons.

Yes, I think that separating Dropdown from ActionMenu is still a possibility. I treated this as a next step in the current PR. Instead, I will look at this morning so we can uncover details and implications sooner.

packages/actionmenu/src/react/item.tsx
```diff
@@ -25,19 +25,18 @@ import { ActionMenuContext } from './context'
 import { Arrow } from './arrow'
 
 const styles = {
-  itemContainer: ({ disabled }: { disabled?: boolean }) =>
+  itemContainer: ({
+    active,
+    disabled
+  }: Pick<BaseItemProps, 'active' | 'disabled'>) =>
```
EdwardIrby, line 31: Nice! Been forgetting that one myself.

packages/dropdown/src/js/__specs__/index.spec.tsx
```diff
@@ -1,62 +1,95 @@
...
+    expect(getLongestMenuLabel([])).toEqual('')
+  })
+
+  it('is placeholder if no items', () => {
+    expect(getLongestMenuLabel([], 'aPlace')).toEqual('aPlace')
+  })
+
+  it('is placeholder if still longest', () => {
+    expect(
+      getLongestMenuLabel(
```
EdwardIrby, line 80: A thing I've noticed is that we have edge cases where the character count isn't enough to determine longest. Sometimes we have words whose character width makes them the longest. Should we address this at some point?

packages/dropdown/src/js/index.ts
```diff
@@ -85,168 +98,218 @@ const sortDropdownProps = ({
...
     evt.preventDefault()
     evt.stopPropagation()
-    setOpen(!isOpen)
-    if (typeof hook.onClick === 'function') hook.onClick(evt)
-  }
 
-  function handleKeyDown(evt: React.KeyboardEvent) {
-    if (evt.key === 'ArrowDown') {
-      setOpen(true)
+    if (isOpen) {
```
EdwardIrby, line 174: Wondering if we just need the ActionMenu markup and styles and not the key handlers on it?

EdwardIrby: Also we seem to have lost keyboard events on the menu. I can't key through them anymore after running clean build and then storybook.

EdwardIrby (changes requested):

packages/dropdown/src/js/__specs__/index.spec.tsx
jaketrent: Probably.  This has gotten us pretty far. Haven't heard complaints about this aspect thus far.

What are the edge cases?  We could start by recording those in stories/test.

packages/dropdown/src/js/index.ts
jaketrent: Not sure what you mean by this:

> Wondering if we just need the ActionMenu markup and styles and not the key handlers on it?

Mostly because we need keyboard support, and that would seem to remove it.

On your 2nd comment:

> seem to have lost keyboard events on the menu

What keyboard events have you lost specifically?  Keyboard support has been a large focus of this refactor, so if I'm messing that up, I'm way off the mark.  Should be supported:

- Tab - focus button
- Enter/Space/ArrowDown - activate button to open menu 
- ArrowDown/ArrowUp - move to items in menu
- Enter/Space - select item 
- Escape - close menu

I have just freshly rebuilt and run test and storybook.  Auto and manual tests are looking good for me.  Please confirm and further specify the problem you're seeing.

jaketrent:
Hey @EdwardIrby , I've just pushed the removal of the ActionMenu from Dropdown and commented on what I think your change request stems from.  Please have a look at my responses and new code.  Thanks!

jaketrent:
@EdwardIrby Ok, ready for another round: the value-less items will now be selected.  They were before, it was just the styling that was unapparent. But there was an actual value setting in that case. Also found a hole in the lack of calling onChange when keyboarding. Tests ftw!  Thanks for your help!

EdwardIrby (approved):

PR #1478, opened 2020-11-30
variable font playground

jaketrent:
- Added example we used with TypeType
- Move examples/ into packages/ so that linking will work (of course, now that we have yarn workspaces doing the linking, maybe this isn't required (a la docs)).  Somebody correct me on this if not needed.


danethurber:
you could also add the `examples` dir to the workspaces and lerna config

jaketrent:
would that be your preference? ^

danethurber:
yes that would be my pref. also the current implementation doesn't work because the projects are nested inside packages/examples. for yarn linking they would need to NOT be nested since our config just looks at the packages/* dir

danethurber (changes requested):
i'm holding of on reviewing until the examples find their new home. i'm unable to run a `yarn build` from the root. i'm getting a error

jaketrent:
@danethurber examples moved back to root, added to lerna config for linking, and root build looks good.  Please have a re-look.

examples/variable-font/README.md
```diff
@@ -0,0 +1,23 @@
+
+## To run
+
+Copy font files into this directory so it looks like this:
+
+```
+.
+├── font-demo.html
+├── fonts
```
danethurber, line 9: we should .gitignore this folder just to be safe

examples/variable-font/README.md
```diff
@@ -0,0 +1,23 @@
...
+│   └── ps-tt-commons
+│       ├── PS_TT_Commons_Var_Italic-VF.ttf
+│       ├── PS_TT_Commons_Var_Italic-VF.woff
+│       ├── PS_TT_Commons_Var_Italic-VF.woff2
+│       ├── PS_TT_Commons_Var_Roman-VF.ttf
+│       ├── PS_TT_Commons_Var_Roman-VF.woff
+│       └── PS_TT_Commons_Var_Roman-VF.woff2
```
danethurber, line 16: FYI. these filenames are slightly different than the files i have

danethurber (changes requested):
please review/merge my pr -> https://github.com/pluralsight/design-system/pull/1483

jaketrent:
Hey @danethurber , I somehow hosed your #1483 branch.

I just ran `git cherry-pick b301ddc5f25408cf78f23709163f5cb92c0b465e` on this branch.  Should be good to roll now. 

Please re-review.

danethurber (approved):

PR #1480, opened 2020-12-01
add ps tt commons to normalize

jaketrent:
- [x] Need to go back and fix TODOs (with real Cloudfront URL of font)
- Decided not to major version bump core or normalize. Reasoned that we could push the new packages to people transparently and remain backwards compatible with Gotham. This will help us transition faster. Gotham and the adjusted PS TT Commons are meant to be as identical in size and proportions as possible as to not require much visual tweaking. Tweaks that might need to be made are anticipated to be small and manageable.


packages/core/css/type.module.css
```diff
@@ -1,6 +1,7 @@
 :root {
-  --psTypeFontFamily: "Gotham SSm A", "Gotham SSm B", sans-serif;
-  --psTypeFontFamilyCode: "Source Code Pro", monospace;
+  /* prettier-ignore */
+  --psTypeFontFamily: 'PS TT Commons', 'Gotham SSm A', 'Gotham SSm B', sans-serif;
```
danethurber, line 3: 👍 nice. didn't think about falling back to gotham but good idea.

@sweeting i'm wondering what the ie11 fallback should be

jaketrent: Thought we had said that IE gets sans-serif.

danethurber: we did. i was just wondering if there was something more specific. like 'Arial'...

jaketrent: Gotcha. I like going straight to sans-serif just to get the system font that's OS-specific.  

That's what `sans-serif` is to me conceptually, but I did also see that there's a new `system-ui` option, which I don't yet understand how it differs.

sweeting: Arial might be a less noticeable difference as fallback for IE 11 than the Windows System font Segoe UI (a humanist sans).

Sample Page - PS TT Commons

Sample Page - Segoe UI

Sample Page - Arial

jaketrent:
Still ready for review.  Red tests are unrelated.  Polyfill fix will come after merge of https://github.com/pluralsight/design-system/pull/1485

danethurber (approved):

PR #1481, opened 2020-12-01
add keyboard tests to datepicker

### LINKED ISSUE #1469: datepicker: make userEvents respond well in test
Two separate issues:
1) userEvent.tab() does not work as expected when tabbing between mm, dd and yyyy. It always focuses back to the body (I’ve also tried using focusTraps to no avail)
2) userEvent.type() for dd and yyyy is not working (even when not tabbing through) (edited) 

See:
https://pluralsight.slack.com/archives/C70HPSJPP/p1605804091239800?thread_ts=1605632188.221700&cid=C70HPSJPP

jaketrent:
Resolves #1469

These tests seem to address the issues reported. No problem in these tests afaik. Want to merge and send to Will to see if they work for him.


danethurber (approved):
👍

PR #1483, opened 2020-12-01
build: add examples to yarn workspaces

danethurber:
@jaketrent just a couple little tweaks needed to use the examples with yarn workspaces


lerna.json
```diff
@@ -4,7 +4,7 @@
     "version": {
       "allowBranch": "master",
       "conventionalCommits": true,
-      "ignore": ["docs"],
+      "ignore": ["docs", "./examples/*"],
```
danethurber, line 7: don't version examples

package.json
```diff
@@ -14,6 +14,7 @@
     "nohoist": [],
     "packages": [
       "docs",
+      "examples/**",
```
danethurber, line 17: yarn linking

examples/create-react-app/package.json
```diff
@@ -3,7 +3,6 @@
   "version": "0.1.0",
   "private": true,
   "scripts": {
-    "build": "react-scripts build",
```
danethurber, line 6: removed since we never need to build the examples and don't want to wait for this when running `yarn build`

jaketrent (approved):

PR #1492, opened 2020-12-04
Chore/dep upgrades

danethurber:
update test, lint, and storybook dependencies.

there are some deprecation warnings we'll need to silence in the future


docs/gatsby-config.js
```diff
@@ -5,6 +5,12 @@ const path = require('path')
 const remarkSlug = require('remark-slug')
 
 module.exports = {
+  flags: {
+    DEV_SSR: true,
+    FAST_DEV: true,
```
jaketrent, line 10: this is experimental, right?

packages/actionmenu/jest.config.js
```diff
@@ -1,9 +1,11 @@
...
 const baseConfig = require('../../jest/base.config.js')
-const { name } = require('./package')
+const { name } = require('./package.json')
 
 module.exports = {
   ...baseConfig,
   displayName: name,
   name: name,
-  testMatch: [`${__dirname}/**/*/?(*.)+(spec|test).(ts|tsx)`]
+  testMatch: [path.join(__dirname, '/**/*/?(*.)+(spec|test).(js|ts|tsx)')]
```
jaketrent, line 10: really not a great reason to *partially* use path.join.  It'll be OS-specific only for that first separator.  But we still have forward-slashes hardcode after that.

jaketrent (approved):

packages/actionmenu/jest.config.js
danethurber: yeah, it's some new linting rule. i think it's when used with `__dirname` 🤷

docs/gatsby-config.js
danethurber: yes. gatsby gave me a message that i should try these out...so i'm trying them out :)

PR #1493, opened 2020-12-04
datepicker date change updates

jaketrent:
- Several breaking API changes
  - onChange instead of onSelect
  - onSubBlur removal
  - onChange receives ChangeEvent instead of FocusEvent in union
- Fix test problem where re-typing a date after a pre-selected date would cause loss of focus and lead to wrong values
- Uncovered existing userEvent.type bug that we can work around but should be suspect for typing tests

After we ship, I'll send Will his updated test that works because of these internal DatePicker changes.  But, also changed:

- Use `screen` because queries don't need scoped to subtree
- Use `getByRole` because there will only be one of what we're filtering to
- Use `toHaveValue` because our codebase didn't have the `to.have.` chains setup
- Use `document.activeElement` in tab tests, otherwise tabbing is less indicative of what's happening
- Updated `dd` initial value to avoid uncovered testing-library [bug](https://github.com/testing-library/user-event/issues/387)

```
  it.only('wills test', async () => {
    render(<DatePicker value={new Date(2019, 0, 2)} />)
    const mm = screen.getByRole('textbox', { name: 'month' })
    const dd = screen.getByRole('textbox', { name: 'day' })
    const yyyy = screen.getByRole('textbox', { name: 'year' })
    expect(mm).toHaveValue('1')
    expect(dd).toHaveValue('2')
    expect(yyyy).toHaveValue('2019')
    mm.focus()
    userEvent.type(document.activeElement!, '07')
    userEvent.tab()
    userEvent.type(document.activeElement!, '15')
    userEvent.tab()
    userEvent.type(document.activeElement!, '2015')
    userEvent.tab()
    expect(mm).toHaveValue('7')
    expect(dd).toHaveValue('15')
    expect(yyyy).toHaveValue('2015')
  })

```


EdwardIrby (approved):
That's an annoying bug if browser testing were faster or we felt like migrating I'd recommend unit testing in cypress  so we could just use the real dom.

PR #1511, opened 2020-12-10
add page heading patterns to docs

jaketrent:
- Used frame labels in figma so that designers and developers are hopefully talking the same language.
- Called the last pattern in figma "Basic alternative". It was unlabeled in figma.
- Seems like some explanatory text would be useful. I see patterns without context. Oooh, I like this one, it has all the widgets I need. For instance, how would one decide which of the "basic" patterns he should use? One has an icon and fewer buttons, but that doesn't seem to indicate enough to be educated about the existence of 2 similar choices.

### DIFF
--- a/docs/src/components/code-block/package-map.ts
+++ b/docs/src/components/code-block/package-map.ts
@@ -48,7 +48,7 @@ import Tag from '@pluralsight/ps-design-system-tag'
 import * as Text from '@pluralsight/ps-design-system-text'
 import TextArea from '@pluralsight/ps-design-system-textarea'
 import TextInput from '@pluralsight/ps-design-system-textinput'
-import Theme from '@pluralsight/ps-design-system-theme'
+import Theme, { useTheme } from '@pluralsight/ps-design-system-theme'
 import Tooltip from '@pluralsight/ps-design-system-tooltip'
 import Typeahead from '@pluralsight/ps-design-system-typeahead'
 import VerticalTabs from '@pluralsight/ps-design-system-verticaltabs'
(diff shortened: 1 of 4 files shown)


sweeting (commented):
It might be better to call these Headers instead Headings, since we have a text component called Heading.

jaketrent:
@sweeting "Header" is fine with me.  Would you like that applied to the side nav and url too, I suppose?

sweeting:
@jaketrent Sorry for the delayed response, yes please.

jaketrent:
@sweeting I made the "heading" -> "header" adjustment.

After thinking about it this morning, I think we should keep it as "heading", because "Page heading" patterns is exactly what it says.  We do have a Heading component.  These are common patterns os what it looks like to put those at page level. Calling these patterns "headers" introduces another word that has no precise meaning in our DS, leaving me more unsure on when to use what name to communicate clearly about these.

sweeting:
@jaketrent I'm not concerned with adding a word (though similar) if it helps distinguish between two different. A "Heading" is a piece of text that helps create visual hierarchy. A "Header" is a collection of things at the top of a page or section. Let's not use the same word for things that have two different purposes. Also, the design for these patterns has always used the word "Header" and there is an html tag called `[<header>](https://html.com/tags/header/)` that is specifically intended for this purpose, which it might be good to use in the examples.

jaketrent:
>  A "Heading" is a piece of text that helps create visual hierarchy. A "Header" is a collection of things at the top of a page or section. 

@sweeting sounds good.  Thank you for identifying the two things and how you see them.

danethurber (approved):
👍  good enough for me

PR #1513, opened 2020-12-10
refactor(core): minor opacity adjustments to text colors

sweeting:
High is now 90% instead of 95%
Low is now 65% for both light and dark theme (before it was 55% on light)


EdwardIrby (approved):
need to run `yarn test -u` or `npm test -- -u`

PR #1519, opened 2020-12-11
Refactor/text core normalize

### LINKED ISSUE #1502: docs: font weights of headings and strong text are incorrectly rendered in Chrome following the normalize update
Following the update to normalize, the heading weights in Chrome are lighter than the font-weight they are set to which is `600`. 

It also appears than I'm unable to adjust the font weight of the headings in Chrome. 

Font weight is being properly applied in Safari and Firefox.

Heading Weights - Chrome

Heading Weights - Safari - Firefox

### LINKED ISSUE #1509: Use woff2 font instead of woff
Update core
Update docs' cdn ref

sweeting:
- Switch from woff to woff2 closes #1509 
- Fix font weight issues in Chrome closes #1502
- Add new font weight variables to core
- Update typography docs


packages/normalize/css/index.css
```diff
@@ -4,18 +4,20 @@
 
 @font-face {
   font-family: 'PS TT Commons Roman';
-  src: url('https://s2.pluralsight.com/fonts/ps-tt-commons/v1/ps-tt-commons-variable-roman.woff')
+  src: url('https://s2.pluralsight.com/fonts/ps-tt-commons/v1/ps-tt-commons-variable-roman.woff2')
     format('woff-variations');
   font-display: swap;
+  font-weight: 100 900;
```
jaketrent, line 10: Teach me about this key addition.  Why was it key to making this work?

jaketrent (approved):
Looks good to me.  

Piecing together what I see here and your comment in slack about "it's working!", I assume that we can still use the variable font.  That's good news.

Looks like rgba and weight are good to go in chrome, ff, safari:
Screen Shot 2020-12-16 at 6 34 34 AM

packages/normalize/css/index.css
sweeting: @jaketrent This section from [an article on css-tricks.com](https://css-tricks.com/using-css-custom-properties-to-adjust-variable-font-weights-in-dark-mode/):

> Keep in mind that, if you’re using the `@font-face` rule to load fonts, you should set a `font-weight` range for each of them at the same time:

```
@font-face {
  src: url('Highgate.woff2') format('woff2-variations');
  font-family: 'Highgate';
  font-weight: 100 900;
}
```

> If you neglect this step, some variable fonts may not properly reflect specific font-weight values in current Chromium browsers.

This fixed the problem with using `font-weight` in Chrome, but it also worked around the rgba opacity bug in Safari, which I suspect is related to the use of `font-variation-settings: "wght"`.

PR #1522, opened 2020-12-14
expose dropdown's uniqueId function for test

jaketrent:
Search team is trying to mock uniqueId's for snapshot stability, similar to what we have done.  The problem is that jest.mock requires a package name to mock a dependency. By the time we build, this package name information is not available for consumers to reach down into our code and identify the uniqueId function. This exposes uniqueId on Dropdown directly.

Could have done this through a static. The override method then would have been similar, using jest.mock.  Chose instead to simply use a prop. This is simpler and doesn't require mocking.

### DIFF
--- a/docs/src/content/components/dropdown.mdx
+++ b/docs/src/content/components/dropdown.mdx
@@ -411,6 +411,11 @@ export default Example
     type="string"
     desc="supporting text or error messaging"
   />
+  <TypesProp
+    name="uniqueId"
+    type="(prefix: string) => string"
+    desc="override internal id generation for test stability"
+  />
 </TypesTable>
 
 ### Dropdown.Item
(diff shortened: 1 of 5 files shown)


danethurber:
just a gut check. i'm sure you've tried this already....but could you do something like we've done here

https://github.com/pluralsight/design-system/blob/6bbdfe4a4ab63929c85ab3be5a77455bac69d0b0/packages/checkbox/src/react/__specs__/storyshots.spec.ts

jaketrent:
@danethurber yes, Dropdown had this, as it used util (you can see it in the diff).

And the Search team tried to use this exactly in their code too (plus a couple variations).  The problem seems to be that by the time tsc is done with the build, the `@pluralsight/ps-design-system-util` package name is not included and jest can't find it to do the mocking.

danethurber (approved):

PR #1523, opened 2020-12-14
[experimental] <Table /> sticky header

danethurber:
Docs are not written yet but i will before merging

@sweeting do you want any dropshadows on the fixed/sticky headers?


packages/table/src/react/sentinel.tsx
```diff
@@ -0,0 +1,85 @@
...
+      observer.disconnect()
+    }
+  }, [start, end, options, onEnter, onLeave])
+}
+
+interface SentinelProps extends Omit<HTMLPropsFor<'div'>, 'ref'> {
+  end?: boolean
+}
+
+export const Sentinel = React.forwardRef<HTMLDivElement, SentinelProps>(
```
jaketrent, line 76: Have a good url for the "sentinels" concept you're using here so that I can learn it?

packages/table/src/react/context.ts
```diff
@@ -0,0 +1,13 @@
...
+  stuck: boolean
+}
+
+export const initialContext: Required<ContextValue> = {
+  stuck: true
+}
+
+const Context = React.createContext<ContextValue>(initialContext)
+
+export default Context
```
danethurber, line 13: i need to remove this. i don't use the context anymore

jaketrent (approved):

packages/table/src/react/sentinel.tsx
danethurber: yep. here you go https://dev.to/jennieji/so-hard-to-make-table-header-sticky-14fj

PR #1524, opened 2020-12-15
feat(date): needs animation,  text input examples, hook tests

### LINKED ISSUE #1498: DateRangePicker: new component
Use 3rd party foundations:
- Can use react-dates from airbnb, if possible
- Could explore [dazed](https://github.com/deseretdigital/dayzed) headless datepicker lib

Handle Flow needs with a pattern:
- See figma for presents for ranges (eg, for Sprints)

Viz Design:
https://www.figma.com/file/YsTklBecdddy9RZ3HXddIseD/PSDS-Web-Components?node-id=5373%3A3816

Features:
- Blackout/unselectable days - disabled days or valid ranges (eg, not before this day)

Related:
- Probably replace current DatePicker too -- would be nice to have UX of single-field input

EdwardIrby:
Ready for review
Resolves #1498


packages/datepicker/src/css/index.ts
```diff
@@ -1,229 +1,179 @@
...
-}
-
 export default {
-  '.psds-date-picker': {
-    display: 'inline-block'
-  },
-  '.psds-date-picker--disabled': {
-    opacity: 0.5
-  },
+  '@keyframes psds-calendar__keyframes__forward': {
```
jaketrent, line 10: Seems worth calling out that the `keyframes__` prefix is a new pattern. I like it because it groups keyframe declarations separate from the other selectors, which we haven't necessarily done in other components. Is this something we should adopt broadly?

packages/datepicker/src/index.ts
```diff
@@ -1,4 +1,2 @@
-export { default } from './react'
-export * from './vars'
-export * from './js'
-export { default as css } from './css'
```
jaketrent, line 4: I don't think we want to delete the css export. The file exists, was used and this pattern is used broadly. Any reason to delete?

packages/datepicker/src/react/calendar-dates.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+interface ChildrenRenderProps extends HTMLPropsFor<'button'> {}
+
+interface CalendarDatesProps
+  extends HTMLPropsFor<'button'>,
+    Pick<RenderProps, 'getDateProps'> {
+  children: (props: ChildrenRenderProps, dateObj: DateObj) => React.ReactNode
+}
+
+export const CalendarDates: React.FC<CalendarDatesProps> = ({
```
jaketrent, line 26: Shouldn't we forwardRef? We have tried to make that a default behavior to be relied upon. If not, why not?

packages/datepicker/src/react/calendar.tsx
```diff
@@ -1,162 +1,186 @@
...
+  headerMonth: () => css(stylesheet['.psds-calendar__header-month']),
+  weekdayHeader: () => css(stylesheet['.psds-calendar__weekday-header']),
+  dateFiller: () => css(stylesheet['.psds-calendar__filler']),
+  dateGrid: () => css(stylesheet['.psds-calendar__date-grid'])
 }
 
-interface CalendarProps {
-  value: Date | undefined
-  onSelect: (evt: React.MouseEvent, date: Date) => void
+const monthNamesShort = [
```
jaketrent, line 53: Month names are available through js i18n. Looks to be IE11 compatible. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleString  Maybe week names too?

packages/datepicker/src/react/utils.ts
```diff
@@ -0,0 +1,244 @@
...
+              ? 'forward'
+              : differenceInCalendarMonths(nextSelected, selected) < 0
+              ? 'backward'
+              : undefined
+        }
+        setSelected(nextSelected)
+        setSlide(nextSlide)
+      }
+    } catch (error) {
+      ;() => {}
```
jaketrent, line 186: This is odd.  Why does this exist?

jaketrent (changes requested):
A few questions inline and a few things that seem like things to address:

- I'd squash the "--wip-- [skip ci]" commits so we don't keep those for the git history.
- Calendar selection will select a date in stories, but input typing will not select a date. I believe it should since that was the old ux, but perhaps I'm missing something.



- Input editing for a date already selected does not seem to work properly. Backspace and typing corrupts the field value. The moment of date selection is off or never happens afaik.



Overall, a great bunch of work. Thanks for working to make this nice. I think it just needs a few use cases polished out.

I also noticed your comments in the PR about what was yet needed.  Are you planning to do that in a future PR after a merge?  Or are you going to refactor and/or add what's missing in this same PR and then want another round of reviews?

packages/datepicker/src/css/index.ts
EdwardIrby: I think so 

> A few questions inline and a few things that seem like things to address:
> 
> * I'd squash the "--wip-- [skip ci]" commits so we don't keep those for the git history.
> * Calendar selection will select a date in stories, but input typing will not select a date. I believe it should since that was the old ux, but perhaps I'm missing something.
> 
> 
> 
> * Input editing for a date already selected does not seem to work properly. Backspace and typing corrupts the field value. The moment of date selection is off or never happens afaik.
> 
> 
> 
> Overall, a great bunch of work. Thanks for working to make this nice. I think it just needs a few use cases polished out.
> 
> I also noticed your comments in the PR about what was yet needed. Are you planning to do that in a future PR after a merge? Or are you going to refactor and/or add what's missing in this same PR and then want another round of reviews?

Though I fixed the input errors. Hmm let me go look again. As for the key frames pattern yep I think we should adopt it. I got a bunch of doc typos and suggestions from Dane to implement.

packages/datepicker/src/react/calendar-dates.tsx
EdwardIrby: Nope it's a headless component that uses the render props pattern.

packages/datepicker/src/react/utils.ts
EdwardIrby: Yeah this is no longer needed will remove

packages/datepicker/src/react/calendar.tsx
EdwardIrby: This won't work but not a bad idea to make the calendar support multiple locales dynamically

jaketrent (commented):
Hey Edward, I've made 2 new commits to this branch, 87090a1 and 52b1870, that make adjustments to the docs. I made several changes in the pursuit of greater readability, highlights including:

- Structure headings with new Functions section and Migration
- Moved migration to bottom
- Renamed some examples
- Defined things like `slides`. These days with TS, I'd probably implement it as I documented it (they're not strictly the same between doc and src).
- Removed the "We do this, next we do this" verbiage to get a terser style. Statements of facts, and command form for instructions.
- Added type tables to better show inputs required. Would have liked to do this for outputs too, but the current design doesn't quite fit. Improvements here still possible.

Please feel free to further adjust my edits or write them out entirely.

Two things I'd still love to see:
- All code examples working (as noted in another comment)
- Visual states like labels, appearance, disabled and error, as seen in the current docs, return to this edition

Biggest issue I sense but am uncertain of:
I have a sense that this DatePicker is much more capable of supporting new or yet-specified use cases. This is a good thing. At the same time, it seems overly complex for the simple case. I believe there's still value for the simple case. I would love to CPT two cases: 1, simple; 2, complex. It would be helpful to know: 1, how well devs can learn and understand the api; 2, how reasonable the usage setup cost for their use case seems for them.  I'll leave that up to you on whether or not you want to investigate this further pre-release or not.

docs/src/content/components/datepicker.mdx
```diff
@@ -3,130 +3,839 @@ name: Datepicker
 route: /components/datepicker
```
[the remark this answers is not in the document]
EdwardIrby: rest of config I did to focus on on api setup but if you weren't so busy with meeting today I'd rather pair with you on this personally. Maybe tomorrow? Also I'll look into making simple easier as well I have some thoughts on that.

EdwardIrby:
@jaketrent ready for re-review. However I want to finish looking at some of the advanced patterns from EpicReact to see if I can figure out a way of dropping the render props and simplifying pattern bases usage a bit more.

packages/datepicker/src/react/dropdown-datepicker.tsx
```diff
@@ -0,0 +1,70 @@
...
+import { CalendarDates } from './calendar-dates'
+import TextInput from '@pluralsight/ps-design-system-textinput'
+import { CalendarIcon } from '@pluralsight/ps-design-system-icon'
+
+interface DropdownDatePickerProps
+  extends Omit<HTMLPropsFor<'div'>, 'onSelect'> {
+  onSelect?: (evt: React.SyntheticEvent, dateObj: DateObj) => void
+}
+
+export const DropdownDatePicker: React.FC<DropdownDatePickerProps> = ({
```
jaketrent, line 16: DropdownDatePicker was a weird name for me.  Seems unfamiliar for the base case.  Maybe just DatePicker? I don't think DatePicker would collide with other named exports.

docs/src/content/components/datepicker.mdx
```diff
@@ -748,7 +748,21 @@ export default Example
 
 ## Migration guide from v11.x and before to v12.x
 
-This guide is to migrate the old DatePicker to a single datepicker using the new components and functions.
+### Simple
```
jaketrent, line 751: I would move this to the top of the docs. It seems to be more of the simple common case and less of a migration issue.  Migration hand-holding definitely more useful in the custom date picker case.

jaketrent (approved):

packages/datepicker/src/react/dropdown-datepicker.tsx
EdwardIrby: Agree but I like to keep it named versus default.

jaketrent: ^ agreed.

PR #1528, opened 2020-12-15
feat(table): drawers

### LINKED ISSUE #1527: Table: row/drawer
animate table drawers. designs are in figma

ref: https://pluralsight.slack.com/archives/C70HPSJPP/p1607978164304900

danethurber:
resolve #1527


packages/table/src/css/index.ts
```diff
@@ -15,7 +16,9 @@ const cellPadding = `${layout.spacingXSmall} ${layout.spacingSmall}`
 const dark = {
   className: `.psds-theme--${themeNames.dark}`,
 
-  borderColor: colorsBorder.lowOnDark,
+  // TODO: should be using colorsBorder.lowOnDark but border collapse used with
```
jaketrent, line 19: What's the TODO here?  Reads to me as more of a long-term note because it's something weird we apparently have to do.  When/how do you plan to come back?

packages/table/src/react/index.tsx
```diff
@@ -241,11 +246,51 @@ const TableRow = forwardRef<HTMLTableRowElement, TableRowProps>(
 )
 TableRow.displayName = 'Table.Row'
 
+interface TableDrawerProps extends Omit<HTMLPropsFor<'tr'>, 'ref'> {
+  expanded: boolean
+  colSpan: number
+  indentWithCell?: boolean
```
jaketrent, line 252: Is this what supports the drawer caret alignment (right/left)?

jaketrent (approved):
Looks good to me.  Looks really good.  I agree with Kaiden that the drawer bg from roleiq seems to be a nice visual tie-in from the drawer to the row.  Not sure Brian's opinion, but off-hand, I'd vote for that as the default style.  Not sure what all this talk about no animation has been, but the animation that is there in storybook looks nice.  What changed that specific conversation?

packages/table/src/css/index.ts
danethurber: yes it should be a NOTE. i'll update

packages/table/src/react/index.tsx
danethurber: yes. if the indentation is on the left there is an option to insert an empty cell to help with the alignment

danethurber:
@jaketrent brian and i have been working through the animation and bg to make sure we get it right. we have been looking at other examples in the wild and there is mixed opinion on height animation as a visual indication and the perf implications :) i'm happy with were we ended up. as for the bg color brian is still noodling - there is some issue with multiple nested tables/drawers that make the bg less than ideal but i'm sure we'll come up with a workable solution

PR #1529, opened 2020-12-16
(docs) Update logo

sweeting:
Updated logo to one that uses PS TT Commons instead of Gotham.


jaketrent (approved):
Didn't realize the logo was changing too.  Nice!

PR #1537, opened 2021-01-04
build(errors): include type declarations

### LINKED ISSUE #1533: Errors: ensure typings are exported correctly
Someone tried to use the package, but the typings weren't there.

danethurber:
the Errors components still has a dependency on a babel build. in order to also get type
declarations, i've added an additional build step that uses tsx to compile the typescript
declarations to place them in the dists folder so they can be referenced.

fixes #1533


jaketrent (approved):
Works, great solution!

PR #1538, opened 2021-01-04
fix(navitem): pass unspecified props to button

jaketrent:
yeeehaa, pass onClicks

### DIFF
--- a/packages/navitem/src/react/context.tsx
+++ b/packages/navitem/src/react/context.tsx
@@ -18,6 +18,7 @@ export interface ContextValue {
   selected: boolean
   // eslint-disable-next-line camelcase
   UNSAFE_stylesFor?: Partial<Record<AllowedSelectors, CSSProperties>>
+  rest?: Record<string, unknown>
 }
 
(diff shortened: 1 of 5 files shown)


packages/navitem/src/react/context.tsx
```diff
@@ -18,6 +18,7 @@ export interface ContextValue {
   selected: boolean
   // eslint-disable-next-line camelcase
   UNSAFE_stylesFor?: Partial<Record<AllowedSelectors, CSSProperties>>
+  rest?: Record<string, unknown>
```
danethurber, line 21: why a named prop for rest?

danethurber (approved):
works for me. left one comment

packages/navitem/src/react/context.tsx
jaketrent: It's not a prop.  It's on context.  Context is the way that we're pass stuff to inner elements now.

danethurber: ah...got it 👍

PR #1543, opened 2021-01-05
refactor(core, normalize): add variables for new type style families and remove ligatures in normalize

### LINKED ISSUE #1547: core: add new vars for the new type style families
The general font size, letterspacing, and line-height vars are no longer accurate and new vars should be created for the new type style families for headers, labels, paragraph.

sweeting:
Closes #1547

core
- Add variables for new type style families (heading, label, paragraph)
 
normalize
- Disable ligatures

text
- use new core type vars

### DIFF
--- a/packages/core/css/type.module.css
+++ b/packages/core/css/type.module.css
@@ -13,6 +13,69 @@
   --psTypeFontSizeSmall: 14px;
   --psTypeFontSizeXSmall: 12px;
 
+  /* heading */
+  --psTypeHeadingXLargeFontSize: 40px;
+  --psTypeHeadingXLargeLineHeight: 52px;
+  --psTypeHeadingXLargeLetterSpacing: -0.01em;
+  --psTypeHeadingXLargeFontWeight: 600;
+
+  --psTypeHeadingLargeFontSize: 34px;
+  --psTypeHeadingLargeLineHeight: 40px;
+  --psTypeHeadingLargeLetterSpacing: -0.01em;
+  --psTypeHeadingLargeFontWeight: 600;
+
+  --psTypeHeadingMediumFontSize: 24px;
+  --psTypeHeadingMediumLineHeight: 32px;
+  --psTypeHeadingMediumLetterSpacing: -0.01em;
+  --psTypeHeadingMediumFontWeight: 600;
+
+  --psTypeHeadingSmallFontSize: 20px;
+  --psTypeHeadingSmallLineHeight: 28px;
+  --psTypeHeadingSmallLetterSpacing: -0.01em;
+  --psTypeHeadingSmallFontWeight: 600;
+
+  --psTypeHeadingXSmallFontSize: 14px;
+  --psTypeHeadingXSmallLineHeight: 20px;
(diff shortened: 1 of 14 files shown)


packages/core/css/type.module.css
```diff
@@ -13,6 +13,74 @@
   --psTypeFontSizeSmall: 14px;
   --psTypeFontSizeXSmall: 12px;
```
jaketrent, line 14: Do you think the old font sizes, weights and line heights should be removed?  If not, why keep?

packages/core/js/type.ts
```diff
@@ -1,7 +1,7 @@
 export default {
   fontFamily:
     '"PS TT Commons Roman", "Gotham SSm A", "Gotham SSm B", Arial, sans-serif',
-  fontFamilyCode: '"Source Code Pro", monospace',
+  fontFamilyCode: '"DM Mono", monospace',
```
jaketrent, line 4: This probably needs updated in the docs too.

packages/drawer/src/react/__specs__/__snapshots__/storyshots.spec.tsx.snap
```diff
@@ -25,7 +25,7 @@ exports[`Storyshots drawer controlled: external state 1`] = `
     role="button"
   >
     <p
-      data-css-r3ja7m=""
+      data-css-1joltl8=""
```
jaketrent, line 28: Can you speak to the reason that the snapshots in this pr have changed?  As I understand it now, they shouldn't have changed because no styles/values have changed, only the variables used to represent those values.

packages/text/src/css/index.ts
```diff
@@ -52,43 +52,48 @@ export default {
     color: colorsTextIcon.lowOnDark
   },
   [`.psds-text__heading--size-${vars.headingSizes.xXXSmall}`]: {
-    fontSize: '12px',
-    lineHeight: '16px',
-    textTransform: 'uppercase',
-    letterSpacing: '.08em'
+    fontSize: type.headingXXXSmallFontSize,
```
jaketrent, line 55: 👍

packages/normalize/css/index.css
```diff
@@ -17,7 +17,7 @@ html {
   font-size: 0.875em;
   font-weight: var(--psTypeFontWeightRegular);
   line-height: 1.71429; /* calc(var(--psTypeLineHeightStandard) / var(--psTypeFontSizeSmall)); */
-  font-feature-settings: 'tnum' on, 'lnum' on;
+  font-feature-settings: 'tnum' on, 'lnum' on, 'liga' off;
```
jaketrent, line 20: This seems to be the only thing I can find that's a new style.  Correct?

Even if that's the case, I don't see this influencing the data-css-hashes, because normalize is a peerDependency that's just expected to be on the page, added to the style cascade in the browser, not present at build time.

jaketrent (approved):

jaketrent:
Team triage: pause for now. Rethink general vars in core. Make sure we can support teams in their usage patterns for the Text components.

PR #1554, opened 2021-01-13
fix card actionBarVisible

jaketrent:
Big changes here


packages/card/src/react/index.tsx
```diff
@@ -34,7 +34,7 @@ const styles = {
       fullOverlay &&
         !visible &&
         css(stylesheet[`.${label}--fullOverlay.${label}--no-actionBarVisible`]),
-      visible && css(stylesheet[`${label}--actionBarVisible`])
+      visible && css(stylesheet[`.${label}--actionBarVisible`])
```
danethurber, line 37: so sneaky

danethurber (approved):
👍

Issue #1544, opened 2021-01-06
datepicker: issue with date selection


danethurber:
> When open if don't select any date of current month but navigate to other months on the right or left, it will stuck in an infinitely loop.

https://pluralsight.slack.com/archives/C70HPSJPP/p1609948607354700

PR #1555, opened 2021-01-13
fix datepicker loop

jaketrent:
- Remove calendar internal state
- Update input fields based on parent state change

Resolves: #1544


danethurber (approved):
makes sense to me.

the docs say value is required but also can be undefined - does that make sense to you?

Screen Shot 2021-01-14 at 10 03 39 AM

jaketrent:
@danethurber yeah, it was meant to show 2 things:

1. It is a controlled component
2. A seed date is not required

This could have been done another way, but I hoped this would be the best way to tersely express the combo here.  Alternatives you'd suggest?

Issue #1546, opened 2021-01-06
Table: solution for different bg colors


danethurber:
possible solution: 

>  add a prop to allow style overrides. at the moment we need the background color to display the overflow gradients but i can write some documentation on how to replicate that with different colors.

https://pluralsight.slack.com/archives/C70HPSJPP/p1609950482356200

danethurber:
new solution. let's remove the bg color all together including the gradients/shadows use to denote overflow

> The overflow shadows are mostly only discernible on the light theme and the row hover kind of breaks the effect of them, so I think we could remove them.
^ @sweeting

PR #1558, opened 2021-01-13
remove table background

jaketrent:
I'm not 100% I fully understand how wide of a background removal we're asking for here, so I took my best guess.  Please verify.  I left a number of background styles because I think we still want them (eg, sticky headers and hover states).

Resolves: #1546


danethurber (approved):
👍  LGTM

Issue #1545, opened 2021-01-06
docs: Table codesandbox creation is missing a dependency


danethurber:
> Could not find dependency: 'invariant' relative to '/node_modules/@pluralsight/ps-design-system-table/dist/esm/react/index.js'

danethurber:
i think adding `invariant` to the codesandbox deps will solve the issue

https://github.com/pluralsight/design-system/blob/master/docs/src/components/code-block/actions.tsx#L78

jaketrent:
deployed, resolved.

PR #1559, opened 2021-01-14
fix sandbox dependencies

jaketrent:
- I'm assuming that `invariant` must come from the `devDependencies` present. It might not be. We could just list it as a new dependencies, as that seems to solve the problem manually within codesandbox.  Waddya think?
- I think the problem is that devDependencies don't show up in the sandbox. https://github.com/codesandbox/codesandbox-client/issues/2794

### DIFF
--- a/docs/src/components/code-block/actions.tsx
+++ b/docs/src/components/code-block/actions.tsx
@@ -81,7 +81,8 @@ export const CodeSandboxAction: React.FC<CodeSandboxActionProps> = props => {
         '@pluralsight/ps-design-system-normalize': 'latest',
         '@pluralsight/ps-design-system-theme': 'latest',
         '@babel/runtime': 'latest',
-        glamor: 'latest'
+        glamor: 'latest',
+        invariant: 'latest'
       }}
       gitInfo={gitInfo}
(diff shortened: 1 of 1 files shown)


danethurber:
@jaketrent did you try the suggestion i put in the issue https://github.com/pluralsight/design-system/issues/1545#issuecomment-760350441

jaketrent:
@danethurber Sorry, I didn't see that suggestion.  No, I did not try it.  I did it manually in a codesandbox at one point, and it worked.  I didn't want to add it, because it didn't make sense. We're not importing it anywhere.  It's not a named dependency in our projects.  Something is using it.  And my guess was that that something was the thing that should be added to the codesandbox dependencies instead. I'm guess it's one of the things that we have in devDeps, since everything else mentioned in the src is in deps.  Based on the experiences in this thread: https://github.com/codesandbox/codesandbox-client/issues/2794#issuecomment-542195016

jaketrent:
ah, so we do.  Ok, I'll look at that route.  Thanks!

jaketrent:
@danethurber k, changed and ready for another review.

danethurber (approved):

PR #1573, opened 2021-01-22
test(dropdown): add disabled menu case

jaketrent:
Seems to work. What am I missing?


jaketrent:
@caseybaggz, would you mind weighing in and helping us make this better so that the Dropdown is fully accessible via testing-library, please?

The disabled test here passes. Perhaps there's a part of the "disabled" use case that we're not covering here. Would you mind clarifying what the testing problem was that you had, please?

caseybaggz (approved):
This looks good!

PR #1576, opened 2021-01-25
feat(storybook-preset): add a11y addon

danethurber:


packages/storybook-preset/src/index.ts
```diff
@@ -16,12 +17,27 @@ export function config(entry = [], options: Options = {}) {
...
+      if (!addonSupportsPreview) {
+        try {
+          const preset = join(pkgName, 'preset')
+          const { config: presetConfig } = require(require.resolve(preset))
+          acc = acc.concat(presetConfig())
+
+          // eslint-disable-next-line
+        } catch (err) {}
+      }
+
```
jaketrent, line 40: This is an odd block of code to me.  I think what you're doing is accessing the config for a preset in potentially 2 ways: First through the package default export, then falling back to a named config function export.  

If this is accurate, I would suggest extracting a function and calling it twice with a parameter specifying the different access strategy. I think it would read a lot easier.

jaketrent (approved):

packages/storybook-preset/src/index.ts
danethurber: yes. the underlying storybook api is changing. previously there was a `preview.js` file that addons used for configuration. the storybookk-a11y doesn't have a `preview.js` file and instead exports a config fn in it's `preset` file. i'll take you suggestion and create some more readable functions

EdwardIrby (approved):

PR #1579, opened 2021-01-25
fix(carousel): remove flash when used with cards

danethurber:
> info in comments

### DIFF
--- a/packages/carousel/src/react/index.tsx
+++ b/packages/carousel/src/react/index.tsx
@@ -221,25 +221,41 @@ interface PageProps extends HTMLPropsFor<'ul'> {
   paged?: boolean
   isActivePage?: boolean
 }
-const Page: React.FC<PageProps> = props => {
-  const ref = React.useRef<HTMLUListElement>(null)
 
+const Page: React.FC<PageProps> = props => {
   const { children, isActivePage, paged, ...rest } = props
   const { offset, transitioning, setTransitioning } = React.useContext(
     CarouselContext
   )
-  React.useEffect(() => {
-    if (paged && isActivePage) {
-      const page = ref.current
-      if (page) {
-        const focusFirstChild = (evt: TransitionEvent) => {
-          setTransitioning(false)
-          evt.target === page && (page.firstElementChild as HTMLElement).focus()
+
+  const ref = React.useRef<HTMLUListElement>(null)
+
+  React.useEffect(
+    function responseToTransitionEnd() {
+      const { current: page } = ref
+      if (!page || !paged || !isActivePage) return
+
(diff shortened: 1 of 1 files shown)


packages/carousel/src/react/index.tsx
```diff
@@ -221,25 +221,41 @@ interface PageProps extends HTMLPropsFor<'ul'> {
...
+  const ref = React.useRef<HTMLUListElement>(null)
+
+  React.useEffect(
+    function responseToTransitionEnd() {
+      const { current: page } = ref
+      if (!page || !paged || !isActivePage) return
+
+      const onTransitionEnd = (evt: TransitionEvent) => {
+        const isPageTransition =
+          evt.target === page && evt.propertyName === 'transform'
```
danethurber, line 240: The issue was caused because we were responding to all self and child transitions. The bug in the example was caused because we were updating state when a transition in the Card(a opacity transition in Card.Actions) was setting the new transition state prematurely and hiding the previous page.

The fix is to only update the transition state if the `transitionend` matches the current page as the event target and only on `transform` transitions

packages/carousel/src/react/index.tsx
```diff
@@ -221,25 +221,41 @@ interface PageProps extends HTMLPropsFor<'ul'> {
...
+
+        if (!isPageTransition) return
+        setTransitioning(false)
+
+        if (page && page.firstElementChild instanceof HTMLElement) {
```
danethurber, line 245: instead of casing the element type i've used a typeguard - this is the recommended approach

packages/carousel/src/react/index.tsx
```diff
@@ -221,25 +221,41 @@ interface PageProps extends HTMLPropsFor<'ul'> {
...
-        page.addEventListener('transitionend', focusFirstChild)
       }
-    }
-  }, [isActivePage, paged])
+
+      page.addEventListener('transitionend', onTransitionEnd)
+
+      return () => {
+        page.removeEventListener('transitionend', onTransitionEnd)
+      }
```
danethurber, line 254: added this line to clean up those event 🧟  🧟‍♀️  🧟‍♂️

EdwardIrby (changes requested):
Docs animation not working locally for me could be a something I need to do on my side

danethurber:
@EdwardIrby I see the same issue with the development build. if i do a full production build it works fine 🤷  - no idea why

EdwardIrby (approved):
Let's role with it for now then and see if it get's kicked back to us

packages/carousel/src/react/index.tsx
jaketrent: wow, nice find!

PR #1581, opened 2021-01-26
Field

danethurber:
> Customizable building blocks for form fields that can be used to build/replace the current form components.

**please review**: i haven't written docs yet - i want to get feedback on the approach before the docs take shape

--- 

storybook examples of building the following components

- [x] TextField
- [x] TextAreaField
- [x] SelectField using downshiftjs
- [x] TagsField using downshiftjs
- [x] MultiSelectField using downshiftjs


packages/field/src/css/field.ts
```diff
@@ -0,0 +1,73 @@
...
+  '.psds-field__container--error': {},
+
+  //   '.psds-text-input__error': {
+  //     position: 'absolute',
+  //     right: `calc(-1 * (${iconWidths.medium} + ${layout.spacingXSmall}))`,
+  //     display: 'flex',
+  //     alignItems: 'center',
+  //     color: colorsStatus.error,
+  //     marginLeft: layout.spacingXSmall
+  //   }
```
danethurber, line 30: uh.....i forgot that i haven't added support for the error icon yet. will update

packages/field/src/react/__specs__/index.spec.tsx
```diff
@@ -0,0 +1,20 @@
+import { render } from '@testing-library/react'
+import React from 'react'
+
+import Field from '..'
+
+describe('Field', () => {
```
danethurber, line 6: these tests are lame - i will provide some better ones 🤞

packages/field/src/react/field.tsx
```diff
@@ -0,0 +1,160 @@
...
+      css(stylesheet['.psds-field']),
+      css(stylesheet[`.psds-field--${opts.size}`])
+    ),
+  prefix: () => css(stylesheet['.psds-field__prefix']),
+  suffix: () => css(stylesheet['.psds-field__suffix'])
+}
+
+type InputElements = HTMLInputElement | HTMLSelectElement | HTMLTextAreaElement
+
+interface FieldProps extends Omit<HTMLPropsFor<'div'>, 'prefix' | 'ref'> {
```
danethurber, line 47: needed to omit the legacy string ref from `HTMLPropsFor`. i don't think we need to support the legacy ref anymore so I propose we just always do this in the utils package

packages/field/src/react/__stories__/tags-input-field.story.tsx
```diff
@@ -0,0 +1,250 @@
...
+} as Meta
+
+const Template: Story<ComponentProps<typeof TagsField>> = args => {
+  return <TagsField {...args} />
+}
+
+export const Basic = Template.bind({})
+
+export const WithSuffix = Template.bind({})
+WithSuffix.args = {
```
jaketrent, line 233: Why is this done?  Is the `.args` thing storybook-specific?

packages/field/src/react/__stories__/select-field.story.tsx
```diff
@@ -0,0 +1,115 @@
+import { CaretDownIcon } from '@pluralsight/ps-design-system-icon'
```
jaketrent, line 1: Would be cool to see a Dropdown replacement using Dropdown.Menu or ActionMenu. Nbd, though.  Would just show that we've already replaced it.

packages/field/src/react/__stories__/tags-input-field.story.tsx
```diff
@@ -0,0 +1,250 @@
+import { Meta, Story } from '@storybook/react/types-6-0'
```
jaketrent, line 1: What's the diff with tags-input-field and multi-select-field?  They seem essentially the same with just a bit of design diff.

jaketrent (approved):
Works for me. I dig the approach. Seems to work really well. My only worry is that this becomes a very different dev process for new apis on several components.

packages/field/src/react/field.tsx
danethurber: thanks for the 👍  - i'll do that in a new pr

packages/field/src/react/__stories__/multi-select-field.story.tsx
```diff
@@ -0,0 +1,293 @@
+import { Meta, Story } from '@storybook/react/types-6-0'
```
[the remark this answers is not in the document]
danethurber: yeah...i think it's a test dom support issue but will investigate

packages/field/src/react/__stories__/multi-select-field.story.tsx
```diff
@@ -0,0 +1,293 @@
...
+
+interface MultiSelectFieldProps extends ComponentProps<typeof Field> {
+  options: Option[]
+  placeholder?: string
+}
+const MultiSelectField: React.FC<MultiSelectFieldProps> = props => {
+  const { options, disabled, placeholder, ...rest } = props
+
+  const [filterTerm, setFilterTerm] = useState<string>('')
+  const handleFilterTermChange: ChangeEventHandler<HTMLInputElement> = useCallback(
```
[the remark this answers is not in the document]
danethurber: good catch. i need to do some investigation into what should be optimized. at least for these stories they are just to proved the approach and if/when we used a similar approach for an official component we can invest the research time them. i think it's best to remove the callbacks since i have no indication they help. i'll do that now

packages/field/src/react/__stories__/tags-input-field.story.tsx
danethurber: yes it's a storybook think. you define default props and bind a Template. then you can pass in new args for each additional story you create

packages/field/src/react/__stories__/select-field.story.tsx
danethurber: agreed. this is just an example of getting the functionality and a11y right. i assume we'll followup with an official `DropdownField` or `SelectField` component that would use an `ActionMenu`

packages/field/src/react/__stories__/tags-input-field.story.tsx
danethurber: `TagsInput` does not use the combobox pattern where `MultiSelectField` does. TagsInput is do prove that we can support @KaidenR custom needs where a traditional combobox(`ActionMenu`) is not used.

packages/field/src/react/input.tsx
```diff
@@ -0,0 +1,34 @@
+import { HTMLPropsFor } from '@pluralsight/ps-design-system-util'
+import { css } from 'glamor'
+import React, { forwardRef } from 'react'
+
+import stylesheet from '../css/input'
+
+const styles = {
+  container: () => css(stylesheet['.psds-input__container']),
```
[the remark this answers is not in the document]
danethurber: agreed. will update

PR #1588, opened 2021-02-01
fix italics for chrome

### LINKED ISSUE #1578: font: verify em tags using PS TT Commons will be italicized
Src: https://pluralsight.slack.com/archives/C70HPSJPP/p1611187347010100

New font doesn't italicize em's. Works in FF, not Chrome.

jaketrent:
Resolves #1578

### DIFF
--- a/packages/normalize/css/index.css
+++ b/packages/normalize/css/index.css
@@ -9,6 +9,13 @@
   font-display: swap;
   font-weight: 100 900;
 }
+@font-face {
+  font-family: 'PS TT Commons Roman';
+  src: url('https://fonts.pluralsight.com/ps-tt-commons/v1/ps-tt-commons-variable-italic.woff2')
+    format('woff-variations');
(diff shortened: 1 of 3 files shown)


packages/normalize/css/index.css
```diff
@@ -9,6 +9,13 @@
   font-display: swap;
   font-weight: 100 900;
```
jaketrent, line 10: Would be great if there was a "font-style range" option like there is for font-weight.

packages/normalize/css/index.css
```diff
@@ -74,6 +81,10 @@ li {
   }
 }
 
+em {
```
jaketrent, line 84: This was one of my more lighter-touch solution attempts.  It doesn't work by itself. But I left it because it makes sense for a normalize. But, on second thought, this should be browser default anyway, right?

danethurber (approved):

PR #1590, opened 2021-02-01
fix(dialog) increase close button color contrast for a11y

jaketrent:
Matched flat Button style. But didn't make Button a dependency. Was going for a smaller package size. Right choice?


danethurber (approved):
@jaketrent your call. i think i would still use a Button 🤷‍♂️

PR #1591, opened 2021-02-01
[experiment] raw selector style override concept

jaketrent:
Not for merging. Proof of concept. 

If we exposed selectors that weren't hashed, they'd be stable and allow users to override all parts of a component, sub elements included.  See snapshot changes for className values.

Hooked this into the existing style generation code.

Would need to provide caveats and training for how to do overrides, esp in shared page environments so that PSDS components don't get clobbered (trade our current problems for the historical problems of global namespace shootouts). See added story for potential advice for containing overrides.

What do you think? Effective? Efficient? Beautiful? Adjustments?

### DIFF
--- a/packages/button/src/react/index.tsx
+++ b/packages/button/src/react/index.tsx
@@ -52,6 +52,13 @@ const selectToString = (...infos: SelectorInfo[]) => {
   return `${hashedSelector} ${dedupedRawSelectors}`
 }
 
+const mergeProps = (first: unknown, second: unknown) => {
+  if (typeof first === 'string' || typeof second === 'string') {
+    return [first, second].filter(Boolean).join(' ')
+  }
+  // TODO: etc
+}
+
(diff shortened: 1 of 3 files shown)


packages/button/src/react/index.tsx
```diff
@@ -5,14 +5,62 @@ import {
...
+    .join(' ')
+  return `${hashedSelector} ${dedupedRawSelectors}`
+}
+
+const mergeProps = (first: unknown, second: unknown) => {
+  if (typeof first === 'string' || typeof second === 'string') {
+    return [first, second].filter(Boolean).join(' ')
+  }
+  // TODO: etc
+}
```
jaketrent, line 60: Would rather find a lib that already does this part.

jaketrent:
This is clutter in the PR list. It's an experiment. I think the idea has merit. Would lot to bring this up when the conversation is more ready. For now, archiving.

PR #1592, opened 2021-02-02
fix(actionmenu): remove useCloseOnDocumentEvents and stop using JSX p…

### LINKED ISSUE #1582: useCloseOnDocumentEvents buggy
Looks like the hook is invoking it's callback immediately now. Will debug and fix.

Slack: https://pluralsight.slack.com/archives/C70HPSJPP/p1611787899020600

EdwardIrby:
Resolves #1582

See Slack: https://pluralsight.slack.com/archives/C70HPSJPP/p1611787899020600

### DIFF
--- a/packages/util/src/use-close-on-document-events.ts
+++ b/packages/util/src/use-close-on-document-events.ts
@@ -23,14 +23,16 @@ export const onGlobalEventsClose = <El extends HTMLElement>(
     return currentAnimationFrame
   }
 
-  document.addEventListener('click', handleClickOutsideMenu)
+  document.addEventListener('click', handleClickOutsideMenu, { capture: true })
 
   window.addEventListener('resize', requestAnimationFrame, {
-    passive: true
+    passive: true,
+    capture: true
   })
 
   document.addEventListener('scroll', requestAnimationFrame, {
-    passive: true
+    passive: true,
+    capture: true
   })
 
   return () => {


packages/actionmenu/src/react/index.tsx
```diff
@@ -66,16 +66,8 @@ export const ActionMenu = React.forwardRef<HTMLUListElement, ActionMenuProps>(
       children,
       ...rest
     },
-    forwardedRef
+    ref
   ) => {
-    const ref = React.useRef<HTMLUListElement | null>(null)
```
jaketrent, line 71: By removing this, we are removing the functionality to close the actionmenu by clicking outside of it.  That doesn't seem like a UX improvement. At the least it seems like it's a breaking change.

Why remove this entirely instead of changing away from the buggy implementation?

EdwardIrby: Considered that actually but still couldn't get it to work because of Position not the hook. I could try and fix Position but that might cause unforeseen breakage. It's basically too coupled.

jaketrent: I don't understand this answer.  Perhaps I need more info in order to:

How are Position and useCloseOnDocumentEvents too coupled?

And I don't understand the comment in standup:

Why does Position or ActionMenu need broken?  And what does broken mean?

EdwardIrby:
It was react 17!!!!

https://reactjs.org/blog/2020/08/10/react-v17-rc.html#fixing-potential-issues

danethurber (approved):

jaketrent (commented):
Awesome!  I'm glad you found the issue.

PR #1593, opened 2021-02-03
Button: a11y tweaks and design true-up

sweeting:
`Button`: design tweaks and improved contrast for a11y
- Change rounding to 3px (coming change in all inputs, buttons, etc)
- Adjust text sizes to match design changes that are better suited to PS TT Commons
- Adjust padding for icon buttons at different sizes to match designs
- Focus ring meets contrast guidelines on both dark/light backgrounds
- Increase contrast of button foreground/background
- Use focus-visible to remove focus ring when clicking with mouse in both Firefox and Chrome

`Core`: color var adjustments for text & primary action color for better contrast
- Increase contrast of textIcon colors for dark backgrounds
- Darken primary action bg colors to meet contrast guidelines
- Darken error, info, success colors to meet contrast guidelines
- Additional backgroundUtility transparencies


packages/button/src/css/index.ts
```diff
@@ -13,18 +13,10 @@ import {
   names as themeNames
 } from '@pluralsight/ps-design-system-theme'
 import { transparentize } from '@pluralsight/ps-design-system-util'
+import { NoEmitOnErrorsPlugin } from 'webpack'
```
jaketrent, line 16: Was this added by mistake?

packages/button/src/css/index.ts
```diff
@@ -46,23 +38,28 @@ export default {
     alignItems: 'stretch',
     padding: `${layout.spacingSmall} ${layout.spacingMedium}`,
     border: 0,
-    borderRadius: '2px',
-    fontSize: type.fontSize200,
+    borderRadius: '3px',
```
jaketrent, line 41: Ah yeah, let's get round!

packages/button/src/css/index.ts
```diff
@@ -46,23 +38,28 @@ export default {
...
+    fontSize: type.fontSize300,
     fontWeight: type.fontWeight500,
     textAlign: 'center',
     whiteSpace: 'nowrap',
     textDecoration: 'none',
-    transition: `all ${motion.speedNormal}`,
+    transition: `all ${motion.speedXFast}`,
```
jaketrent, line 47: tangential note: been reading lately about the perf benefits of transitioning only the properties we are actually wanting changed. Then tweening of other properties is not attempted.

packages/button/src/css/index.ts
```diff
@@ -46,23 +38,28 @@ export default {
...
     '&:not([disabled]):hover': {
       cursor: 'pointer'
     },
     '&:focus': {
       outline: 'none',
-      boxShadow: `0 0 0 3px ${transparentize(
-        0.5,
-        colorsPrimaryAction.background
-      )}`
+      boxShadow: `0 0 0 2px ${colorsBlue[4]}`
```
jaketrent: stark, alpha 100, nice.

jaketrent: This makes it look surprisingly different.

packages/button/src/css/index.ts
```diff
@@ -46,23 +38,28 @@ export default {
...
+    },
+    '&:focus:not(:focus-visible)': {
+      outline: 'none',
+      boxShadow: 'none'
+    },
+    '&:focus-visible': {
+      outline: 'none',
+      boxShadow: `0 0 0 2px ${colorsBlue[4]}`
```
jaketrent: These selectors don't work in IE or Safari.

packages/button/src/css/index.ts
```diff
@@ -46,23 +38,28 @@ export default {
     alignItems: 'stretch',
     padding: `${layout.spacingSmall} ${layout.spacingMedium}`,
     border: 0,
-    borderRadius: '2px',
-    fontSize: type.fontSize200,
+    borderRadius: '3px',
+    fontSize: type.fontSize300,
```
jaketrent, line 42: Gray area of what "breaking change" means in UI code....

With the larger font size, do you see the buttons as wider and thus creating reflows when deployed to pages?

packages/button/src/css/index.ts
```diff
@@ -74,79 +71,84 @@ export default {
...
-      background: colorsBlue[8]
+      background: colorsBlue[9]
     }
   },
   [`.psds-button--appearance-${vars.appearances.primary}.psds-theme--${themeNames.light}`]: {
     color: colorsTextIcon.highOnDark,
     background: colorsPrimaryAction.background,
     '&:not([disabled]):hover': {
-      background: colorsBlue[7],
+      background: colorsBlue[8],
```
jaketrent, line 127: Thanks for tweaking all these.  I'd love to have worked with you to see your process in a11y color checking all these.

packages/button/src/css/index.ts
```diff
@@ -74,79 +71,84 @@ export default {
...
     },
     '&:not([disabled]):active': {
-      background: colorsBlue[8]
+      background: colorsBlue[9]
     }
   },
   [`.psds-button--appearance-${vars.appearances.secondary}.psds-theme--${themeDefaultName}`]: {
     color: colorsTextIcon.highOnDark,
-    background: colorsSecondaryAction.backgroundDefaultOnDark,
+    background: colorsBackgroundUtility[25],
```
jaketrent, line 144: I appreciate this instead of the extra layer of abstraction we had before.

packages/button/src/css/index.ts
```diff
@@ -74,79 +71,84 @@ export default {
...
     '&:not([disabled]):focus': {
-      background: colorsSecondaryAction.backgroundHoverOnLight
+      background: colorsBackgroundUtility[25]
     },
     '&:not([disabled]):active': {
-      background: colorsSecondaryAction.backgroundPressedOnLight
+      background: colorsBackgroundUtility[30]
     }
   },
   [`.psds-button--appearance-${vars.appearances.stroke}.psds-theme--${themeDefaultName}`]: {
```
jaketrent, line 170: Instead of updating stroke appearance, when do we get to just remove it as once planned?

packages/button/src/css/index.ts
```diff
@@ -246,10 +252,22 @@ export default {
...
+    paddingRight: layout.spacingXXSmall
+  },
+  [`.psds-button--iconAlign-${vars.iconAligns.left}.psds-button--not-iconOnly.psds-button--size-${vars.sizes.small}`]: {
+    paddingLeft: layout.spacingXXSmall
+  },
+  [`.psds-button--iconAlign-${vars.iconAligns.right}.psds-button--not-iconOnly.psds-button--size-${vars.sizes.medium}`]: {
     paddingRight: layout.spacingXSmall
   },
-  [`.psds-button--iconAlign-${vars.iconAligns.left}.psds-button--not-iconOnly`]: {
+  [`.psds-button--iconAlign-${vars.iconAligns.left}.psds-button--not-iconOnly.psds-button--size-${vars.sizes.medium}`]: {
```
jaketrent, line 286: Generally, if we change the selector in the css/index.ts, we'll have to change its usage in the react/index.ts.  

I don't see that accompanying change here, but I think the case is shown in this story:

http://localhost:6006/?path=/story/button-icon--medium

And it looks fine to my eye and has padding-left of 8px.  

Will you confirm that this ^ sounds reasonable to you and that you're accomplishing what you intend?

packages/core/css/colors.module.css
```diff
@@ -126,6 +126,9 @@
   --psColorsBackgroundLight3: #ffffff;
 
   --psColorsBackgroundUtilityBase: #b0b0b0;
+  --psColorsBackgroundUtility10: rgba(138, 153, 168, 0.1);
```
jaketrent, line 129: So many new elevations possible!

packages/core/css/colors.module.css
```diff
@@ -146,13 +149,13 @@
...
+  --psColorsPrimaryActionBackground: #0074ab;
   --psColorsPrimaryActionTextDarkTheme: #2ea0d6;
   --psColorsPrimaryActionTextLightTheme: #0074ab;
 
-  --psColorsStatusError: #d21c09;
-  --psColorsStatusInfo: #0084bd;
-  --psColorsStatusSuccess: #009e52;
+  --psColorsStatusError: #c00f00;
+  --psColorsStatusInfo: #0074ab;
+  --psColorsStatusSuccess: #007c3a;
```
jaketrent, line 158: Why not, right?  Were these also inaccessible?

packages/button/src/css/index.ts
```diff
@@ -49,20 +49,32 @@ export default {
...
+    '&:focus': {
+      outline: 'none',
+      boxShadow: `0 0 0 2px ${colorsBlue[8]}`
     },
-    '&:not([disabled]):active': {
-      transform: 'scale(0.98)'
+    '&:focus-visible': {
+      outline: 'none',
+      boxShadow: `0 0 0 2px ${colorsBlue[8]}`
     }
```
jaketrent: Again, won't work in IE or Safari.

packages/button/src/css/index.ts
```diff
@@ -117,7 +129,11 @@ export default {
     },
     '&:not([disabled]):focus': {
       background: colorsBlue[8],
-      cursor: 'pointer'
+      cursor: 'pointer',
+      boxShadow: `0 0 0 2px ${colorsBlue[5]}`
+    },
+    '&:not([disabled]):focus-visible': {
+      boxShadow: `0 0 0 2px ${colorsBlue[5]}`
     },
```
jaketrent, line 137: Ditto, IE/Safari not supported.

jaketrent (commented):
Hey Brian, great work here!  I'm glad you were able to tweak all these for accessibility.

There are a couple comments that might want your attention.  Let's at least talk about them before advancing to merge.

packages/button/src/css/index.ts
sweeting: I guess so. I'm not sure how that got added. I'll remove it.

packages/core/css/colors.module.css
sweeting: They aren't, but I'm going to remove these changes from this PR since they aren't related to Button.

packages/button/src/css/index.ts
sweeting: @jaketrent I will need some help because I'm not sure what I need to do here. Before, the spacing between the icon and the edge of the button was the same for medium, small, xsmall buttons. I adjusted it so the space between the icon and edge of the button is xxSmall (4px) for xSmall and small button sizes.

jaketrent: Great, happy to help.

packages/button/src/css/index.ts
sweeting: changed transition from all to specific properties

jaketrent (approved):
Looks like there's a git conflict. But code looks great. Please merge when green.

PR #1594, opened 2021-02-03
Feat/field block

danethurber:
Update field so it can be used as a block element where contents will stretch to full width

### DIFF
--- a/docs/src/content/components/field.mdx
+++ b/docs/src/content/components/field.mdx
@@ -58,6 +58,54 @@ const Example = () => <TextInputField />
 export default Example
 ```
 
+Fields are designed to display inline but can be changed to block display. Doing
+so will allow the contents to stretch fully to the parent's width.
+
+> You can apply additional styles, like `display`, to the outermost container using the prop `renderContainer`.
+
+```typescript
+import React, { ComponentProps, forwardRef } from 'react'
+import Field from '@pluralsight/ps-design-system-field'
+
+const BlockRenderContainer = forwardRef((props, ref) => (
+  <div ref={ref} {...props} style={{ display: 'block'}} />
+))
+
+const TextInputField: React.FC<ComponentProps<typeof Field>> = (props) => {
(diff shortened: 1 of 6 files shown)


packages/field/src/react/field.tsx
```diff
@@ -51,7 +53,7 @@ interface FieldProps extends Omit<HTMLPropsFor<'div'>, 'prefix' | 'ref'> {
   label?: ReactNode
   onClick?: MouseEventHandler<HTMLDivElement>
   prefix?: ReactNode
-  renderContainer?: ForwardRefExoticComponent<unknown>
+  renderContainer?: ForwardRefExoticComponent<RefAttributes<any>>
```
jaketrent, line 56: why the switch?

jaketrent (approved):

KaidenR (approved):

packages/field/src/react/field.tsx
danethurber: i was getting errors trying to apply the prop because the RefAttribute was necessary - i feel like i'm still learning how all these types work

jaketrent: srsly

docs/src/content/components/field.mdx
```diff
@@ -58,6 +58,54 @@ const Example = () => <TextInputField />
...
+
+const TextInputField: React.FC<ComponentProps<typeof Field>> = (props) => {
+  const { disabled, placeholder, type = 'text', ...rest } = props
+
+  const labelId = 'this-needs-to-be-a-unique-label-id-1'
+  const inputId = 'this-needs-to-be-a-unique-input-id-1'
+
+  return (
+    <Field
+      renderContainer={BlockRenderContainer}
```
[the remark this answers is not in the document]
danethurber: it allows more that just applying a style. but i do think we should explore and maybe super two methods of customization. simple class overrides and full element overrides

PR #1596, opened 2021-02-04
test(actionmenu): add getByRole tests

### LINKED ISSUE #1575: actionmenu: add spec for testing-library getByRole for Item that's disabled.
caseybaggz  2:41 PM
Thanks again for that great conversation. Here is the link to the headless components Alex & I have been designing. It will be usePinDrawerWrapper and useTabDrawer:
https://github.com/Git-Prime/flow-design-system-extension/tree/master/src/hooks
:thankyou:
2

jaketrent:glass_of_milk:  2:41 PM
Hey, thanks, Casey!
:call_me_hand:
1

caseybaggz  3:12 PM
Alright, I confirmed it’s the ActionMenu.Item that is not accessible via the chrome tools Elements > Accessibility tab because it’s an a using a role of menuItem but the parent list item is using the  attr data-disabled and not a aria-role.
I think we can just update it to this and it should be fine? Total assumption though? Maybe the list item should use the menuItem role?
<a href="#" role="button" aria-disabled="true">whatever</a>
If we have a test validating it’s disabled, then we should be good. (edited) 
3:14
In this very rare use case in Flow, it appears the design team wants to keep the dropdown clickable and not totally disable it because at some point in time, only 1 list item may be enabled. :woman-shrugging: (edited) 

jaketrent:glass_of_milk:  3:21 PM
I see, I think.  I think you’re describing a Dropdown scenario. Your use case is that the Dropdown can be open but that individual menu items may be enabled while some are disabled.  Is that right?

caseybaggz  3:24 PM
Yesh. But I know we are using ActionMenu.Item for this case. It’s this unfortunate thing (I didn’t author this, just fixed type errors):
https://github.com/Git-Prime/flow-design-system-extension/blob/master/src/components/BulkActionMenu/index.tsx

jaketrent:glass_of_milk:  3:29 PM
Oh, I see, so you’re not using Dropdown from the PSDS at all.  You’re saying that you made a custom dropdown and are using the PSDS ActionMenu as the menu.
If that’s accurate, I probably need to learn more, because ActionMenu has not had the same recent testing-library getByRole audit and testing done to it as Dropdown has.
I think you’re reporting that you can’t access the ActionMenu.Item via getByRole.  (I guess I’m implying that from the context of the convo, when you’re just saying “not accessible” and offering a potential adjustment.) Is that right?  One way or the other, would you mind sending the test line/approach that fails, any error that you may be getting and again just what you expect to have happen?

caseybaggz  AM
Morning!
Yes! That’s exactly it - you can’t really select the “disabled” list item via getByRole due to the nature of the aria roles assigned. I think if we can get the same aria audit on these, we would be able to remove dupe tests within Flow. :thumbsup:

caseybaggz  3:32 PM
Yeah! But, I’ll do that on Monday. Heading out for the weekend. See you guys then. :wave-animated:

jaketrent:glass_of_milk:  3:32 PM
Perfecto, have a great weekend.

EdwardIrby:
Resolves #1575

### DIFF
--- a/packages/actionmenu/src/react/__specs__/__snapshots__/storyshots.spec.ts.snap
+++ b/packages/actionmenu/src/react/__specs__/__snapshots__/storyshots.spec.ts.snap
@@ -16,6 +16,7 @@ exports[`Storyshots customized styles item style 1`] = `
     tabIndex={-1}
   >
     <a
+      aria-disabled={false}
       aria-haspopup={false}
       data-css-1v7ok0h=""
       onClick={[Function]}
@@ -48,6 +49,7 @@ exports[`Storyshots customized styles item style 1`] = `
     tabIndex={-1}
   >
     <a
+      aria-disabled={false}
       aria-haspopup={false}
       data-css-1v7ok0h=""
       onClick={[Function]}
@@ -80,6 +82,7 @@ exports[`Storyshots customized styles item style 1`] = `
     tabIndex={-1}
   >
     <a
+      aria-disabled={false}
       aria-haspopup={false}
       data-css-1v7ok0h=""
       onClick={[Function]}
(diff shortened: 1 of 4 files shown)


jaketrent (approved):
awesome, thanks!

PR #1597, opened 2021-02-04
MultiSelect/TagsInput

danethurber:
Adds two new components: MultiSelect and TagsInput

> NOTE: docs are not written yet. please review and i'll add docs after

tradeoffs:

- Tags focus/halo doesn't look great so the focus state is just browser default for now
- ActionMenu is not used. tried to make it happen but there were conflicting events and attrs
- Position is used but it isn't perfect. some future Position updates may be needed to address


jaketrent (commented):
Early experience:

On any of the stories, click in the input area, and I get this error:

Screen Shot 2021-02-09 at 11 38 52 AM

I did a full rebuild of DS, and still had this.

Separately, the disabled state looks funky with the box around the input area.  Is that on purpose?

Screen Shot 2021-02-09 at 11 43 10 AM

packages/multiselect/src/react/utils.ts
```diff
@@ -0,0 +1,33 @@
...
+}
+
+/* eslint-disable no-prototype-builtins */
+export function switchcase<T = unknown>(
+  cases: Record<string, T>,
+  defaultCase: T,
+  key: string
+): T {
+  return cases.hasOwnProperty(key) ? cases[key] : defaultCase
+}
```
jaketrent, line 32: What was the reason for creating this?  I'm assuming it was the only way to get a { [str]: Fn } may for selecting a Fn dynamically to TS compile.

packages/multiselect/src/react/utils.ts
```diff
@@ -0,0 +1,33 @@
+import { Option } from './types'
+
+export function arraysAreEqual<T>(a: T[] = [], b: T[] = []) {
+  return a.length === b.length && a.every(item => b.includes(item))
```
jaketrent, line 4: Array.includes doesn't work in IE11.  Do we ship a polyfill?  If not, might just want to use indexOf as is done elsewhere in the PR.

packages/multiselect/src/react/menu.tsx
```diff
@@ -0,0 +1,51 @@
+import { compose, css } from 'glamor'
```
jaketrent, line 1: Nice organization here.  I like that this is not taken to the nth degree either. That said, I did find index.ts at 340 lines to be a bit of a jumble to read all the way through.  I don't know if you feel similarly. Maybe an opp for further encapsulation of that code.

packages/multiselect/src/react/menu.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+  menuItem: (opts: { highlighted: boolean }) =>
+    compose(
+      css(stylesheet['.psds-multi-select__menu__item']),
+      opts.highlighted &&
+        css(stylesheet['.psds-multi-select__menu__item--highlighted'])
+    )
+}
+
+interface MenuProps extends Omit<HTMLPropsFor<'ul'>, 'ref'> {
```
jaketrent, line 22: Why is ref omission required? tsc still successfully w/o it for me.

packages/multiselect/src/react/utils.ts
```diff
@@ -0,0 +1,33 @@
...
+}
+
+/* eslint-disable @typescript-eslint/no-empty-function */
+export function noop() {}
+/* eslint-enable @typescript-eslint/no-empty-function */
+
+export function simpleTextFilter(term = '', options: Option[]) {
+  if (term.length === 0) return options
+
+  const needle = term.toLowerCase()
```
jaketrent, line 14: :awesome: needle.

packages/multiselect/src/react/index.tsx
```diff
@@ -0,0 +1,340 @@
...
+  SyntheticEvent,
+  cloneElement,
+  forwardRef,
+  isValidElement,
+  useCallback,
+  useLayoutEffect,
+  useMemo,
+  useRef,
+  useState
+} from 'react'
```
jaketrent, line 19: The React API surface area has definitely grown.

jaketrent (commented):
Here's another round of comments, learning about what you've created.

I started in to the tagsinput portion. When starting storybook, I get this error on page load, so I stopped at that point.  Either a problem on my end or something that needs addressed, and then I'll finish up.

Screen Shot 2021-02-09 at 2 09 47 PM

packages/multiselect/src/react/utils.ts
danethurber: i wanted a switchcase fn that used objects instead of using native switch cases. i find it easier to reason about than a native switch because there is no chance of fall-through and there is a require defaultCase. Also...it's probably faster because it's lookup by key...but that's an assumption

packages/multiselect/src/react/__fixtures__/options.ts
```diff
@@ -0,0 +1,42 @@
+export const periodicElements = [
```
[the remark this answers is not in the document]
danethurber: uh.....remind me where the npmignore is...

packages/multiselect/src/react/index.tsx
```diff
@@ -0,0 +1,340 @@
...
+
+interface MultiSelectFieldStatics {
+  Label: typeof Field.Label
+  SubLabel: typeof Field.SubLabel
+}
+
+type MultiSelectFieldComponent = React.FC<MultiSelectFieldProps> &
+  MultiSelectFieldStatics
+
+const MultiSelect: MultiSelectFieldComponent = props => {
```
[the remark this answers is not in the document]
danethurber: i'll look into adding

packages/multiselect/src/react/menu.tsx
danethurber: yeah it's pretty large. mostly glueing stuff together...i'd say we keep as is now. splitting out the component would pretty pretty easy but i don't see a huge gain right now

danethurber (commented):
@jaketrent that error should be fixed now. please continue your review'n

packages/multiselect/src/react/index.tsx
```diff
@@ -0,0 +1,340 @@
...
+    return <Field.Label {...getLabelProps()}>{label}</Field.Label>
+  }, [label, getLabelProps])
+
+  const SubLabel = useMemo(() => {
+    if (isValidElement(subLabel)) return subLabel
+
+    return <Field.SubLabel>{subLabel}</Field.SubLabel>
+  }, [subLabel])
+
+  const afterInputSentinel = useRef<HTMLDivElement>(null)
```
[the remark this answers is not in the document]
danethurber: changed to mark 🤷

packages/multiselect/src/react/__fixtures__/options.ts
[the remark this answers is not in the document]
jaketrent: It's in the root of the package dir for most packages.  See `packages/actionmenu/.npmignore`

packages/multiselect/src/react/index.tsx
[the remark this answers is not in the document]
jaketrent: 🤷 Yeah, I'm not sure either. I wasn't intending to encourage a change. I was just trying to learn the meaning as envisioned.

packages/multiselect/src/react/index.tsx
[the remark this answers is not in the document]
jaketrent: Did you look into it?  What did you decide?

packages/tagsinput/src/react/index.tsx
```diff
@@ -0,0 +1,218 @@
...
+    if (isValidElement(subLabel)) return subLabel
+
+    return <Field.SubLabel>{subLabel}</Field.SubLabel>
+  }, [subLabel])
+
+  const afterInputMark = useRef<HTMLDivElement>(null)
+  useLayoutEffect(
+    function keepInputInView() {
+      if (!afterInputMark.current) return
+      if (!afterInputMark.current.scrollIntoView) return
```
jaketrent, line 119:
```suggestion
      if (!afterInputMark?.current?.scrollIntoView) return
```

Possibly?

jaketrent (approved):
Code looks good.  I have made a couple followup comments.

Generally, I have a couple observations:

1. It looks like the scrollbar is always visible.  Might be nice to hide that when not required.

Screen Shot 2021-02-10 at 4 43 46 PM
Screen Shot 2021-02-10 at 4 42 22 PM

2. I can't quite figure out why TagsInput might have an editable input. I'm not sure even what the ux is intended to be with the component. And there are no user docs in this PR to confirm the design thinking. I can just compare Multiselect and TagsInput. Multiselect seems more of a typeahead. In fact, it feels like a replacement Typeahead that can handle multiple selections. TagsInput can't type to search, but it can rack up pills.  Odd, then, to me that I can still type but not press enter to submit or apparently affect anything. What am I missing in the intended usage?

Screen Shot 2021-02-10 at 4 53 53 PM

danethurber:
@jaketrent i'll try to explain the difference in the components/patterns

- Typeahead is a single value that using the combobox aria pattern (would like to update to use Field and downshift [useCombobox](https://www.downshift-js.com/use-combobox))
- Multiselect allows multi values and uses [two hooks](https://www.downshift-js.com/use-multiple-selection#usage-with-combobox) from downshift that are combined to get the right interactions
- TagInput is a little special in that it's multiple values without a combobox - so it's a different enough pattern that the logic and aria needs are different. I do agree that it's weird that i can't hit enter to add a pill. i'm now thinking that i should add an example of that.

@sweeting do you want to help me outline the differences and why you would use one over the other. i'm going to start on the docs now that we've gotten to this point

EdwardIrby:
Docs coming in separate PR?

danethurber:
@EdwardIrby i'm writing them now. i'll merge with docs

packages/multiselect/src/react/index.tsx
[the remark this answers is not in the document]
danethurber: we need to explore better ref support with the Field. these refs get tricky. i'm going to merge this pr and then later explore handling refs better with Field

sweeting:
The height of the field should be 40px, but for some reason there's a 2px margin on all sides of the pills causing it to be 44px.
CleanShot - Safari -2021-02-12 at 11 35 34

Multi Select docs:
The text in the menu on dark theme is white
CleanShot - Safari -2021-02-12 at 11 26 50

When the menu is open and I hit tab, it inserts a pill. I expected hitting tab to close the menu and return focus to the input.

When the input is focused and pills are in the field, I expected hitting the delete key would focus the most recently added pill and then hitting delete key again would remove it like this example in Google's sharing dialog.
CleanShot - Safari -2021-02-12 at 11 37 07

PR #1599, opened 2021-02-06
dropdown: optionally pop upwards

culli:
When a dropdown is near the bottom of the viewport it can be cut off.  This change calculates the remaining space and if it's not enough it pops upward instead.

I don't feel like I know enough about react-testing-library to write a good test for this, but it might look something like the tests in `position` for above/below/etc.  I'm happy to work with y'all on that.

It could probably also use some styling enhancements since the `transform: translate()` looks slightly strange when popping above the component, it sort of slides down from above.  And I hard-coded the reverse 8px for the 4px margin the dropdown normally has.


jaketrent (commented):
I'm a fan. I don't think it's in the design yet, but it seems reasonable that it should be.  

Team, opinions?

jaketrent:
Team triage: Team feels good from a design perspective. Will make build green, and approve/merge.

PR #1608, opened 2021-02-11
new icon size and updated button

sweeting:
`icon`
- new size xsmall (16x16) changed small to (20x20)

`button`
- spacing tweaks to match design
- using new small & xsmall icon sizes

`docs`
- Updated docs with new icon size


packages/button/src/react/index.tsx
```diff
@@ -76,8 +79,8 @@ const styles = {
 
 const mapIconSize = (size: string) => {
   const btnToIconSizes = {
-    [vars.sizes.xSmall]: iconSizes.small,
-    [vars.sizes.small]: iconSizes.medium,
+    [vars.sizes.xSmall]: iconSizes.xSmall,
+    [vars.sizes.small]: iconSizes.small,
```
EdwardIrby, line 83: This is suppose to be 20 right

packages/button/src/css/index.ts
```diff
@@ -83,25 +81,25 @@ export default {
   [`.psds-button--size-${vars.sizes.xSmall}`]: {
     fontSize: type.fontSize100,
     letterSpacing: type.letterSpacingLooser,
-    padding: `0 ${layout.spacingXSmall}`,
+    padding: `4px 8px`,
```
EdwardIrby, line 84: Thinking we want to use layout spacing values where possible. I can help you with that if you need it.

EdwardIrby (changes requested):

packages/button/src/css/index.ts
sweeting: @EdwardIrby I specifically removed them inside the component because there aren't spacing variables for all of the values I needed. So in some cases I'd end up with a mix like this:

`padding: \`6px ${layout.spacingSmall}\``

I wasn't really sure what value the variable was providing. Using the actual pixel values seemed clearer to me. I'm not sure about you, but I have to look up the pixel values for each spacing variable whenever I use them.

packages/button/src/react/index.tsx
sweeting: Yes, small is now 20x20 and xSmall is 16x16.

EdwardIrby (approved):

Issue #1598, opened 2021-02-05
dropdown: subsequent onChanges aren't called


jaketrent:
I'm having issues with the latest version of DropDown > 11.  It allows me to make one selection, but then it doesn't save any subsequent changes to the selection.  I'm seeing the same problem on the CodeSandbox example - https://codesandbox.io/s/2ruxy?module=/example.tsx.  Monday problem, but wanted to post it now so I don't forget next week!

https://pluralsight.slack.com/archives/C70HPSJPP/p1612568051061700

EdwardIrby:
Cause is this PR #1592. Investigating why adding capture as [documented](https://reactjs.org/blog/2020/08/10/react-v17-rc.html#fixing-potential-issues) causes failure. Pretty sure it's a DOM event lifecycle thing.

PR #1611, opened 2021-02-12
feat: use onGlobalEventsClose instead of hook

EdwardIrby:
React 17 definitely made global event handling different and ref changed slightly as well it seems.

Actionmenu: https://pluralsight.slack.com/archives/C70HPSJPP/p1613085725106500
Dropdown: https://github.com/pluralsight/design-system/issues/1598 https://codesandbox.io/s/2ruxy?module=/example.tsx

So the useCloseOnDocumentEvents hook needs to be configured differently for dropdown and actionmenu. Drodpwon needs to track ref and reapply effect on re-render. Actionmenu does not and it actually throws a bug as seen in the slack thread above. 

It's easier to just use the onGlobalEventsClose helper and add a useEffect hook. What do you think?


danethurber:
@EdwardIrby i'm trying to catch up here. seems like there are two issues:

1. in the Dropdown, onChange isn't firing when the items are clicked
2. in ActionMenu, the menu is accidentally closed when clicking inside the menu

^ is that correct?

I think both of the issues are fixed. I am wondering if the Dropdown should close when clicking outside the menu - it does not do that for me.

packages/actionmenu/src/react/index.tsx
```diff
@@ -73,8 +74,20 @@ export const ActionMenu = React.forwardRef<HTMLUListElement, ActionMenuProps>(
...
+
+      if (ref.current) {
+        removeListeners = onGlobalEventsClose<HTMLUListElement>(
+          ref.current,
+          onClose
+        )
+      }
+
+      return removeListeners
+    }, [])
```
danethurber, line 90: should `onClose` be in the dependency array?

packages/dropdown/src/react/menu.tsx
```diff
@@ -69,7 +70,20 @@ export const Menu = forwardRef<HTMLDivElement, DropdownMenuProps>(((
...
+
+    if (ref.current) {
+      removeListeners = onGlobalEventsClose<HTMLDivElement>(
+        ref.current,
+        context.onDocumentEvents
+      )
+    }
+
+    return removeListeners
+  }, [ref])
```
danethurber, line 86: should `context.onDocumentEvents ` be in the dependency array?

matthewma7:
Bump this

matthewma7:
I noticed another issue probably related to this.
When a new handler is passed to the `onClose` prop, the <ActionMenu> will still call the old one instead of the new one.
image

packages/dropdown/src/react/menu.tsx
```diff
@@ -84,14 +75,21 @@ export const Menu = forwardRef<HTMLDivElement, DropdownMenuProps>(((
         top: buttonRect.top - menuRect.height - 8
       })
     }
+    isOpen && ref.current.focus()
```
EdwardIrby, line 78: Think I can remove this line

packages/dropdown/src/react/label.tsx
```diff
@@ -17,15 +17,14 @@ const styles = {
...
-interface LabelProps {
+interface LabelProps extends HTMLPropsFor<'label'> {
   label?: ReactNode
-  inputId: string
 }
 
 export const Label: FC<LabelProps> = props => {
   const themeName = useTheme()
   return (
-    <label htmlFor={props.inputId}>
```
EdwardIrby, line 28: Think I should put this back

packages/dropdown/src/react/item.tsx
```diff
@@ -30,28 +39,44 @@ interface DropdownItemProps
...
     const valueExists = typeof value !== 'undefined'
-    const isActive =
-      (valueExists && context.activeItem?.value === value) ||
-      context.activeItem?.label === children
     const isSelected =
       (valueExists && context.selectedItem?.value === value) ||
       context.selectedItem?.label === children
 
-    const handleClick = (evt: React.MouseEvent) => {
+    const handleClick = (evt: React.SyntheticEvent) => {
```
danethurber, line 60: is this always going to me a MouseEvent? if so i'd suggest using MouseEvent or MouseEventHandler so you can avoid casting the events below

packages/dropdown/src/js/index.ts
```diff
@@ -114,26 +110,21 @@ export const useDropdown = (
   props: UseDropdownProps,
   forwardedRef: Ref<HTMLButtonElement>
 ) => {
+  const id = uniqueId()
```
danethurber, line 113: please memoize. this will change on each render and could cause unnecessary dom updates

packages/dropdown/src/react/__specs__/storyshots.spec.ts
```diff
@@ -3,10 +3,20 @@ import initStoryshots, {
...
   test: snapshotWithOptions({ createNodeMock })
 })
+
+beforeEach(() => {
+  jest.spyOn(global.Math, 'random').mockReturnValue(0.123456789)
+})
+
+afterEach(() => {
+  jest.spyOn(global.Math, 'random').mockRestore()
+})
```
danethurber, line 22: if you're having testing issues with this mock - you can also mock out the util+hook

[example](https://github.com/pluralsight/design-system/blob/master/packages/field/src/react/__specs__/storyshots.spec.ts#L6)

packages/dropdown/src/react/label.tsx
danethurber: i don't think using the `for` attr works to reference a `button` element. using this [example](https://www.w3.org/TR/wai-aria-practices/examples/listbox/listbox-collapsible.html) aria-labelledby seems to be the right approach. aria-labelledby should reference the label and the button.

packages/util/src/use-close-on-document-events.ts
```diff
@@ -7,12 +7,14 @@ export const onGlobalEventsClose = <El extends HTMLElement>(
...
   callback: Callback
 ) => {
-  if (!canUseDOM()) return
+  if (!canUseDOM()) return noop
 
   const handleClickOutsideMenu = (evt: MouseEvent) => {
     if (evt.target instanceof HTMLElement) {
       if (el.contains(evt.target)) return
-      callback(evt)
+      ;(function cb(): void {
```
danethurber, line 15: what's going on here. nextTick maybe. trying to fix a race condition?

danethurber (changes requested):
awesome work! interaction seems to work well for me. i have a couple suggestions/comments. I also noticed some issues in the stories. maybe these are preexisting problems and should be carded to fixed later. please review and see if these issues are related to the PR:

1. Focus style has native browser outline. i believe this is a regression and should be addressed:

Screen Shot 2021-02-23 at 10 17 27 AM

2. Story: layouts > bottom of viewport. The menu appears above the button which seems incorrect based on button's label

Screen Shot 2021-02-23 at 10 17 59 AM


3. Story: layouts > custom width. component width seems much wider than expected.

Screen Shot 2021-02-23 at 12 42 33 PM

4. Story: menu > max height. story/example is rendering in an unusable state
Screen Shot 2021-02-23 at 10 19 15 AM


cc @jaketrent - this pr will switch the aria pattern Dropdown uses from combo-box(using a hidden input) to the [listbox pattern](https://www.w3.org/TR/wai-aria-practices/examples/listbox/listbox-collapsible.html). I agree that this is the appropriate change to make - but i'm wondering if you provide some quick context on why you chose the combobox pattern in the past.

packages/dropdown/src/js/index.ts
```diff
@@ -50,7 +48,6 @@ interface UseDropdownProps extends Omit<HTMLPropsFor<'button'>, 'onChange'> {
   size?: ValueOf<typeof vars.sizes>
   style?: React.CSSProperties
   subLabel?: ReactNode
-  uniqueId?: (prefix: string) => string
```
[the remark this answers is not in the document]
EdwardIrby: The mocking pattern you shared should be sufficient for most teams no? Also those id's really do need to be unique for accessibility. We'd have to rely on teams actually adding their own when implementing. I kinda like the id of allowing a data-test-id to be used instead for testing. What do you think?

packages/dropdown/src/react/item.tsx
EdwardIrby: I also use this same function for keydown `Enter`. Could simply make a copy but for keydown was reducing lines.

packages/dropdown/src/react/__specs__/storyshots.spec.ts
EdwardIrby: It works locally but this might solve the CI bug. Thanks!

packages/util/src/use-close-on-document-events.ts
EdwardIrby: Opps, part of my earlier debugging and forgot to revert

packages/dropdown/src/js/index.ts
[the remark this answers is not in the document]
danethurber: i wish mocking would work but as of yet teams have said they can't do it and need the DI pattern to override in test. mocking probably would work but i haven't dug into it at all - i just know it's the reason that a couple components have a `uniqueId` prop. 

Yes i agree we need the uniqueId so we should still be handling that. the props is for overriding specifically for snapshot testing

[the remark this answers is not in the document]
danethurber: @EdwardIrby this is where we are allowing a custom `uniqueId` prop to be provided. this is the DI pattern that I was referencing in standup were the `uniqueId` prop can be override in test by consumers

cc @jaketrent thoughts? i know this was recently added to Carousel as well and I see this as an emerging standard. is this something we should continue to support - or is using DI for unique id the wrong approach and we could stepback and rethink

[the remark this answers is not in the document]
jaketrent: My thoughts:

- jest.mock of @ps/ds-util didn't work. By the time we were mocking, the subdep package name was not addressable.
- jest.mock of Math.random does seem like a impl detail. In a way jest.mock of @ps/ds-util is as well, just not as low-level. It's also something we control.
- uniqueId props seem like a simple solution to this problem. 
- I'm not super surprised that "uniqueId" as a name caused confusion.  Maybe it needs something that shows it to be test-centric (eg, TESTONLY_uniqueId) or private (eg, _uniqueId). 
- My favorite solution would be to not have a uniqueId prop. This would require on-page uniqueness and stability between runs. Stability could come from the auto-incrementer (a la downshift) that was mentioned. I like that. But what about on-page uniqueness.  You'd have to auto-increment within a namespace per Dropdown. Where would that uniqueness come from if not a gen'ed uuid? Hash some other data (eg, labels and menu items)?

packages/dropdown/src/react/item.tsx
jaketrent: Have done a union type in other code, eg:

```
React.MouseEvent | React.KeyboardEvent
```

And then done type narrow (no cast) in the body as needed.

packages/dropdown/src/js/index.ts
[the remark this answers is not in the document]
EdwardIrby: I'm going to put back uniqueID. It is confusing ands seems like a thing I'd need to add as a consumer of the comonent. I honestly wish this was just a data-testid because that's more explicit. But this is blocking other teams so let's keep as is.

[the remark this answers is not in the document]
EdwardIrby: @jaketrent is there a reason we're putting [formated i'ds](https://github.com/pluralsight/design-system/blob/4f79ccea73ba982bde80c5dfa2b9d05518ba915d/packages/dropdown/src/js/index.ts#L340) on Items. It's not for a11y so I'm just trying to understand the reasoning behind it.

packages/dropdown/src/react/item.tsx
EdwardIrby: I'll give that a shot just to be clear this an internal function you know? Like how important is this so long as the onClick passed to the component has the signature we desire?

danethurber (approved):
👍  LGTM

PR #1617, opened 2021-02-16
banner: a11y & design true-up

sweeting:
`Banner`: contrast boost and type tweaks

- Text meets WCAG AA contrast guidelines (4.5:1) for normal text
- Change font-weight to Default from 500
- Change font-size and letter-spacing to match Paragraph Medium text style


jaketrent (approved):
Looks good to me.  Is there an opp here to use the new `Text.Label` component instead of defining the text styles?

sweeting:
@jaketrent That's definitely how I envisioned using the Text components within these kinds of components, but it's beyond my ability. I'd be happy to pair on it with you or create an issue for someone to pick up.

jaketrent:
@sweeting Would love to pair.  I think that'd be fun.  Name the date!

PR #1623, opened 2021-02-19
fix(multiselect): keybindings

danethurber:
@sweeting this should be the interaction we talked about - please review and let me know how it feels


packages/multiselect/src/react/use-keybinding.ts
```diff
@@ -0,0 +1,91 @@
+import { useCallback, useEffect, useMemo, useReducer } from 'react'
+import { isString } from '@pluralsight/ps-design-system-util'
+import { switchcase } from './utils'
+
+type Callback = () => void
+type State = Record<string, boolean>
+type Action = { type: string; data?: State; key?: string }
+
+export function useKeybinding(
```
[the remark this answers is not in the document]
EdwardIrby: Until proven otherwise I say go with it for now

EdwardIrby:
Is there some design review available from @sweeting. I'm thinking this is based on [this](https://www.figma.com/file/tWOE9v2OLa2BuRT0vVHamP/Master%3A-Assignment-Components?node-id=1531%3A0) I'm a bit worried width of the input and dropdown. Also would of like a reference design showing various states to compare too.

danethurber:
@EdwardIrby this pr is just focused on keybindings. the figma link you sent does not use a MultiSelect component - can explain what you're referencing?

packages/multiselect/src/react/__specs__/index.spec.tsx
```diff
@@ -33,6 +45,97 @@ describe('MultiSelectField', () => {
       expect(results).toHaveNoViolations()
     })
   })
+
+  describe('Basic story', () => {
+    const { Basic } = stories
+
+    it('should focus input when label clicked', () => {
+      render(<Basic {...(Basic.args as any)} />)
```
jaketrent, line 53: Why the cast?

packages/multiselect/src/react/index.tsx
```diff
@@ -201,11 +207,20 @@ const MultiSelect: MultiSelectFieldComponent = props => {
...
+      }
     })
   )
 
+  useKeybinding(
+    ['alt', 'arrowdown'],
+    () => {
+      if (!isOpen) return
+
+      window.requestAnimationFrame(closeMenu)
```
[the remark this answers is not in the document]
jaketrent: ng-wait :D

packages/multiselect/src/react/__specs__/index.spec.tsx
```diff
@@ -33,6 +45,97 @@ describe('MultiSelectField', () => {
...
+      expect(input).toHaveFocus()
+      expect(menu).toBeVisible()
+
+      userEvent.type(input, '{esc}')
+
+      expect(input).toHaveFocus()
+      expect(menu).not.toBeVisible()
+    })
+
+    it('should open the menu when alt+up pressed', async () => {
```
jaketrent, line 105: Would you please reference the WCAG guideline so I can see the guidance around binding "alt+up/down" for menus?

packages/multiselect/src/react/use-keybinding.ts
```diff
@@ -0,0 +1,91 @@
...
+        ...acc,
+        [key.toLowerCase()]: false
+      }),
+      {}
+    )
+
+    return map
+  }, [binding])
+
+  const [keys, setKeys] = useReducer(reducer, initialMapping)
```
jaketrent, line 26: Why is tracking the pressed keys in internal state required?

jaketrent (commented):
Took a look through the code.  I'd like to test the interaction too.  Will you please describe what you're fixing and what the intended interaction now is?

packages/multiselect/src/react/__specs__/index.spec.tsx
danethurber: downshift provides all the WCAG needed interactions but @sweeting and I didn't feel like it was complete so we've enhanced the interaction following this component library's docs: https://www.telerik.com/kendo-react-ui/components/dropdowns/multiselect/keyboard-navigation/

here is their a11y statement https://www.telerik.com/kendo-react-ui/components/dropdowns/multiselect/accessibility/

packages/multiselect/src/react/use-keybinding.ts
danethurber: so i can get a combo. alt+up for example. this was the best approach i could come up with but i'm open to alternatives

packages/multiselect/src/react/__specs__/index.spec.tsx
danethurber: storybook typings don't infer all the types. i figured since it was a test and the stories are typed this was an ok recast

packages/multiselect/src/react/__specs__/index.spec.tsx
jaketrent: Would you say that MultiSelect follows the combobox pattern?

If so, looking at the WCAG recs here:
https://www.w3.org/TR/wai-aria-practices-1.2/examples/combobox/combobox-autocomplete-list.html

Looks like the menu will pop open on regular down arrow if it's not open.  That seems both similar and simpler.

Would that feel complete?  cc @sweeting

packages/multiselect/src/react/use-keybinding.ts
jaketrent: Could you use just a regular eventHandler that detects the alt key?

Pseudo code:

```
useEffect
  node.addEventListener 'keydown', evt => 
    if evt.key is down arrow and evt.altKey
      open menu
```

danethurber: 🤦  yeah...i'll do that. thanks for the coffeescript :)

packages/multiselect/src/react/__specs__/index.spec.tsx
danethurber: up and down arrow are already used when the menu is **open** - they move selection of the focused item.

`alt+up` is really the only new interaction which closes the menu when `up` isn't a viable option. looking at the description of this test i might have gotten up/down switched up

danethurber: here is the standard. they are optional events

https://www.w3.org/TR/wai-aria-practices/#keyboard-interaction

danethurber: here is the less readable reference https://www.w3.org/TR/wai-aria-practices/#textbox-keyboard-interaction

alt+up/down are option i guess 🤷

packages/multiselect/src/react/use-keybinding.ts
danethurber: updated!

packages/multiselect/src/react/__specs__/index.spec.tsx
EdwardIrby: I kinda feel this thing this is both a comboxbox and pills. 
So really two components composited together. 

https://www.lightningdesignsystem.com/components/pills/#Accessibility
https://www.lightningdesignsystem.com/components/combobox/#Keyboard-Interaction

The whole alt+up is not really something I'm familiar with. Did we check with Amber to see if that was a common interaction pattern. Also I kinda like some of the rules found here and applying them to the pills.

danethurber: yes it's very similar to the [`TagsInput`(pills) component](https://github.com/pluralsight/design-system/blob/master/packages/tagsinput/src/react/index.tsx#L71). the `TagsInput` component uses downshift [useMultipleSelect](https://www.downshift-js.com/use-multiple-selection) but does not employ a combobox so there is a large enough difference that i separated them initially. this is the [PR](https://github.com/pluralsight/design-system/pull/1597) that introduced the TagsInput and MultiSelect components

some references you can explore if you're interested.
- here is the [downshift guide](https://www.downshift-js.com/use-multiple-selection#usage-with-combobox) that provided the initial implementation
- [here](https://github.com/microsoft/sonder-ui/tree/master/src/components/multiselect) is one of the best implementations i had initially research with an accompanying [article testing various approaches](https://www.24a11y.com/2019/select-your-poison-part-2/)

As for the keybindings: [here](https://www.w3.org/TR/wai-aria-practices/#textbox-keyboard-interaction) are the combobox keyboard interaction docs and they explain the alt+up/down but they are marked as optional. I think running it by amber is a great idea - i'll do that now

danethurber: amber gives the 👍 . her response

> Yes! I use it ALL the time, even when I'm not testing for a11y!

EdwardIrby (approved):

PR #1625, opened 2021-02-19
fix core module path

jaketrent:
Anything that detects modules and pulls code through this entry will fail, because js/index.js doesn't exist.

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -7,7 +7,7 @@
   "author": "jaketrent",
   "keywords": [],
   "main": "dist/index.js",
-  "module": "js/index.js",
+  "module": "js/index.ts",
   "scripts": {
     "build": "run-s build:js build:css",
     "build:css": "node ./build",


packages/core/package.json
```diff
@@ -7,7 +7,7 @@
   "author": "jaketrent",
   "keywords": [],
   "main": "dist/index.js",
-  "module": "js/index.js",
+  "module": "js/index.ts",
```
danethurber, line 10: can you use `ts` as a module. what about 'dist/index.js'

jaketrent:
nm, this won't work, because we need an esm built all the way down. The ts import rules are diff: eg, no extension.

packages/core/package.json
jaketrent: nope, can't. I was hoping, but then I tried it. :)  dist/index.js is cjs, so no go there.  We really should just remove this module line, because I don't think we generate a true module in core. Well, better, I guess would be to gen it.  Maybe I'll do that.

Issue #1630, opened 2021-02-24
TagsInput: adding icon to prefix increases component height


KaidenR:
### Expected behavior
Adding an icon via the prefix prop should not affect the component's height

### Actual behavior
We added `prefix={<SearchIcon />}` and the height of the TagsInput grew from 40px to 42px

### Related packages

- tagsinput@1.1.3

### Environment

- OS - MacOS
- Browser version - Chrome

PR #1635, opened 2021-02-25
maintain height for field with prefix and other tweaks

jaketrent:
Resolves: #1630

And other design tweaks.  Comments inline.


packages/field/src/css/field.ts
```diff
@@ -32,12 +32,19 @@ export default {
     color: colorsTextIcon.highOnLight,
     display: 'flex',
     fontWeight: type.fontWeightDefault,
-    padding: layout.spacingSmall,
+    padding: `0 ${layout.spacingMedium}`,
```
jaketrent, line 35: 0 out padding. Use `height` on field to guarantee it. horz padding to match figma.

packages/field/src/css/input.ts
```diff
@@ -12,12 +14,17 @@ export default {
     outline: 'none',
     padding: 0,
     width: '100%',
+    height: '38px',
```
jaketrent, line 17: 40px - 2px for vert border

packages/field/src/css/input.ts
```diff
@@ -12,12 +14,17 @@ export default {
...
     '&:focus': { outline: 'none' },
 
     '&[type=search]::-webkit-search-decoration': { appearance: 'none' },
     '&[type=search]::-webkit-search-cancel-button': { appearance: 'none' },
     '&[type=search]::-webkit-search-results-button': { appearance: 'none' },
     '&[type=search]::-webkit-search-results-decoration': { appearance: 'none' }
+  },
+
+  [`.psds-field__input--${sizes.small}`]: {
+    height: '30px'
```
jaketrent, line 28: 32px - 2px for vert border

packages/field/src/css/field.ts
```diff
@@ -48,6 +55,7 @@ export default {
     alignSelf: 'stretch',
     color: colorsTextIcon.lowOnLight,
     display: 'flex',
+    alignItems: 'center',
```
jaketrent, line 58: reqd w/o padding

packages/field/src/css/field.ts
```diff
@@ -32,12 +32,19 @@ export default {
...
   },
   [`.psds-field--${sizes.small}`]: {
-    padding: layout.spacingXSmall
+    padding: `0 ${layout.spacingXSmall}`
+  },
+  '.psds-field--prefix': {
+    paddingLeft: '11px'
+  },
+  '.psds-field--suffix': {
+    paddingRight: '11px'
```
jaketrent, line 47: to match figma; tighter horz padding when pre/suffix

packages/field/src/css/text-area.ts
```diff
@@ -1,8 +1,12 @@
+import { type } from '@pluralsight/ps-design-system-core'
+
 export default {
   '.psds-field__text-area__container': {
     label: 'field__text-area__container',
 
-    flex: '1'
+    flex: '1',
+    paddingTop: '10px',
+    paddingBottom: '10px'
```
jaketrent, line 9: textarea has no set height; need to reintro vert padding.

packages/field/src/react/__stories__/text-area-field.story.tsx
```diff
@@ -87,9 +87,6 @@ const Template: Story<ComponentProps<typeof TextAreaField>> = args => {
 
 export const Basic = Template.bind({})
 
-export const SizeSmall = Template.bind({})
```
jaketrent, line 90: rm unsupported use cases.

packages/field/src/react/context.ts
```diff
@@ -0,0 +1,12 @@
+import { ValueOf } from '@pluralsight/ps-design-system-util'
```
jaketrent, line 1: added to share size prop info with `Input`.

danethurber (changes requested):
Screen Shot 2021-02-25 at 12 05 21 PM

some issues on TagsInput and MultiSelect where the tags are escaping their bounds

jaketrent:
^ Wait, isn't that a part of the newly-improved design? :D  Hehe, yeah, I'll make sure to spin up those stories and check that out.  Thanks.

jaketrent:
This is now TagsInput:

TagsInputCustomInput

And MultiSelect:

Multiselect

I think that's right, but I'm not sure.  Please confirm w a re-review.

packages/field/src/css/input.ts
EdwardIrby: Are we not using `box-sizing: border-box`?

jaketrent: Hey, yeah, good point.  Why am I doing this? Why is this required? We're using border-box in normalize.

https://github.com/pluralsight/design-system/blob/master/packages/normalize/css/index.css#L20

EdwardIrby:
> This is now TagsInput:
> 
> TagsInputCustomInput
> 
> And MultiSelect:
> 
> Multiselect
> 
> I think that's right, but I'm not sure. Please confirm w a re-review.

I"m still seeing some height issues on[Tagsinput](http://localhost:6006/?path=/story/components-tagsinput--react-node-label) after a new clean and build off the most recent commit.

Screen Shot 2021-02-26 at 10 51 22 AM

jaketrent:
^ Well, what's up with that?  TagsInput diff for me and you.  Which story is this on, @EdwardIrby ?

jaketrent:
^ Ok, yeah, I rebuild and get the same thing.  Not sure how I had something else happening.  Will work on this.  Thanks!

jaketrent:
Padding was not sticking because of selector order. Maybe storybook hot loading worked in my (bad) favor last time. Props for style overrides are spread after existing data-css attrs, so this might just be a glamor limitation. Just throw it on the bonfire. Tried !important and felt pad about it [sic]. Hohoho.  Switched to add new container with pad for the TagsInput and Multiselect components.

Ready for a re-review.

EdwardIrby (approved):
LGTM!, just update snapshots

jaketrent:
@danethurber , we didn't talk too much about this.  Apart from the result, which should fix the bug and align visual design, what do you think of the *approach*? Do you like it? Have alternative thoughts? Would love to hear your opinion.

danethurber (approved):
LGTM. couple quick questions.

- i think the approach you wanted me to 👍  is adding padding to the div containing the pills. well 👍 
- why are the styles for the pill container different on TagsInput than on MultiSelect. what about the two components necessitated a different styling solution?
- the DatePicker snapshots should be unrelated. is the fix going to work long term? or should we card it and provided a more stable solution?

jaketrent:
Hey @danethurber , thanks for the weigh in.

- Thanks for the 👍 .  I'll go ahead and merge.
- Why was the css for TagsInput diff vs MultiSelect. Not sure. The principle is the same and the resulting height is the same, but the innards of each component were apparently different enough to need diff px vals on the padding. And the overflow rule on the one.
- Yes, agreed that DatePicker snapshots are unrelated. I just updated to get full build green. I don't think the fix will be long term. I'll card.

PR #1640, opened 2021-03-01
fix(typeahead): add z-index to typeahead dropdown and update dropdown

### LINKED ISSUE #1631: Typeahead: menu layer is placed behind the modal overlay making it unusable
when a typeahead is used inside of the Dialog(modal=true) the typeahead menu falls behind the dialog in stacking order.

https://codesandbox.io/s/codesandboxer-example-forked-wqppx?file=/example.tsx

EdwardIrby:
resolves #1631

### DIFF
--- a/packages/typeahead/src/css/index.ts
+++ b/packages/typeahead/src/css/index.ts
@@ -3,7 +3,8 @@ import {
   colorsBlue,
   colorsTextIcon,
   layout,
-  type
+  type,
+  layers
 } from '@pluralsight/ps-design-system-core'
 
 export default {
@@ -21,7 +22,8 @@ export default {
     maxWidth: '320px',
     minWidth: '160px',
     overflowY: 'auto',
-    padding: `${layout.spacingXXSmall} 0`
+    padding: `${layout.spacingXXSmall} 0`,
+    zIndex: layers.globalDropdown
   },
   '.psds-typeahead__menu__item': {
     alignItems: 'center',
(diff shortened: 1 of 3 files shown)


packages/typeahead/src/css/index.ts
```diff
@@ -21,7 +22,8 @@ export default {
     maxWidth: '320px',
     minWidth: '160px',
     overflowY: 'auto',
-    padding: `${layout.spacingXXSmall} 0`
+    padding: `${layout.spacingXXSmall} 0`,
+    zIndex: layers.globalDropdown
```
jaketrent, line 26: We probably have a lack of clarity on whether we should use these layers for global nav only or for other stuff in the DS.  Put another way, should we put in-page dropdowns at a lower z-index than the global nav, which is layered on top of the page app (in whole)?

EdwardIrby: Oh didn't realize that. Hmm Probably @sweeting got a value?

EdwardIrby: I think this is okay as is. Do you imagine a situation where this dropdown and another dropdown type component or the global nav dropdown will all be open and visible in the same space?

jaketrent: Ok, that's fine. I don't have a better answer. I suppose when we run into the concrete problem, we can adjust.

Yes, I can see a scenario.  Top-left global nav profile menu is open while a typeahead table filter in the top-left of the page are open at the same time. `#ps-main` (the app) comes *after* the profile menu in `#ps-chrome`, so the app typeahead appears on top of the other items on the same z-index. Should it?  (assuming that seems plausible.) Definitely corner case and probably doesn't matter.

PR #1647, opened 2021-03-02
text: add mono to label

### LINKED ISSUE #1602: text.label: add mono prop
To render monospace font instead of regular.

Wait on hosting the monofont and the normalize/other solution that's decided upon in #1563

jaketrent:
Resolves #1602

### DIFF
--- a/docs/src/content/components/text-label.mdx
+++ b/docs/src/content/components/text-label.mdx
@@ -109,6 +109,30 @@ const Example: React.FC = props => {
 export default Example
 ```
 
+### Monospace font
+
+The `mono` prop will render the label in a monospace font.
+
+```typescript
+import { Label } from '@pluralsight/ps-design-system-text'
+import React from 'react'
+
+const Example: React.FC = props => {
+  return (
+    <React.Fragment>
+      <div>
+        <Label mono>Label - mono</Label>
+      </div>
+      <div>
+        <Label mono strong>Label - strong! mono</Label>
+      </div>
+    </React.Fragment>
+  )
+}
+
+export default Example
+```
+
 ## Usage & types
 
 <Usage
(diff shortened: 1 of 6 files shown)


docs/src/content/components/text-label.mdx
```diff
@@ -132,15 +156,21 @@ export default Example
     default={Label.colors.primary}
   />
   <TypesProp
```
jaketrent, line 158: alpha ordered, hence the git noise.

This is the only significant addition.

packages/normalize/css/index.css
```diff
@@ -17,6 +17,55 @@
   font-style: italic;
 }
 
+@font-face {
```
jaketrent, line 20: Adds 1Kb to built stylesheet.

sweeting: @jaketrent What about the weight added from the font files? The monospace font is pretty seldom used, is there a different way this could be done where it's pulled in when needed rather than adding it to normalize where it will request the monospace fonts whether they're used or not?

jaketrent: Hehe, yeah, I verified that. And I've wondered about it several times, so I wrote down what I learned: https://jaketrent.com/post/when-font-face-is-downloaded

Yes, the font file weight will be pulled in when needed.

sweeting: Nice!

packages/normalize/css/index.css
```diff
@@ -17,6 +17,55 @@
...
+  font-family: 'DM Mono';
+  src: url('https://fonts.pluralsight.com/dm-mono/v1/dm-mono-light.ttf')
+    format('truetype');
+  font-display: swap;
+  font-weight: 300;
+  font-style: normal;
```
sweeting, line 26: Adding `font-feature-settings: 'ss05' on;` to the DM Mono `@font-face` declarations will switch to a better stylistic alternate of the character 'f'.

danethurber (commented):
LGTM. i'm not approving since brian has some unresolved questions

Issue #1607, opened 2021-02-11
DatePicker: support additional props


danethurber:
DatePicker convenience wrapper should support Field props.

The previous DatePicker component supported many additional props such as error, label, sublabel, appearance, disabled, size, placeholder, etc. Using the new Field component, we can add support for these props that all form controls need. If the Field component doesn't currently support a props in the list above - we should update the Field component to support it.

MultiSelect is an example of using the Field component for convenience wrappers https://github.com/pluralsight/design-system/pull/1597.

PR #1649, opened 2021-03-03
datepicker: support all field props

jaketrent:
Resolves: #1607


packages/datepicker/src/react/datepicker.tsx
```diff
@@ -1,18 +1,36 @@
...
-import { useDateSelectChange } from './utils'
-import { ValueOf, HTMLPropsFor } from '@pluralsight/ps-design-system-util'
-import { slides } from '../vars'
+import React, {
+  FC,
+  RefAttributes,
+  ForwardRefExoticComponent,
+  ReactNode,
+  ComponentProps
+} from 'react'
```
danethurber, line 11: lots of these imports are not used

packages/datepicker/src/react/text-input-field.tsx
```diff
@@ -0,0 +1,58 @@
...
+    value,
+    ...rest
+  } = props
+  const labelId = useUniqueId('text-input__label-')
+  const inputId = useUniqueId('text-input__input-')
+
+  if (!prefix) {
+    prefix = <CalendarIcon onClick={onClick} style={{ cursor: 'pointer' }} />
+  }
+  if (typeof label === 'string') {
```
danethurber, line 30: should this be memoized like i've done in other components https://github.com/pluralsight/design-system/blob/master/packages/multiselect/src/react/index.tsx#L203

not sure but might want to look into it. i've memoized because of two issues i'd encountered
1. unnecessary renders
2. refs issues that i don't think would matter here

packages/datepicker/src/react/text-input-field.tsx
```diff
@@ -0,0 +1,58 @@
...
+    onChange,
+    onClick,
+    placeholder,
+    prefix,
+    subLabel,
+    value,
+    ...rest
+  } = props
+  const labelId = useUniqueId('text-input__label-')
+  const inputId = useUniqueId('text-input__input-')
```
danethurber, line 25: how are teams going to mock this out in snapshot tests? not something i expect you to address since we have this issue in other places - just mentioning

packages/datepicker/src/react/__stories__/index.story.tsx
```diff
@@ -141,7 +142,40 @@ storiesOf('SingleDate', module)
       </div>
     )
   })
-  .add('DatePicker', () => <DatePicker />)
+  .add('DatePicker', () => (
```
danethurber, line 145: nitpick - this should be it's own top level story category

Screen Shot 2021-03-04 at 2 42 31 PM

danethurber (changes requested):
menu doesn't close on blur, outside click or esc. is this a regression in the PR or an issue that should be tracked and fixed

Screen Shot 2021-03-04 at 2 42 31 PM

jaketrent:
@danethurber thanks for the review. Addressed each item. Commit messages should be indicative of where to look for changes.

Lacking a click outside of menu to close feature is not a regression. It never has been created. I carded it: https://github.com/pluralsight/design-system/issues/1655

Please weigh in on the card as to the "blur" functionality (that you mention in your review) that you think should be there.

Ready for re-review.

danethurber (approved):
LGTM 👍

Issue #1657, opened 2021-03-08
Create a11y test beds


danethurber:
Team attempted to audit a single component in isolation - didn't give great results. The new alternative plan is to build mock pages and audit pages as a whole to find issues. 

Create pages that use our DS components in a mock scenario. examples might be

- [ ] Nav/Appframe with navItems(label variations)
- [x] page with large form ([Figma](https://www.figma.com/file/JdOMBGN5JUtP7A6nOTxy4a/A11y-Test-Bed?node-id=1%3A21310))
- [x] page with sortable table
- [x] page with search results (Row/Carousel/Card)
- [ ] something with a steps and a dialog ([Figma](https://www.figma.com/file/JdOMBGN5JUtP7A6nOTxy4a/A11y-Test-Bed?node-id=4%3A10517))

PR #1659, opened 2021-03-08
a11y testbed

danethurber:
A set of basic page templates. we can go through the process of making these pages WCAG compliant as a team. Then we can use these pages going forward to perform ongoing a11y audits.

resolves: https://github.com/pluralsight/design-system/issues/1657


sweeting (approved):
Looks good, although I needed to add an .env file with `SKIP_PREFLIGHT_CHECK=true` in order to run `yarn start` successfully.

Issue #1652, opened 2021-03-04
[search input] button/icon not disappearing after clearing input


willsoto:
### Expected behavior

the `x` icon/button should disappear from the search input after being cleared or if there is nothing in the input to clear.

### Actual behavior

https://user-images.githubusercontent.com/2791730/109984245-92553980-7cd1-11eb-8fba-5c7e36257a07.mp4

### Steps to repro

1. Type something into the search input providing the `onClear` prop
2. Press `x` to clear input
3. `x` continues to be present after input is cleared

### Related packages

- search input@8.0.4

### Environment

- OS - MacOS 10.12
- Browser version - FF 86

PR #1661, opened 2021-03-08
docs: update onClear example for searchinput

jaketrent:
The SearchInput is controlled by a value prop. onClear internally clears
the input DOM node, but that doesn't re-render react. To get the
re-render, the usage pattern should be to clear the value prop.

It's possible that we could/should just remove the internal DOM clearing
too.

Resolves: #1652


danethurber (approved):
> It's possible that we could/should just remove the internal DOM clearing too.

yeah probably. if it's a controlled value then it should remain controlled all the way through

jaketrent:
@danethurber Would you mind reviewing a few more changes? Do you feel like this is a breaking change? I feel like the incoming value prop made this controlled before. But it had some internal state management (incomplete, as shown by the bug) before. Has this PR fundamentally changed its controlled nature or not, do you think?

jaketrent:
All right.  I'm gonna merge this given the previous approval.  I'll assume for now that Dane feels that the component was controlled before and still is. This is the pattern we prefer, and that's what it seems to have been.

PR #1666, opened 2021-03-12
The Great ESM Migration

jaketrent:
## Overture

Please see full commit messages for more context on the changes you see.

This is for ActionMenu only. Let's get one package right first. If you dig it, we'll expand to the rest of packagekind.

Once approved, I do not plan to merge but to do more commits and merge all the packages in one PR.

A couple other links re the file extensions (TS-specific) that might be helpful:

- https://github.com/microsoft/TypeScript/issues/16577#issuecomment-754941937
- https://github.com/microsoft/TypeScript/issues/16577#issuecomment-703190339


jaketrent:
Note: Build is broken. Note avatar build:css busted. This is because the change to build makes it esm compatible only. Once we update all the packages to be esm, we should be set.

packages/actionmenu/src/react/arrow.tsx
```diff
@@ -2,7 +2,7 @@ import { HTMLPropsFor } from '@pluralsight/ps-design-system-util'
 import { css } from 'glamor'
 import React from 'react'
 
-import stylesheet from '../css'
+import stylesheet from '../css/index.js'
```
EdwardIrby, line 5: @jaketrent did importing `.ts` break tree shaking?

jaketrent: No, I don't think so. But I'm not sure why you ask, so maybe there's something I'm missing. Is there something you know about that we should verify related to tree shaking?

This change was made to provide explicit file extensions, required by esm, that TS allows but will not rewrite for us in compilation. (see linked TS issues in the PR)

EdwardIrby: Didn't read those it's clear now. Thanks!

EdwardIrby: However it looks like that is what's causing the test failures.  🤷🏿‍♂️

danethurber (approved):
approach seems simple to me. the js on imports seems weird to me - but if that's the requirement that let's do it. i could have expected tsc to add the js extensions 🤷

jaketrent:
@danethurber Ditto, I would have thought tsc would do it. In the linked issues to the TS project, you can see the maintainers saying that rewriting the files like that is no in line with their project philosophy.

PR #1669, opened 2021-03-16
avatar: convert stories to CSF

danethurber:


packages/avatar/src/react/__stories__/index.story.tsx
```diff
@@ -1,107 +1,120 @@
...
+SizeSmall.args = { ...defaultArgs, size: Avatar.sizes.small }
+
+export const SizeMedium = Template.bind({})
+SizeMedium.args = { ...defaultArgs, size: Avatar.sizes.medium }
+
+export const SizeLarge = Template.bind({})
+SizeLarge.args = { ...defaultArgs, size: Avatar.sizes.large }
+
+export const SizeXLarge = Template.bind({})
+SizeXLarge.args = { ...defaultArgs, size: Avatar.sizes.xLarge }
```
jaketrent, line 114: Could convert size stories to a single story.

packages/avatar/src/react/__stories__/index.story.tsx
```diff
@@ -1,107 +1,120 @@
...
+NameUndefined.args = { ...defaultArgs, name: undefined }
+
+export const NameSingleChar = Template.bind({})
+NameSingleChar.args = { ...defaultArgs, name: 'x' }
+
+export const NameSpecialChar = Template.bind({})
+NameSpecialChar.args = { ...defaultArgs, name: '% Special Char' }
+
+export const VariousNames: Story = () => {
+  const names = useMemo(
```
jaketrent, line 37: Why is useMemo required static data like this? And why required in a story?

jaketrent (approved):
Story format looks great. 

As you likely agree, the effort to import stories into specs seems bespoke and will need to be remembered per component. This will require some discipline from all who do these upgrades.

Not sure if this is a regression: 
The "Image Error" story shows the alt text in large type (ie, "Jake Trent") in the too-small space before the image fails and the initials are finally shown.

packages/avatar/src/react/__stories__/index.story.tsx
danethurber: uh...no reason. just me doing weird things. i'll fix

PR #1671, opened 2021-03-16
ESM migrations (collaborative ver 2)

jaketrent:
Replacement for branch: https://github.com/pluralsight/design-system/pull/1666

## The conceptual approach:

refactor(build): update build-css to be esm to be able to build from esm

Used to build css from the cjs.  Now there is not going to be a cjs
build of packages. Only esm. So build-css has to be updated to be able
to import and trace through dependencies (at least in the css)

refactor(actionmenu): add extensions and make relative imports concrete

File extensions on relative imports is an esm requirement:

https://nodejs.org/api/esm.html#esm_terminology

This will be nice if we ever get to the point of using native browser
imports, which I hope we do.

This is also required for the build:css process once it points at the
esm output.

Related:
- https://github.com/microsoft/TypeScript/issues/16577#issuecomment-754941937
- https://github.com/microsoft/TypeScript/issues/16577#issuecomment-703190339

refactor(actionmenu): remove file path imports

These were initially added when tree shaking wasn't assumed. It allowed
importing only what is needed. Now tree shaking is the standard and
there is no CJS.

BREAKING CHANGE: Remove file path imports (eg, 'packagename/react')

refactor(actionmenu): drop cjs; set type module; sideEffects false

BREAKING CHANGE: Drop CJS build

## My process:

Commit msg:
```
refactor(): drop cjs; convert to esm only

BREAKING CHANGE: Remove file path imports (eg, 'packagename/react')
BREAKING CHANGE: Remove cjs build
```

Step 1: Search for relative imports/exports:

```
ag "(export|import).*\..*" src
```

And adjust manually.

Step 2: Remove file import files

```
rm css.js
rm js.js
rm react.js
rm vars.js
```

Step3: Adjust package.json:

``` 
+ "sideEffects": false,
+ "type": "module",

+    "build": "run-s build:esm build:css",
-    "build:cjs": "tsc --project tsconfig.build.json --module commonjs --target es5 --outDir dist/cjs",
+    "build:css": "build-css --useGlamor -i dist/esm/css/index.js",

+  "main": "dist/esm/index.js",
```

Step 4: Verify

```
rm -rf dist
yarn run build
```

And spot check

## Ranges

- Jake - working from top, to `icon`
- Dane - working from bottom, to `switch`


jaketrent:
pr replaced.

Issue #1670, opened 2021-03-16
Storybook CSF conversion and jest-axe tests


danethurber:
A coordinated effort to convert all stories to [CSF](https://storybook.js.org/docs/react/api/csf) and integrate [jest-axe](https://github.com/nickcolley/jest-axe) into the test suite for each component.

An example of the refactor can be found here https://github.com/pluralsight/design-system/pull/1674/commits/5ccdbaa8a3c08d60b1a269b88cb985d99f541189

---
Per component steps:

**1. Find a component from the list below**
Put your initials next to the component name signifying your working on it

**2. Convert stories**
For simple stories you can attempt to use the official storybook codemod like so:
`npx -p @storybook/cli@next sb migrate storiesof-to-csf --glob="**/*.story.tsx"`

But there is a good chance you'll need to do it manually. If manual refactor is needed, use your best judgement to organize and convert the stories.

**3:  Import stories into unit tests and use jest-axe to check for violations**
Some of existing tests might now be redundant with the new overlapping stories. Use your best judgment to provide clean and simple test cases.

If fixing the axe violations is straight-forward then please fix - if not then leave the failing tests to be fixed later.

**4: Convert tests**
Update existing tests to use stories for unit testing. Use your best judgement to determine if a story is appropriate for the test.

Test names should be concise and avoid using the word `should`. [explanation of change](https://github.com/pluralsight/design-system/pull/1674#discussion_r597782578)

**5. Open a pull request**
If tests are passing please open for review. If tests are failing, then open a draft pr and we can review as a team to prioritize necessary changes.

---

### Progress:

Please add your initials when you pickup a component to refactor. when you open a pull request please add a reference.

- [x] actionmenu
- [x] appframe (D) #1677
- [x] avatar (D) #1669
- [x] badge (B)
- [x] banner (B)
- [x] breadcrumb (B)
- [x] button (D) #1674 
- [x] card
- [x] carousel
- [x] checkbox #1688
- [x] circularprogress
- [x] datawell
- [x] datepicker
- [x] dialog
- [x] drawer
- [x] dropdown (D) #1688
- [x] emptystate 
- [x] errors
- [x] featureflags
- [x] field (D) #1688
- [x] focusmanager
- [x] form
- [x] halo (D) #1688
- [x] icon 
- [x] layout
- [x] linearprogress
- [x] link
- [x] menu
- [x] multiselect (D) #1692
- [x] navbar (D) #1677
- [x] navbrand (D) #1677
[...]

PR #1674, opened 2021-03-19
refactor(button): convert stories to csf. add jest-axe tests

danethurber:
related #1670

### DIFF
--- a/packages/button/src/react/__specs__/index.spec.tsx
+++ b/packages/button/src/react/__specs__/index.spec.tsx
@@ -1,80 +1,107 @@
-import { fireEvent, render } from '@testing-library/react'
+import { PlaceholderIcon } from '@pluralsight/ps-design-system-icon'
+import { convertStoriesToJestCases } from '@pluralsight/ps-design-system-util'
+import { fireEvent, screen, render } from '@testing-library/react'
+import { axe } from 'jest-axe'
 import React from 'react'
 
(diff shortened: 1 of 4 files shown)


packages/button/src/react/__specs__/index.spec.tsx
```diff
@@ -1,80 +1,107 @@
...
-    )
-    const el = getByTestId('undertest')
+  describe('.sizes', () => {
+    it('exists', () => {
+      expect(Button.sizes).toEqual(vars.sizes)
+    })
+  })
 
-    expect(el.tagName.toLowerCase()).toEqual('a')
+  it('should forward ref', () => {
```
jaketrent, line 21:
```suggestion
  it('forwards ref', () => {
```

If you're interested, while we're working through each of these specs, I fully welcome dropping the repeating "should" from descriptions.

jaketrent (approved):

packages/button/src/react/__specs__/index.spec.tsx
danethurber: yeah now it a great time. using should is just the convention i've typically followed - but happy to adopt a new standard. if you can summarize the approach i'll add it to the issue. maybe `use concise test names`?

jaketrent: Sure, I can add it to our style guide.  

Section 1.1.b
Be it known that the spec description shall not include the word "should". This is right.

PR #1675, opened 2021-03-19
[experimental] polymorphic forwardRef methods

danethurber:


packages/util/src/polymorphism.ts
```diff
@@ -0,0 +1,193 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+
+/**
+ * NOTE: Polymorphic approach inspired by(lifted from) reach-ui
```
jaketrent, line 4: Would love this in 1+n commits, if possible:

1 - The "lifted from" state
n - The modifications made to it

Would be great to see what had to change to fit our approach.

packages/util/src/__specs__/polymorphism.spec.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+const CompoundComp = forwardRefWithAsAndStatics<Props, 'div', Statics>(
+  (props, ref) => {
+    const { as: Comp = 'div', optional, ...rest } = props
+
+    return <Comp data-optional={optional} ref={ref} {...rest} />
+  }
+)
+CompoundComp.Anchor = Anchor
+CompoundComp.Button = Button
```
jaketrent, line 34: What is being proven by doing this?

packages/util/src/__specs__/polymorphism.spec.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+describe('polymorphism', () => {
+  describe('#forwardRefWithAs', () => {
+    it('renders without type errors', () => {
+      expect.assertions(0)
+
+      render(
+        <>
+          {/* Anchor accepts optional prop */}
```
jaketrent, line 43: Idea: Could write as separate `it('does a thing', () => {})` calls and avoid the comments.

packages/util/src/__specs__/polymorphism.spec.tsx
```diff
@@ -0,0 +1,107 @@
...
+          {/* Button as "a" accepts onClick prop */}
+          <Button
+            as="a"
+            onClick={(evt: MouseEvent<HTMLAnchorElement>) =>
+              evt.currentTarget.href
+            }
+          />
+
+          {/* Button as "Anchor" accepts href prop */}
+          <Button as={Anchor} href="#" />
```
jaketrent, line 78: That's pretty cool that the types allow it. Not sure when I'd use though.

packages/util/src/__specs__/polymorphism.spec.tsx
danethurber: that we can create polymorphic components that are also compound components. the new elements could replace the following code:

```
interface AvatarProps extends HTMLPropsFor<'div'> {
  ...
}

interface AvatarStatics {
  ...
}

type AvatarComponent = RefForwardingComponent<
  AvatarProps,
  HTMLDivElement,
  AvatarStatics
>

const Avatar = forwardRef<HTMLDivElement, AvatarProps>((props, ref) => {
  ...
}) as AvatarComponent
```

with this

```
interface AvatarProps extends HTMLPropsFor<'div'> {
  ...
}

interface AvatarStatics {
  ...
}

const Avatar = forwardRefWithAsAndStatics<AvatarProps, 'div', AvatarStatics>(
  (props, ref) => {
    ...
  }
)
```

The latter example requires less typing, no typecasting of the component, adds polymorphism

packages/util/src/polymorphism.ts
danethurber: commit history updated. all code is copied with the exception of the `forwardRefWithAsAndStatics` method with is an addition

packages/util/src/__specs__/polymorphism.spec.tsx
danethurber: i could also create a `forwardRefWithStatics` method if there is the need to create a compound component that doesn't support polymorphism. it could replace the `RefForwardingComponent` and would match this components in name and style

packages/util/src/__specs__/polymorphism.spec.tsx
danethurber: yeah i can look into that

packages/util/src/__specs__/polymorphism.spec.tsx
danethurber: maybe more of an advanced need. it's similar to our `render*` props where we allow passing in a ReactElement.

it also might help with composition using the various css-in-js libraries. here is the example i've tested with @emotion/styled - but i'm not sure if there is a better way to do this

```
export const Input = forwardRefWithAs<Props, "input">((props, ref) => {
  const { as: Comp = "input", children, ...rest } = props;

  return (
    <Comp ref={ref} {...rest}>
      {isFunction(children) ? children() : children}
    </Comp>
  );
});
```

```
import { Input } from 'some/ds/path'

const StyledInput = styled("input")`
  color: hotpink;
`;

export const CustomInput = (props) => <StyledInput as={Input} {...props} />;
```

there are some other ways to use emotion that i've tested here -> https://codesandbox.io/s/flamboyant-cloud-p7unf?file=/src/App.tsx

EdwardIrby:
This works for me better than typecasting. How's it working in your ide?

danethurber:
yep works well for me. i cleaned up the util methods and did some quick updating of related components to fix the build. there are some tests/stories that need fixed still. i'm going to leave this open while we get through our story(CSF) refactor. once we have all the stories converted we can double back and get this ready for merge/review

PR #1676, opened 2021-03-20
Css/poc

EdwardIrby:
# CSS/POC

### Migration strategy
- I used a new temporary export `tokens` in core to make migration easier
- Run a build after swapping tokens for named exports from core and then you can build a css file that uses the core css vars
-  CSS Pretty Print the the generated file
-  Refactor and use new classNames util
- Use new storybook [.main.js config](https://github.com/pluralsight/design-system/blob/39fdb574d936d530e40df53a842b3cd99ccbcb26/packages/link/.storybook/main.js)

### Generated dist output
- A postcss plugin generates a css file that then has the fallback values to support ie11
 Outstanding questions
- should this css approach happen before esm? 
- right now for this to work we need to generate the css file in the same relative location in the dist directory.


package.json
```diff
@@ -119,11 +119,13 @@
     "@types/vfile-message": "^2.0.0",
     "@typescript-eslint/eslint-plugin": "^4.9.0",
     "@typescript-eslint/parser": "^4.9.0",
+    "autoprefixer": "^9.8.6",
```
jaketrent, line 122: Dunno if it seems like a better option to you, but postcss-preset-env would bundle up some of these features into a set https://github.com/csstools/postcss-preset-env

packages/util/src/__specs__/css.spec.ts
```diff
@@ -0,0 +1,34 @@
...
+    .ExampleInner {
+      background-color: blue;
+      padding: 0;
+    }
+    [data-active='${true}'] {
+      color: red;
+      margin-left: auto;
+    }
+  `
+  console.log(actual)
```
jaketrent, line 32:
```suggestion
```

packages/banner/src/css/index.ts
```diff
@@ -47,17 +47,21 @@ export default {
       color: 'currentColor',
 
       '&:hover, &:active, &:focus': {
-        backgroundColor: transparentize(0.9, colorsBlack)
+        backgroundColor: `rgba(${parseToRgb(colorsBlack)}, 0.1}`
```
jaketrent, line 50:
```suggestion
        backgroundColor: `rgba(${parseToRgb(colorsBlack)}, 0.1)`
```

packages/banner/src/css/index.ts
```diff
@@ -47,17 +47,21 @@ export default {
...
     }
   },
 
   [`.psds-banner__button--color-${vars.colors.blue}`]: {},
   [`.psds-banner__button--color-${vars.colors.green}`]: {},
   [`.psds-banner__button--color-${vars.colors.yellow}`]: {
     ':not(#fake-id-to-force-higher-specificity)': {
-      '&:focus': { boxShadow: `0 0 0 3px ${transparentize(0.5, colorsBlack)}` }
+      '&:focus': {
+        boxShadow: `0 0 0 3px rgba(${parseToRgb(colorsBlack)}, 0.5`
```
jaketrent, line 63:
```suggestion
        boxShadow: `0 0 0 3px rgba(${parseToRgb(colorsBlack)}, 0.5)`
```

.eslintrc.js
```diff
@@ -3,114 +3,113 @@ const WARNING = 1
 const ERROR = 2
 
 module.exports = {
```
jaketrent, line 5: Will you please help me by highlighting what's different in this file? This diff is unhelpful.

packages/core/src/js/colors.ts
```diff
@@ -3,7 +3,6 @@ export const colorsWhite = '#FFFFFF'
 export const colorsBlack = '#000000'
 
 export const colorsBlue = {
-  base: '#0084BD',
```
jaketrent, line 6: Why the removal of the base colors?

packages/core/src/js/colors.ts
```diff
@@ -147,7 +138,7 @@ export const colorsBackgroundLight = {
 }
 
 export const colorsBackgroundUtility = {
-  base: '#8A99A8',
+  base: '138,153,168',
```
jaketrent, line 141: Breaking change for core.

I like the direction of the change. I would like to converse a bit more about timing. 

I would like to (1) focus this POC on component style overrides, (2) be comprehensive in a move toward rgb colors in core, and (3) come to an agreement on the best timing for an rgb move.

packages/core/src/tokens/index.ts
```diff
@@ -0,0 +1,35 @@
...
+  breakpoints,
+  layers,
+  layout,
+  type,
+  motion,
+  ...colors
+}).reduce((acc, [group, obj]) => {
+  const keys: Record<string, string> = {}
+  Object.keys(obj).forEach(key => {
+    keys[key] = `var(--ps${capitalizeFirstLetter(group)}${capitalizeFirstLetter(
```
jaketrent, line 30: To me, design tokens are usually abstracted from a tech implementation such as css custom props. Design tokens can be tricky enough to talk about and have mutual understanding about, that I am wary of stripping that distinction here.

packages/link/postcss.config.js
```diff
@@ -0,0 +1,16 @@
+module.exports = {
+  plugins: [
+    require('autoprefixer'),
+    require('postcss-nesting'),
+    require('postcss-custom-properties')({
+      importFrom: [
```
jaketrent, line 6: Is importFrom strictly necessary?  

I don't understand all the constraints of this solution, so I'm not sure. But conceptutally, I'm thinking it's not, based on:

1. Components all have psds-normalize as a peerDep
2. psds-normalize puts all core vars in :root

So, the vars should simply be usable (even without import), because they're in scope at runtime.

packages/link/preview.js
```diff
@@ -0,0 +1 @@
+// import '@pluralsight/ps-design-system-normalize'
```
jaketrent, line 1: Why remove from storybook env? I *think* it's because you include it via webpack now. Was the change necessary? If so, why?

packages/link/src/react/index.tsx
```diff
@@ -1,11 +1,13 @@
...
-import { RefForwardingComponent } from '@pluralsight/ps-design-system-util'
+import {
+  RefForwardingComponent,
+  classNames
+} from '@pluralsight/ps-design-system-util'
 
 import React, { HTMLAttributes } from 'react'
 
-import stylesheet from '../css'
+import '../css/index.css'
```
jaketrent, line 10: I want to hear more of your perspective in this POC of how the developer experience has changed for users of the DS.  

Specifically, a component used to be npm installable and then usable without further bundler configuration. Has that changed? If so, how?

packages/util/src/__specs__/__snapshots__/css.spec.ts.snap
```diff
@@ -0,0 +1,20 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`css() 1`] = `
```
jaketrent, line 3: Snapshot looks a little odd.  Is this supposed to be prettified or minified? It seems to be somewhere in the middle.

packages/util/src/__specs__/class-names.spec.ts
```diff
@@ -0,0 +1,11 @@
...
+  expect(classNames('class-1', 'class-2')).toBe('class-1 class-2')
+  const condtionTrue = true
+  const conditionFalse = false
+  const actual = classNames(
+    'class-1',
+    conditionFalse && 'class-2',
+    condtionTrue && 'class-3'
+  )
+  expect(actual).toBe('class-1 class-3')
+})
```
jaketrent, line 11: Opinion: Splitting a spec into different tests, with (commonly) one assertion each can be good for lots of things, including tdd flow, easier debugs on tests failures, better documentation in spec as to what is being asserted.

packages/util/src/__specs__/color.spec.ts
```diff
@@ -1,9 +1,9 @@
...
 
-describe('#transparentize', () => {
+describe('#parseToRgb', () => {
   test('happy path', () => {
-    expect(transparentize(0, '#000000')).toEqual('rgba(0,0,0,1)')
-    expect(transparentize(0.5, '#ff0000')).toEqual('rgba(255,0,0,0.5)')
-    expect(transparentize(0.2, '#ffffff')).toEqual('rgba(255,255,255,0.8)')
+    expect(parseToRgb('#000000')).toEqual('0,0,0')
+    expect(parseToRgb('#ff0000')).toEqual('255,0,0')
+    expect(parseToRgb('#ffffff')).toEqual('255,255,255')
```
jaketrent, line 7: Does this handle the common UPPER case hex codes as well?

packages/link/.storybook/main.js
```diff
@@ -1,8 +1,40 @@
...
   stories: ['../src/**/*.story.@(js|ts|tsx)'],
   typescript: {
     check: true,
     reactDocgen: 'react-docgen-typescript'
-  }
+  },
+  webpackFinal: (config) => {
+    config.module.rules = config.module.rules.filter(
+      f => f.test.toString() !== '/\\.css$/'
+    );
```
jaketrent, line 16: I dig the approach for .css rule replacement. Nice!

packages/core/package.json
```diff
@@ -9,10 +9,13 @@
   "main": "dist/cjs/index.js",
   "module": "dist/esm/index.js",
   "scripts": {
-    "build": "run-s build:cjs build:esm build:css",
+    "build": "run-s build:cjs build:esm build:css build:tokens",
     "build:cjs": "tsc --project tsconfig.build.json --module commonjs --target es5 --outDir dist/cjs",
     "build:css": "node ./build",
-    "build:esm": "tsc --project tsconfig.build.json --module es2015 --target es5 --outDir dist/esm"
+    "build:esm": "tsc --project tsconfig.build.json --module es2015 --target es5 --outDir dist/esm",
+    "build:tokens": "ts-node ./src/tokens/index.ts",
```
jaketrent, line 16: As far as I can tell, src/tokens/index.ts doesn't have any sideEffects. What is actually built here/there?

packages/button/src/css/index.ts
```diff
@@ -171,17 +171,21 @@ export default {
...
-      background: transparentize(0.85, colorsPrimaryAction.textDarkTheme),
+      background: `rgba(${parseToRgb(
+        colorsPrimaryAction.textDarkTheme
+      )}, 0.85)`,
       color: colorsBlue[4]
     },
     '&:not([disabled]):active': {
       border: `2px solid ${colorsBlue[7]}`,
-      background: transparentize(0.8, colorsPrimaryAction.textDarkTheme)
+      background: `rgba(${parseToRgb(colorsPrimaryAction.textDarkTheme)}, 0.2)`
```
jaketrent, line 188: Observation: There's a fair bit of transparentize -> parseToRgb conversion in this PR. It feels like a separate issue to me and a distraction from component style flexibility POC.

packages/button/postcss.config.js
```diff
@@ -0,0 +1,16 @@
+module.exports = {
```
jaketrent, line 1: If we're going to have postcss.config.js in every package now, I'd try to inherit from some base config, similar to our approach with jest.config.js

jaketrent (commented):
I left lots of inline comments, questions and suggestions.

In general, I like the approach because authoring is in css. I am concerned about the developer experience for users. 

Thanks for your work on this. I look forward to continuing the conversation.

packages/util/src/__specs__/css.spec.ts
EdwardIrby: I need to delete this test.

packages/core/src/js/colors.ts
EdwardIrby: Conversation with @sweeting. He uses the scale and references it in the design docs too.

packages/core/src/tokens/index.ts
EdwardIrby: This is temporary call it what you want. It's a utility for migration alone.

packages/core/src/js/colors.ts
EdwardIrby: Necessary for handling transparentize most common usage also actually reflects how this color se works.

packages/link/postcss.config.js
EdwardIrby: yep it's necessary to get those css vars during compilation.

packages/link/preview.js
EdwardIrby: need to undo

packages/util/src/__specs__/__snapshots__/css.spec.ts.snap
EdwardIrby: need to be deleted

packages/util/src/__specs__/class-names.spec.ts
EdwardIrby: Sure I copied some old code and a test I wrote 3 or so years ago over for this. I can clean it up.

packages/core/package.json
EdwardIrby: Artifact I need to remove from when I was trying something else

packages/button/src/css/index.ts
EdwardIrby: Probably Can make a separate issue with instructions.

PR #1677, opened 2021-03-22
convert nav component stories and add jest-axe tests

danethurber:
rel #1670


packages/navbrand/src/react/__specs__/index.spec.tsx
```diff
@@ -1,51 +1,72 @@
...
+
+      expect(el).toBeInTheDocument()
+      expect(el.tagName.toLowerCase()).toEqual('button')
+    })
+  })
+
+  describe('AsLink story', () => {
+    const { AsLink } = stories
+
+    it('renders a anchor el', () => {
```
jaketrent, line 64:
```suggestion
    it('renders an anchor el', () => {
```
nit

jaketrent (approved):

Issue #1680, opened 2021-03-23
ViewToggle: axe-core validation fails


danethurber:
getting an error that `aria-selected` is not a valid attribute on a `button` element. I believe we need to do some research on the appropriate pattern.

PR #1681, opened 2021-03-23
viewtoggle: convert stories to csf

danethurber:
`ViewToggle` fails the axe-core test. [ticket filed](https://github.com/pluralsight/design-system/issues/1680)


jaketrent (approved):
Looks great. 

I think adding more detail to the axe-core tickets would be good.  That is, if we have the feedback from axe-core. I'm not sure we do. I haven't had any failures yet. I think the detail would help me know what I'm getting into when I pick up the ticket. It would also be great to know why it wasn't immediately fixed: what was the judgement call that led to this being put off?

PR #1688, opened 2021-03-24
Convert form components to CSV part 1

danethurber:
there was more complexity than expected so i've split the form csf conversions into two tasks. this is part 1

- [x] checkbox
- [x] dropdown
- [x] field
- [x] halo


packages/dropdown/src/react/__stories__/index.story.tsx
```diff
@@ -1,834 +1,285 @@
...
-          Two
-        </Dropdown.Item>,
-        <Dropdown.Item key="3" value="h">
-          Three
-        </Dropdown.Item>
-      ]}
-    />
-  ))
+export default {
+  title: 'Components/Dropdown',
```
danethurber, line 54: for the stories: i noticed some overlap on what was being validated. I attempted to simplify and also broke the file up into 3

jaketrent (approved):

PR #1690, opened 2021-03-25
util: polymorphic primatives

danethurber:
new util methods to support typed(typescript) forwardRef components that support static properties, polymorphism and memoization.

for an example of how the methods would be used please see #1675 which shows refactoring the Button component and related components. Button is by far the most complicated and widely used component that supports the old polymorphic style.


packages/util/src/primatives.ts
```diff
@@ -0,0 +1,208 @@
...
+export interface MemoExoticComponentWithAs<
+  DefaultComponentType extends As,
+  ComponentProps
+> extends NamedExoticComponentWithAs<DefaultComponentType, ComponentProps> {
+  readonly type: DefaultComponentType extends ComponentType
+    ? DefaultComponentType
+    : FunctionComponentWithAs<DefaultComponentType, ComponentProps>
+}
+
+export interface ForwardRefWithAsRenderFunction<
```
danethurber, line 129: oops. forgot to document using a function render props. i'll add an example - the approach is pretty straightforward

interface Props {
  children?: React.ReactNode | (() => JSX.Element);
}
const TestInput = forwardRefWithAs<Props, "input">((props, ref) => {
  const { as: Comp = "input", children, ...rest } = props;

  return (
    <Comp ref={ref} {...rest}>
      {isFunction(children) ? children() : children}
    </Comp>
  );
});
```

**Comment by jhewlett** [comment]
@danethurber So the idea here is that by asking the consumer to specify the underlying DOM element to be used, you're able to provide better autocomplete and static guarantees? (e.g. if they specify `as="button"`, then `href` wouldn't be accepted as a valid prop?)

File: docs/src/content/patterns/polymorphic.mdx [hunk 2]
```diff
@@ -1,27 +1,179 @@
...
+    <StyledButton as="span" />
+  </>
+)
+```
+
+## Creating a polymorphic component
+
+The `utils` package of the design system exports a genericized higher order
+function that you can use to replace the native `React.forwardRef`. Using the
+`forwardRefWithAs` utility with provide you with greater type intropection of
```
KaidenR, line 125:
```suggestion
`forwardRefWithAs` utility will provide you with greater type introspection of
```

docs/src/content/patterns/polymorphic.mdx
```diff
@@ -1,27 +1,179 @@
...
+
+const Button = forwardRefWithAs<Props, 'button'>((props, ref) => {
+  const { as: Comp = 'button', ...rest } = props
+  return <Comp ref={ref} {...rest} />
+})
+```
+
+If you need a component that also support static properties (like for instance
+you are using the compound components pattern) we provide a higher order
+function to create a typesafe components with typed static properties.
```
KaidenR, line 140:
```suggestion
function to create typesafe components with typed static properties.
```

docs/src/content/patterns/polymorphic.mdx
```diff
@@ -1,27 +1,179 @@
...
+
+> Official [emotion polymorphic docs](https://emotion.sh/docs/styled#as-prop)
+
+```typescript noRender startExpanded
+import emotion from "@emotion/styled";
+import { PolymorphicButton } from 'some/other/psds/component/path'
+
+/* NOTE: I've had some issues with the emotion typings to fully support our
+ *       polymorphic `as` typings. Below is the approach I've found works the
+ *       best and avoides type errors - although it's not as clean or as helpful
```
KaidenR, line 97:
```suggestion
 *       best and avoids type errors - although it's not as clean or as helpful
```

docs/src/content/patterns/polymorphic.mdx
```diff
@@ -1,27 +1,179 @@
...
+ *       with it's introspections as the styled-component examples. I'm not as
```
KaidenR, line 98:
```suggestion
 *       with its introspections as the styled-component examples. I'm not as
```

KaidenR (commented):
This looks like a really great pattern that should help a LOT with customization. 

Is the idea that a component could have multiple `<subcomponent>As` props so that I can override different parts of the component? Like when rendering a Dropdown component I could have `buttonAs` and a `menuAs` props to customize the rendering of either of them. 

(Also submitting fixes for a few typos)

danethurber:
@justin yes the types should be able to tell which attrs/props are valid for the given tag name. the component's tag name also needs to have a default value. so for example the following component defaults to a `button` but can be changed as needed.

`const Button = forwardRefWithAs<Props, 'button'>(() => ...)`

will be typed to only allow valid `button` props, including using `HTMLButtonElement` for the `ref` and event types. but if you used `as` like so

`<Button as="a" />`

the typings would now related to a `HTMLAnchorElement` so attributes like `href` would be valid

---

@KaidenR i don't have a standard answer for how we get to the internal elements yet - but your approach could definitely be something we finalize on. we might end up with a collection of compound components meant to be hyper flexible and each should using the `as` prop when needs. And we will most likely have a set of components that are convenience wrappers and provide a simpler interface to encapsulate common patterns. Some current examples being our `Field` components vs a convenience wrapper like the `MultiSelect` component.

With the addition of these typings, our component compounds would allow changing the underlying dom elements at whatever level you need. an example being.

```
const Button = forwardRefWithAs<Props, 'button'>(() => ...)
interface Statics {
  Button: typeof Button
}
const Compound = forwardRefWithAsAndStatics<unknown, 'div', Statics>(() ...)
Compound.Button = Button

render(
  <Compound as="section">
    <Button  />
    <Button as="a" />
  </Compound>
)
```

EdwardIrby (approved):

PR #1691, opened 2021-03-25
fix(select): dropdown zindex added

EdwardIrby:


KaidenR:
This could cause an issue if a dialog needs to show while a Select menu is open on the page underneath it. But hopefully menus will always be closed when dialogs open so maybe it's okay.

KaidenR (approved):

PR #1696, opened 2021-03-29
flexibility: remove base from color scale & add parseToRgb

### LINKED ISSUE #1684: flexibility: update color vars
breaking to change
- remove base from color scales
- change colorsBackgroundUtility.base to '138,153,168'

EdwardIrby:
Removes base from color scale based on conversation with @sweeting 
Adds new `parseToRgb()` to make refactoring transparentize usage easier.
resolves #1684


packages/core/src/js/colors.ts
```diff
@@ -147,7 +138,7 @@ export const colorsBackgroundLight = {
 }
 
 export const colorsBackgroundUtility = {
-  base: '#8A99A8',
+  base: '138,153,168',
```
jaketrent, line 141: Having some values rgba()-wrapped and some with raw csv will lead to unpredictable usage of these values. I would rather do them all at once and cut over. It's already a breaking change. 

What do you think?

EdwardIrby: This is the only color we really need these raw values for.  because of how often we combine it with transparentize. The other colors are more one offs.

jaketrent (approved):

Issue #1683, opened 2021-03-23
[searchinput]: typescript requires a css prop


willsoto:
### Expected behavior

`<SearchInput>` can be used as [documented](https://design-system.pluralsight.com/components/searchinput#search-input) (no `css` prop required)

### Actual behavior

`<SearchInput>` requires (according to TypeScript) a `css` prop. See [example codesandbox](https://codesandbox.io/s/72l9k?module=/example.tsx&file=/example.tsx) from the docs.

### Steps to repro

1. https://codesandbox.io/s/72l9k?module=/example.tsx&file=/example.tsx

### Related packages

- searchinput@8.0.7
_Noting these packages since these upgrades fix the `enterKeyHint` error_
- @types/react@17.0.3
- @types/react-dom@17.0.3

### Environment

- OS - MacOS 10.12
- Browser version - N/A

willsoto:
After more debugging, this seems to be an issue (I think) between the design packages and their interaction with `@types/react@17.x.x` and/or `@types/react-dom@17.x.x`. I get this same TypeScript error for `"@pluralsight/ps-design-system-drawer": "~10.0.5"` and `"@pluralsight/ps-design-system-table": "~13.1.1"`

PR #1698, opened 2021-03-31
fix(searchinput): weird type bug resolved

EdwardIrby:
That was a weird one. I ended up playing with it until I got the build to output something more inline with that TextInput was outputting as it was doing a weird extension of it.  Should resolve the issue I hope. 

Resolves: #1683


jaketrent (approved):

packages/searchinput/src/react/index.tsx
```diff
@@ -18,14 +20,13 @@ const styles = {
   field: () => css(stylesheet['.psds-searchinput-field'])
 }
 
-export interface SearchInputProps
-  extends React.ComponentProps<typeof TextInput> {
+export interface SearchInputProps extends TextInputProps {
```
EdwardIrby, line 23: @jaketrent this the example I'm talking about

PR #1704, opened 2021-04-06
tagsinput: match height of buttons & text inputs

KaidenR:
### What You're Solving

- Redo some layout to make tagsinput consistent with buttons and text inputs at 40px tall
- Fixes text input box being too tall which was adding extra space above the last row of tags
- Fixes placeholder being cut off because it wasn't flex-growing
- Adds "Empty" and "Multiple lines" stories

### Design Decisions

- The gutter around the edges of the box is 8px and the spacing between tags is 4px

### How to Verify

- Verify the tagsinput component is 40px tall
- Verify scrolling a multi-line tagsinput looks correct (scroll bar in the right place, spacing around the edge of the box is correct, etc...)
- Verify a longer placeholder text no longer gets cut off
- Verify the space above the third row of tags is not larger than above the second row

### DIFF
--- a/packages/tagsinput/src/css/index.ts
+++ b/packages/tagsinput/src/css/index.ts
@@ -1,6 +1,7 @@
 import { layout } from '@pluralsight/ps-design-system-core'
 
-const PILL_GUTTER_SIZE = 4
+const GUTTER_SIZE = layout.spacingXSmall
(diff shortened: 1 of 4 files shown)


danethurber:
👍  LGTM. we can probably make the same changes to MultiSelect - i can do that though

danethurber (approved):

PR #1705, opened 2021-04-07
Convert form components to CSV part 4

danethurber:
ref #1670 

addressed the following components

- [x] Radio
- [x] Select
- [x] Switch


jaketrent (approved):
Looks like the red build is Switch snapshots.

PR #1712, opened 2021-04-15
expose package.json in conditional exports

jaketrent:
Conditional exports whitelist places in a package that can be imported from. It makes entry points explicit.

We want to make it the de facto standard to import the package name and avoid pathing files beneath the package.

But Storybook apparently pulls in package.json, so this is enabling that.
https://github.com/storybookjs/storybook/blob/32c63568d6f7f6ef90afd206405524543d5a7068/lib/core-server/src/manager/manager-config.ts#L18

Fixes this error:
https://github.com/storybookjs/storybook/issues/13099#issuecomment-805233273

### DIFF
--- a/packages/avatar/package.json
+++ b/packages/avatar/package.json
@@ -21,10 +21,13 @@
     "test:watch": "yarn test --watchAll"
   },
   "exports": {
-    "import": "./dist/esm/index.mjs",
-    "require": "./dist/cjs/index.js",
-    "node": "./dist/cjs/index.js",
-    "default": "./dist/cjs/index.js"
+    ".": {
+      "import": "./dist/esm/index.mjs",
+      "require": "./dist/cjs/index.js",
+      "node": "./dist/cjs/index.js",
+      "default": "./dist/cjs/index.js"
+    },
+    "./package.json": "./package.json"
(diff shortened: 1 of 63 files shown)


packages/actionmenu/package.json
```diff
@@ -21,10 +21,13 @@
     "test:watch": "yarn test --watchAll"
   },
   "exports": {
-    "import": "./dist/esm/index.mjs",
-    "require": "./dist/cjs/index.js",
-    "node": "./dist/cjs/index.js",
-    "default": "./dist/cjs/index.js"
+    ".": {
```
danethurber, line 24: what is this witchcraft

danethurber (approved):

packages/actionmenu/package.json
jaketrent: It is a path. It means, when importing, "the package name, with no paths beyond it". So, this case:

`import Button from '@pluralsight/ps-design-system-button'

https://nodejs.org/api/packages.html#packages_subpath_exports

If you don't have multiple paths, you can assume the default "no path" path, which is what we had. 

In other words, yes, witchcraft.

PR #1714, opened 2021-04-15
add style condition on exports and add missing style and types fields

jaketrent:
A couple missing `style` and `types` fields in package.json.

Added `exports.style` condition to all with styles. Webpack5 + postcss-loader can use this when `@import`ing or `<link href="">`ing.


danethurber (approved):
👍  once you fix the ordering

PR #1723, opened 2021-04-22
fix: makes storybook theme addon work

EdwardIrby:
This fixes it and doesn't effect bundling of these two packages, that still tree shake. Checked their sizes using https://github.com/AdrieanKhisbe/bundle-phobia-cli and it didn't change.

- Removed module field from package.json in addon because storybook is all *.cjs
- Removed exports field from package.json. from both to make all test and builds works
- Removed rewrite to *.mjs as well since it breaks theme addon in storybook. Further this field is completely unnecessary for any modern bundler which is what this code is actually used with.


danethurber (commented):
i don't know enough about the extensions and which is proper. I do think this looks simpler and if it's the right approach we should do it across all packages. let's wait for @jaketrent to review on monday since he has much more context around this than I

jaketrent:
Resolves: #1729

jaketrent (approved):

PR #1724, opened 2021-04-22
fix(viewtoggle): update aria attribute

### LINKED ISSUE #1680: ViewToggle: axe-core validation fails

EdwardIrby:
resolves #1680


danethurber:
https://github.com/pluralsight/design-system/blob/master/packages/viewtoggle/src/react/__specs__/index.spec.tsx#L28

^ please update the jest test to run the axe tests not that it should be passing

danethurber (approved):

PR #1726, opened 2021-04-23
Typeahead fixes and improvements

### LINKED ISSUE #1725: Typeahead 10.0.0.: bugs
- OnChange is not firing all the time
- Need to fix filter function or add way to pass in alternative one
-  empty list appearing according to Jim Cullison  when clicking in for first time

EdwardIrby:
resolves #1725


packages/typeahead/src/react/index.tsx
```diff
@@ -65,6 +75,20 @@ interface TypeaheadFieldStatics {
...
+    label: React.ReactText
+    value: React.ReactText
+  }[],
+  inputValue?: string
+) =>
+  options.filter(({ label }: { label: React.ReactText }) =>
+    `${label}`
+      .toLowerCase()
+      // eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion
+      .includes((inputValue as string).toLowerCase())
```
jaketrent, line 89: Function param is nullable string.  Should probably check for null before operating on it and avoid the cast.

packages/typeahead/src/react/index.tsx
```diff
@@ -135,16 +160,19 @@ const Typeahead: TypeaheadFieldComponent = props => {
...
-        )
-      )
+      setInputItems(filterFunction(options, inputValue))
     }
   })
+  React.useEffect(() => {
+    onChange &&
+      onChange(
+        null,
+        activeItem as {
```
jaketrent, line 170: Does downshift have a generically-typed `useCombobox` by chance?  That would probably enable the avoidance of the cast here.

If not, or perhaps in addition, I might create an interface type like `DropdownItemProps` (or whatever might seem a more appropriate name) so that this shape doesn't need redefined multiple times and that the relationship in code will be more apparent.

jaketrent (commented):
Went through code. Have a few questions:

- The bug, #1725, says that the filter function needs parameterized "or fixed". I don't understand that option. Would you mind explaining the problem and the approach you took related to that option, please? 
- How do I test this to see if the bug is fixed?  How did you test it?  Perhaps the addition of an automated test would be best.

packages/typeahead/src/react/index.tsx
EdwardIrby: selectedItem renamed activeItem can actually be what you want. As for a reusable interface sure I could do that will still require cast.

packages/typeahead/src/react/index.tsx
```diff
@@ -135,16 +160,12 @@ const Typeahead: TypeaheadFieldComponent = props => {
...
-            .toLowerCase()
-            // eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion
-            .startsWith((inputValue as string).toLowerCase())
-        )
-      )
+      setInputItems(filterFunction(options, inputValue))
     }
   })
+  React.useEffect(() => {
+    onChange && onChange(null, activeItem || undefined)
```
jaketrent, line 167:
```suggestion
    onChange && onChange(null, activeItem)
```

`activeItem` can already be undefined and passed as such. Is there another reason to add the `|| undefined`?

jaketrent (commented):
Nice change to remove the cast, thanks.

I still have a couple questions:

- The bug, #1725, says that the filter function needs parameterized "or fixed". I don't understand that option. Would you mind explaining the problem and the approach you took related to that option, please?
- How do I test this to see if the bug is fixed? How did you test it? Perhaps the addition of an automated test would be best.

PR #1727, opened 2021-04-26
Text: storybook CSF

danethurber:
ref: #1670


packages/text/src/react/__specs__/code.spec.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+import * as stories from '../__stories__/code.story'
+
+describe('Text', () => {
+  describe('Code', () => {
+    const cases = convertStoriesToJestCases(stories)
+
+    it.todo('forwards the ref')
+
+    it.todo('supports polymorphism')
```
jaketrent, line 14: why add the todos?

jaketrent (approved):

packages/text/src/react/__specs__/code.spec.tsx
danethurber: because i feel like these components should support these features. i'm sure they will come up in the flexibility theme work that is planned

PR #1732, opened 2021-04-28
fix(icon): bin path corrected

EdwardIrby:


jaketrent:
Please do a review next time before merging master.  It looks like it didn't happen here, but maybe it did.

Issue #1744, opened 2021-05-11
actionmenu: menu closes when scrolling inside menu


jaketrent:
Same requirements as #1741

PR #1748, opened 2021-05-12
Fix ActionMenu scroll event w/in menu

jaketrent:
Does this seem like the right general solution? 

I intend the code to be: Use the hook to close the menu when events are triggered *outside* of the menu. Resize will always be. Click might be in or out. We have had a check for if the click is within the menu. We haven't had that for scroll. Seems like we need to do the same check.

If this seems like the right way to go, I think I'd like to revert the recent patch to Dropdown to rely on this, if possible.

Resolves: #1744


packages/util/src/use-close-on-document-events.ts
```diff
@@ -9,8 +9,23 @@ export const onGlobalEventsClose = <El extends HTMLElement>(
...
+      evt.target instanceof HTMLElement ||
+      evt.target instanceof HTMLDocument
+    ) {
+      if (el.contains(evt.target)) return
+      currentAnimationFrame = window.requestAnimationFrame(() => callback(evt))
+    }
+    return currentAnimationFrame
+  }
+
   const handleClickOutsideMenu = (evt: MouseEvent) => {
```
EdwardIrby, line 24: evt: Event | UIEvent

EdwardIrby (changes requested):
Please look at Select stories. It seems like the first three stories are now broken. I'd prefer maybe just copying this into dropdown and using it separate as this effects both Menu and Select components.

packages/util/src/use-close-on-document-events.ts
```diff
@@ -9,8 +9,23 @@ export const onGlobalEventsClose = <El extends HTMLElement>(
 ) => {
   if (!canUseDOM()) return noop
 
+  const handleScrollOutsideMenu = (evt: MouseEvent) => {
```
EdwardIrby, line 12: evt: Event | UIEvent

jaketrent:
@EdwardIrby Good call on the Select story effects. I didn't catch that.

In this iteration, I felt like we should keep these event handler hooks centralized in util because of the vast reusability. But that breaking them up into separate events and  callbacks would allow for more flexibility in their application.

Please have another review.

EdwardIrby (approved):

PR #1757, opened 2021-05-13
fix(docs): remove extra parens in @media rules

### LINKED ISSUE #1737: docs: mobile styles are being applied even on large screen widths.
Them menu is hidden behind the hamburger button and the ToC on component pages are inline instead of on the right.

CleanShot 2021-05-03 at 09 59 00

sweeting:
Removing the extra set of parens in the `@media` CSS rules fixed the issues with the docs always showing the mobile layout at all screen widths.

Closes #1737

### DIFF
--- a/docs/src/components/frame/aside.module.css
+++ b/docs/src/components/frame/aside.module.css
@@ -26,7 +26,7 @@
     width: 100%;
   }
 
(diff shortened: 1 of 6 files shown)


jaketrent (approved):
Genius!  Great find.

PR #1760, opened 2021-05-21
fix: enter selects and closes menu

EdwardIrby:
Forgot about enter doing select. Added test to cover that
https://pluralsight.slack.com/archives/C70HPSJPP/p1621604860015000?thread_ts=1621363066.003300&cid=C70HPSJPP

### DIFF
--- a/packages/dropdown/src/react/__specs__/index.spec.tsx
+++ b/packages/dropdown/src/react/__specs__/index.spec.tsx
@@ -123,7 +123,7 @@ describe('Dropdown', () => {
       expect(onChange).toHaveBeenCalledWith(expect.anything(), 1)
     })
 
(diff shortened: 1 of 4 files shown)


packages/dropdown/src/react/__specs__/index.spec.tsx
```diff
@@ -132,12 +132,13 @@ describe('Dropdown', () => {
...
       btn.focus()
       pressEnter()
-
+      const menu = await screen.findByRole('listbox')
       pressArrowDown()
       pressArrowDown()
       pressEnter()
 
       expect(onChange).toHaveBeenCalledWith(expect.anything(), 3)
+      expect(menu).not.toBeInTheDocument()
```
jaketrent, line 141: good call!

jaketrent (approved):

Issue #1762, opened 2021-05-24
Select: First item is not being focused when opened


EdwardIrby:
When using `renderOption` or `children` the first option is not being focused when the Select opens.

jaketrent:
Hey @EdwardIrby , when I exercise the code, I get the proper, expected results.

This [existing story, CustomRenderOption](https://github.com/pluralsight/design-system/blob/1752eb304beb22762fdf00739f39ae84fb1030a0/packages/select/src/react/__stories__/index.story.tsx#L131-L144) uses `renderOption`.

When I click the select element to open it, it opens, then I go to console and detect the focus and get the first `li`, as expected:

<img width="1120" alt="Screen Shot 2021-06-17 at 10 11 47 AM" src="https://user-images.githubusercontent.com/552044/122434574-7447d800-cf54-11eb-8cae-77fa9f3ba1b7.png">

Please help me determine if this is still a bug. Does this story contain the causal conditions for the bug? Do you have a different outcome using this story?

jaketrent:
closing as non issue. please reopen if more evidence is uncovered for a bug.

PR #1763, opened 2021-05-24
Fix/focus trap dropdown

EdwardIrby:
So there were two focus traps and one was the result of tabbing pass the select where it would open the dropdown unintentionally. Our original conditional somehow let Tab through. I'm not sure why but I modified it to use an Array.includes which solved the problem. 

As for Select I updated the stories to include the basic example that got removed on CSF conversion in addition to the MenuItemWithDescription ones. In doing so I noticed an unrelated bug I'm going to create an issue for #1762

### DIFF
--- a/packages/dropdown/src/js/index.ts
+++ b/packages/dropdown/src/js/index.ts
@@ -143,13 +143,11 @@ export const useDropdown = (
       }
     }
   }, [itemMatchingValueIndex, isOpen])
-
+  const openKeyEvents = new Set(['Enter', 'ArrowDown', 'ArrowUp', ' '])
   function handleButtonEvent(evt: React.MouseEvent | React.KeyboardEvent) {
     if (
       evt.type === 'click' ||
-      (evt.type === 'keydown' &&
-        'key' in evt &&
-        ['Enter', 'ArrowDown', 'ArrowUp', ' '].includes(evt.key))
+      (evt.type === 'keydown' && 'key' in evt && openKeyEvents.has(evt.key))
     ) {
       evt.preventDefault()
       evt.stopPropagation()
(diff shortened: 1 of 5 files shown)


packages/select/src/react/index.tsx
```diff
@@ -43,39 +43,40 @@ interface SelectProps extends UseListboxProps {
...
-      position = 'belowLeft',
-      renderOption = defaultRenderOption,
-      children,
-      ...rest
-    } = props
-    const { buttonProps, selectedProps, menuProps, isOpen } = useListbox(
-      rest,
-      ref
-    )
+const Select = forwardRefWithStatics<
```
jaketrent, line 46: Double-checking: Do we not need to support the `as` prop?

packages/dropdown/src/js/index.ts
```diff
@@ -149,7 +149,7 @@ export const useDropdown = (
       evt.type === 'click' ||
       (evt.type === 'keydown' &&
         'key' in evt &&
-        (evt.key === 'Enter' || 'ArrowDown' || 'ArrowUp'))
+        ['Enter', 'ArrowDown', 'ArrowUp', ' '].includes(evt.key))
```
jaketrent, line 152:
```suggestion
        (evt.key === 'Enter' || evt.key === 'ArrowDown' || evt.key === 'ArrowUp'))
```

This would make the old `||` combination work.  I like the array approach too.

https://caniuse.com/array-includes doesn't work in IE. Could use `indexOf` instead.

jaketrent (approved):

packages/select/src/react/index.tsx
EdwardIrby: Not on Select. Only the Menu.Item needs that.

Issue #1600, opened 2021-02-08
vertical tabs: tabs not updating with dynamic collapsed prop


willsoto:
### Expected behavior

Collapsed state is toggled when the `collapsed` prop is dynamic

### Actual behavior

The collapsed state is stuck on whatever initial state was provided

### Steps to repro

https://codesandbox.io/s/codesandboxer-example-forked-t4wu7?file=/example.tsx (check console to confirm collapsed state is being toggled)

### Related packages

- @pluralsight/ps-design-system-verticaltabs: "~5.0.2"

### Environment

- OS - MacOS 10.12
- Browser version - FF85

jaketrent:
Hey @willsoto ,

The `collapsed` prop is currently meant to "start item collapsed if collapsible also true".

https://design-system.pluralsight.com/components/verticaltabs/#verticaltabstier1

It looks like in your codesandbox, you're trying to programmatically collapse the section, right?  What's your real-world use case for programmatic collapse and open?

willsoto:
We have tabs that start closed until the user is able to access them (via progressing through the project). Only once they have progressed are they allowed to see the other sections.

jaketrent:
Thanks for the use case.  Totally makes sense to me.

Team, cool w/ making this possible?

There was a feeling in planning mtg from the team that we might want to revamp the props on this component.  I think that's fine to do if we have a vision for how it might be reorganized.

Or, I think we could simply add a prop to handle this. I suppose it's an addition instead of an adjustment to `collapsed` because it's changing how the prop works. This, even if `collapsed` as a name might be somewhat mismatched from its functionality.

Team, opinions? cc @danethurber , @EdwardIrby

sara-berrios:
Hello!

@BenGu3 and I were hoping to get this looked at again. We ran into a similar issue as @willsoto and saw he had a solution posted on this [slack thread](https://pluralsight.slack.com/archives/C70HPSJPP/p1612537180052800). 

@jaketrent How would you feel about us submitting a PR that adds the following code after [this line](https://github.com/pluralsight/design-system/blob/master/packages/verticaltabs/src/react/item.tsx#L138)?
``` 
useEffect(() => {
  setCollapsed(initialCollapsed)
}, [initialCollapsed])
```

jaketrent:
@sara-berrios I'd be happy to have a PR that enables this use case. Please make sure to add tests to verify it, and we'll be off to the races.

PR #1764, opened 2021-05-24
verticaltabs: make collapse prop dynamic

osha90:
### What You're Solving

- https://github.com/pluralsight/design-system/issues/1600


packages/verticaltabs/src/react/item.tsx
```diff
@@ -136,6 +136,10 @@ const Item = React.forwardRef<HTMLLIElement, ItemProps>((props, ref) => {
   } = props
 
   const [collapsed, setCollapsed] = React.useState<boolean>(initialCollapsed)
+  useEffect(() => {
+    setCollapsed(initialCollapsed)
+  }, [initialCollapsed])
+
```
EdwardIrby, line 142: We need a test for this

EdwardIrby (changes requested):

PR #1768, opened 2021-05-26
Singleton

EdwardIrby:
Single Package


jaketrent:
@EdwardIrby, when you get back, let's chat about a possible `2b` approach. (Essentially the one that you suggested before and where I see some new potential pro's.)

```
1. Today

- Mono repo
  - packages/
    - button/
      - package.json
      - src/
    - dropdown/
      - package.json
      - src/

2a. Combining Code

- Single repo, not mono
  - package.json - THE single package
  - src/
    - button/
    - dropdown/

2b. Releasing a single package

- Con:
  - Tooling overhead - mono repo
  - 2 release schedules

- Pro:
  - Easy rollback the idea of singleton package - "doesn't solve the stability problem"
  - 2 release schedules - some teams prefer one or the other
  - Single "latest" npm channel - use single packages for "next" line
    - deps:
      - singleton@1
      - button@25

- Mono repo
  - packages/
    - singleton/ - release 6 months, stable - upgrade culture, event, narrative on improvements
      - package.json
        - button@24 - ^take bug fixes, maybe feature adds?
        - dropdown@12
    - button/ - release whenever
      - package.json
      - src/
    - dropdown/
      - package.json
      - src/
```

components/css/index.css
```diff
@@ -0,0 +1,149 @@
+@import 'normalize.css';
```
jaketrent, line 1: Why is this moved out of normalize?

components/package.json
```diff
@@ -0,0 +1,96 @@
...
+    "@pluralsight/ps-design-system-textinput": "*",
+    "@pluralsight/ps-design-system-theme": "*",
+    "@pluralsight/ps-design-system-tooltip": "*",
+    "@pluralsight/ps-design-system-typeahead": "*",
+    "@pluralsight/ps-design-system-verticaltabs": "*",
+    "@pluralsight/ps-design-system-viewtoggle": "*"
+  },
+  "devDependencies": {
+    "cssnano": "^5.0.5",
+    "esbuild": "^0.12.7",
```
jaketrent, line 91: is this used?

packages/avatar/src/react/index.tsx
```diff
@@ -44,60 +44,61 @@ const styles: { [key: string]: StyleFn } = {
...
-
-  const handleImageLoadSuccess: React.ReactEventHandler<HTMLImageElement> = evt => {
-    const pixel = evt.target as HTMLImageElement
-    const isFallbackPixel = pixel.naturalWidth === 1
-    setImageState(isFallbackPixel ? 'error' : 'success')
-  }
-
-  const handleImageLoadError: React.ReactEventHandler<HTMLImageElement> = () => {
-    setImageState('error')
+export const Avatar = React.forwardRef<HTMLDivElement, AvatarProps>(
```
jaketrent, line 47: Is there anything of substance in this file diff?

components/src/index.ts
```diff
@@ -0,0 +1,110 @@
+export { AppFrame } from '@pluralsight/ps-design-system-appframe'
```
jaketrent, line 1: We are missing some of our exports.

If you look in most `packages/**/src/index.ts` files, there are exports for `css`, `vars` and `js`. I want users of the stable grouped package or the latest individual packages to get the same functionality. I think we can find a way to do that.

I think we can drop `css` because we're transitioning to vanilla css.

But I think `vars` and `js` should get exported too.  

0. What do you think: should we export those? 
1. What is the naming convention to avoid collision in this single namespace? Not sure yet.
2. Is that this PR? Probably not, because I don't have an answer to number 1. If not this PR, would you mind carding it?

packages/field/src/react/index.ts
```diff
@@ -1,2 +1,2 @@
-export { default } from './field'
 export * from './field'
+export { FieldInputProps } from './input'
```
jaketrent, line 2: Why was this change required?

packages/multiselect/src/react/index.tsx
```diff
@@ -1,18 +1,23 @@
...
+import {
+  FilterFn,
+  OnStateChangeFn,
+  MultiSelectOption,
+  StateReducer
+} from './types'
 import { noop, simpleTextFilter, switchcase } from './utils'
 
-export { Option }
+export { MultiSelectOption }
```
jaketrent, line 20: Why is this wrapped in curlies?

components/package.json
EdwardIrby: Nope

components/css/index.css
EdwardIrby: Left over from original direction. I'll remove that.

packages/multiselect/src/react/index.tsx
EdwardIrby: I was refactoring the name not how it had already been coded

packages/avatar/src/react/index.tsx
EdwardIrby: Named export now

packages/field/src/react/index.ts
EdwardIrby: That type is useful when composing components using field so I'm exposing it as that's the purpose of field.

components/src/index.ts
EdwardIrby: When I originally started this refactor using 2a approach these exports seemed most reasonable. They now seem even more so. I actually think this will create too many exports if we export vars and js. Also I say this factor makes a good argument for keeping the statics. As for css no there is no need to export that js as we're moving to vanilla css.

jaketrent: 1. How do we know we have "too many" exports? What's the threshold for that standard?

2. I feel like if we cut off the `js` exports we lose a good amount of functionality that we could expose. Do you sense this?

3. If we lean on statics, why have some exports like `themeNames` at all? One reason is that statics will bloat bundle size. Things will be attached to each other and unable to be tree-shaken. Named exports will solve that. Do you see the same problem?

packages/field/src/react/index.ts
jaketrent: We often accomplish this through: `React.ComponentProps<typeof FieldInput>`. 

But we aren't consistent and sometimes export the specific prop interface too. Preference?

packages/multiselect/src/react/index.tsx
jaketrent: Fair enough. I was asking because I wasn't sure about the syntax generally and hoped you'd be able to explain to me how it works. Is this a short-hand for exporting local vars?

packages/field/src/react/index.ts
EdwardIrby: I prefer interfaces. I've found this actually contributed to the css prop error in #1698

packages/multiselect/src/react/index.tsx
EdwardIrby: oh that yea it is. If you import a variable use and also want to re-export it.

packages/field/src/react/index.ts
jaketrent: orly... well, that'd be a great reason not to use React.ComponentProps. Look at the #1698 PR though, I don't see it used; I'm missing the connection there. Mind specifying that for me?

EdwardIrby: https://github.com/pluralsight/design-system/pull/1698/files#r648525051

PR #1769, opened 2021-05-26
Carousel: Flex and Paging

jaketrent:
The new carousel ui/ux is attempting to match this FB carousel (the “featured games” one at the top) https://www.facebook.com/games/instantgames/?store_visit_source=left_rail_navigation

Item width flexibility was overhauled. 

New "extra wide"/"large" size added.

The future paging has similarities and differences with the current:

Same:
1. Responsive - fit # items based on container width
2. Forward/back buttons - move the # items on the “page” at a time
3. Left/right arrows - ditto

Different:
1. Can tab through items, advancing the carousel per item
2. Because the window of visible items can advance per item, now there isn’t a consistent page number or page count. Eg, in the current paging, for a carousel of 4 items at a time, item index 0-3 was on page 1 and index 4-7 was on page 1. In the future, the carousel starts with 0-3 visible, tab through to index 4, now 1-4 is visible, then press/arrow next and 5-8 is visible
3. Arrowing and pressing the next/previous buttons does not change the focused items; it only changes the scrolling. So future tabbing will follow document order from the active element.
4. No partial pages pages will be allowed. Eg, a carousel showing 4 items at a time, when given 6 items will start on item index 0-3, press next, and will see 2-5. Exception is where item count * item width max doesn’t fill the available space.


packages/carousel/src/react/control.tsx
```diff
@@ -43,45 +43,36 @@ interface ControlProps extends Omit<HTMLPropsFor<'button'>, 'onClick'> {
...
-    pageCount = 0,
-    next,
-    prev,
-    setTransitioning
-  } = context
-
   const isPrev = direction === Control.directions.prev
-  const visible = isPrev ? activePage > 0 : activePage !== pageCount - 1
+  const isVisible = isPrev ? context.isPrevVisible : context.isNextVisible
+  const label = isPrev ? 'Previous items' : 'Next items'
```
EdwardIrby, line 59: Is this better language than what we previously had for accessibility?

EdwardIrby (approved):
I think it looks clean. I have a question about testing but realize the previous test suite did not really exercise interactions like we do with dropdown or select. I wonder if we should add that to this now or in the future.

packages/carousel/src/react/control.tsx
jaketrent: I don't know that it's better or not. It matches the Facebook naming. At the least, "items" seems better than "page # of #" since the concept of pages has been changed and diminished.

jaketrent:
Resolves: #1776

PR #1774, opened 2021-05-27
feat(datepicker): custom date format

toddmcneill:
### What You're Solving

- Currently, the display format for is hard-coded to use 'MM/dd/yyyy'
- With this change, someone using this function can pass their own date format string to control how the date is displayed and entered in the text field.

### Design Decisions

- I added an optional `dateFormat` parameter to the `useDateSelectChange` and `useRangeSelectChange` functions. It defaults to the same format that was already used, so it should be backward compatible.
- The input is validated with a regex `/[0-1][0-9]\/[0-3][0-9]\/[0-9][0-9][0-9][0-9]/`, which doesn't catch some of the nuances of dates (for example, Feb 30th is invalid, but matches the regex)
- I replaced the regex test with the `isMatch` function from date-fns. Any RangeError thrown by that function is caught, since it typically means the date string doesn't match the format string. Other errors are rethrown.
- Replacing the regex broke one test. But it looks like the test was incorrect. The date string used had a "13" for the month, which should be invalid, but passed the regex test, leading to `nextSlide` being set to 'backward'. Now, using the more restrictive `isMatch`, the date string is detected as invalid, and `nextSlide` is now correctly being set to `undefined`. I updated the test to match.

### DIFF
--- a/docs/src/content/components/datepicker.mdx
+++ b/docs/src/content/components/datepicker.mdx
@@ -302,6 +302,12 @@ return (
     type="React.Dispatch<React.SetStateAction<ValueOf<typeof slides>>>"
     desc="the set state action for the selected date"
   />
+  <TypesProp
+    name="dateFormat"
+    optional
+    type="string"
+    desc="a <a href='https://date-fns.org/v2.16.1/docs/format'>date-fns compatible format string</a>"
+  />
 </TypesTable>
 
 **Returns:** an array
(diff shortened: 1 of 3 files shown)


docs/src/content/components/datepicker.mdx
```diff
@@ -302,6 +302,12 @@ return (
     type="React.Dispatch<React.SetStateAction<ValueOf<typeof slides>>>"
     desc="the set state action for the selected date"
   />
+  <TypesProp
+    name="dateFormat"
+    optional
+    type="string"
+    desc="a <a href='https://date-fns.org/v2.16.1/docs/format'>date-fns compatible format string</a>"
+  />
```
toddmcneill, line 310: After running `yarn build` and `yarn start`, these additions don't show up in the documentation. Is there something else I need to do?

PR #1781, opened 2021-06-02
Refactor/view toggle

sweeting:
New styling to visually align with secondary buttons and increased text size.

CleanShot 2021-06-01 at 15 52 58


jaketrent (approved):
So fresh! I like it.

Tests good. Storybook good.

I have a small feeling that it doesn't feel super clickable and might not be immediately perceived as interactive. I'm not sure where that comes from or what to do about it yet. Maybe something about the lack of margin around the active item. Maybe the overall muted colors.

Also noticing that we don't have a blue halo on this, and I think we should add it sometime later.

jaketrent:
Oh, another thought: I think we should mark this as a BREAKING CHANGE because the outer dimensions have changed.

That might require a commit amendment (or new commit if you have more changes) before merge.

sweeting:
I'd love to add hover and focus styles, but I'll need some help.

jaketrent:
Happy to help on the focus styles. I lean toward doing it in a separate PR, but if you feel the design change makes it more immediately needful, happy to help immediately.

PR #1782, opened 2021-06-02
refactor(viewtoggle): larger text style

sweeting:
BREAKING CHANGE: increase vertical height from 24px to 32px


jaketrent (approved):
Amazing!

PR #1786, opened 2021-06-03
PR of CSP

jaketrent:
- test(position): convert stories to csf; add axe test
- test(row): convert stories to csf; add axe test


EdwardIrby (changes requested):
Position stories need some work. Sorta borked right now. Otherwise looks good.

jaketrent:
@EdwardIrby Thanks for the review. Will you please provide some more detail: What changes would you like to see? What does "borked" mean in this case? Thanks!

EdwardIrby:
> @EdwardIrby Thanks for the review. Will you please provide some more detail: What changes would you like to see? What does "borked" mean in this case? Thanks!

The following stories are visually broken compared to what's on master. When you combined stories you didn't take into account absolute and fixed positioning.

- http://localhost:6006/?path=/story/components-position--positions
- http://localhost:6006/?path=/story/components-position--positions-in-custom-portal
- http://localhost:6006/?path=/story/components-position--target
- http://localhost:6006/?path=/story/components-position--javascript-positions

Also this story seems redundant you only need the one example. Please see current master stories for reference.
- http://localhost:6006/?path=/story/components-position--positions-on-edge

jaketrent:
Hey @EdwardIrby, thanks for additional specifics.

I don't have visually broken stories at the links you provided:

1. For the Position component: http://localhost:6006/?path=/story/components-position--positions

one

2. For the custom portal: http://localhost:6006/?path=/story/components-position--positions-in-custom-portal

two

3. For the target node: http://localhost:6006/?path=/story/components-position--target

three

4. For the js functions: http://localhost:6006/?path=/story/components-position--javascript-positions

four

I also don't think the last story is redundant. Its purpose is to show through page layout that the Tooltip will not position offscreen when triggered near the edge. It uses the Position components again, yes, but it is the context in which they are used that makes this story uniquely useful.

jaketrent:
@EdwardIrby Given that we're seeing differently on the visual layout, would you mind pulling the branch, rebuilding and refreshing to see if it can be better on your end. And I'll try some other browsers. And would you respond to my reasoning on the non-redundancy of the last story, please? Thank you.

EdwardIrby:
@jaketrent I always run the following when doing a PR `yarn clean & yarn && yarn build && yarn test`. I did another just in case. On chrome I'm seeing the following using  34 inch monitor. Even when I pop the stories out of the chrome I see this unless I maximize the window to full height

### http://localhost:6006/?path=/story/components-position--positions
Screen Shot 2021-06-07 at 3 14 36 PM

### http://localhost:6006/?path=/story/components-position--positions-in-custom-portal
Screen Shot 2021-06-07 at 3 15 19 PM

### http://localhost:6006/?path=/story/components-position--target
Screen Shot 2021-06-07 at 3 16 26 PM

### http://localhost:6006/?path=/story/components-position--javascript-positions
Screen Shot 2021-06-07 at 3 16 57 PM

### http://localhost:6006/?path=/story/components-position--positions-on-edge
This one seems to needlessly repeat 8 times in the story. 
Screen Shot 2021-06-07 at 3 18 22 PM
Screen Shot 2021-06-07 at 3 18 52 PM

jaketrent:
@EdwardIrby I checked master, and the issue you have pointed out pre-exists there. The tooltip for offscreen items sticks to the bottom of the viewport until that item is scrolled into view. The story organization (long pages) now only makes it more apparent. I propose that this is a separate card we could create that wouldn't hold up CSF conversion (and thus single package work). Do agree with that proposal? If not, what is your alternate proposal?

To the PositionsAtEdge story, I still don't see this story as redundant. There are 8 examples in the story because there are 8 components. This story checks to see if each component contains the logic for keeping the positioned element in-viewport. Do you agree with that assessment? If not, what do you think needs to change and why?

EdwardIrby:
@jaketrent lets card the tooltip offscreen ones. But the position at edge as you say "test keeping the positioned element in-viewport". The combined story doesn't allow us to verify that for those eight positions. They need to each be in a separate story.

jaketrent:
@EdwardIrby thanks for your thoughts. I have carded the off-screen scrolling issue in #1793. I have updated the "position at edge" story to exercise all the different position components. I misread the loop before and didn't realize that it wasn't getting them exercised as I thought I was seeing. Hopefully it seems legitimately useful now. Please re-review.

EdwardIrby (approved):

Issue #1746, opened 2021-05-12
verticaltabs: Tab navigation when collapsible


justinclary:
### Expected behavior

* When a collapsible vertical tab has been opened and then closed, it's children should not be able to be tabbed to

### Actual behavior

* After a collapsible vertical tab has been opened and then closed, it's children are able to be tabbed to

### Steps to repro

1. Open a collapsible vertical tab with items
2. Close the vertical tab
3. Use tab for page navigation, the child items are tabbed to before moving to top level

### Related packages

- verticaltabs@7.1.4

### Environment

- OS - MacOS 11.3 (Big Sur)
- Browser version - Chrome 90.0.4430.93

jaketrent:
Repro'ed here: https://o2uxc.csb.app/

As mentioned, can only tab to tier1 headers at first, and this is right. Mouse over, then out. It collapses, but later tabs focus (invisibly) on the now-collapsed items.

PR #1802, opened 2021-06-14
fix(verticaltabs): disallow focusing collapsed tier2 headers

jaketrent:
Resolves: #1746


packages/verticaltabs/src/react/item.tsx
```diff
@@ -242,6 +242,11 @@ const Tier2Header = React.forwardRef<any, ItemHeaderProps>((props, ref) => {
       <span {...wrapperProps} ref={ref as RefFor<'span'>} />
     )
 
+  if (props.collapsed || hideLabels) {
+    rest['aria-hidden'] = 'true'
+    rest.tabIndex = -1
+  }
+
```
EdwardIrby, line 249: I'd prefer it if you create a new object here instead of mutation the props and passed that object as a spread after rest.

EdwardIrby (approved):

PR #1809, opened 2021-06-15
A11y/circular progress landmark

### LINKED ISSUE #1749: circularprogress: resolve unique landmark problem
Likely in linearprogress too.

Likely use progressbar and aria-valuenow
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_progressbar_role

EdwardIrby:
resolves #1749


packages/circularprogress/src/react/__specs__/index.spec.tsx
```diff
@@ -11,7 +11,10 @@ describe('CircularProgress', () => {
 
   it('renders', () => {
     const { getByTestId } = render(
-      <CircularProgress data-testid="mock-component" />
+      <CircularProgress
+        data-testid="mock-component"
+        aria-label="ciruclar progress"
```
jaketrent, line 16:
```suggestion
        aria-label="circular progress"
```

packages/circularprogress/src/react/__specs__/index.spec.tsx
```diff
@@ -20,19 +23,19 @@ describe('CircularProgress', () => {
   it('forwards refs', () => {
     const ref = React.createRef<HTMLDivElement>()
 
-    render(<CircularProgress ref={ref} />)
+    render(<CircularProgress ref={ref} aria-label="ciruclar progress" />)
```
jaketrent, line 26:
```suggestion
    render(<CircularProgress ref={ref} aria-label="circular progress" />)
```

packages/circularprogress/src/react/__specs__/index.spec.tsx
```diff
@@ -20,19 +23,19 @@ describe('CircularProgress', () => {
...
 
     expect(ref.current).not.toBeNull()
   })
 
   // TODO: fix the landmark unique id problem;
   // Likely use progressbar and aria-valuenow
   // https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_progressbar_role
```
jaketrent, line 33:
```suggestion
```

packages/circularprogress/src/react/__stories__/index.story.tsx
```diff
@@ -17,14 +17,19 @@ export const ValuesAndSizes: Story = () => (
         ...acc,
         <div key={'val' + value}>{value}</div>,
         ...Object.values(CircularProgress.sizes).map(size => (
-          <CircularProgress key={size + value} value={value} size={size} />
+          <CircularProgress
+            aria-label="ciruclar progress"
```
jaketrent, line 21:
```suggestion
            aria-label="circular progress"
```

packages/circularprogress/src/react/__stories__/index.story.tsx
```diff
@@ -17,14 +17,19 @@ export const ValuesAndSizes: Story = () => (
...
+            size={size}
+          />
         ))
       ],
       []
     )}
     <div>Indeterminate</div>
     {Object.values(CircularProgress.sizes).map(size => (
-      <CircularProgress size={size} key={size} />
+      <CircularProgress aria-label="ciruclar progress" size={size} key={size} />
```
jaketrent, line 32:
```suggestion
      <CircularProgress aria-label="circular progress" size={size} key={size} />
```

packages/circularprogress/src/react/__stories__/index.story.tsx
```diff
@@ -44,7 +49,7 @@ export const Animation: Story = () => {
 
     return (
       <div>
-        <CircularProgress value={value} />
+        <CircularProgress aria-label="ciruclar progress" value={value} />
```
jaketrent, line 52:
```suggestion
        <CircularProgress aria-label="circular progress" value={value} />
```

packages/circularprogress/src/react/__stories__/index.story.tsx
```diff
@@ -57,11 +62,14 @@ export const Animation: Story = () => {
 }
 
 export const StyleOverride: Story = () => (
-  <CircularProgress style={{ outline: '1px solid red' }} />
+  <CircularProgress
+    aria-label="ciruclar progress"
```
jaketrent, line 66:
```suggestion
    aria-label="circular progress"
```

packages/circularprogress/src/react/index.tsx
```diff
@@ -62,6 +62,7 @@ const styles: { [key: string]: StyleFn } = {
 interface CircularProgressProps extends HTMLPropsFor<'div'> {
   size?: ValueOf<typeof vars.sizes>
   value?: number
+  'aria-label': string
```
jaketrent, line 65: Now required = breaking change. But I don't see the commit message indicating this.

Also, I see a "--wip--" commit that needs reworded to conventional commit or dropped.

packages/circularprogress/src/react/index.tsx
```diff
@@ -80,18 +81,44 @@ const CircularProgress = React.forwardRef<
...
+  const dashOffset =
+    ((100 - (value || defaultIndeterminateValue)) / 100) * circumference
+  const [busy, setBusy] = React.useState<'false' | 'true'>('false')
+  React.useEffect(() => {
+    if (value) {
+      value > 0 && setBusy('true')
+      value === 100 && setBusy('false')
+    }
+    setBusy('true')
+  }, [value])
```
jaketrent, line 96:
```suggestion
  const busy = value 
    ? value === 100
      ? 'false'
      : value > 0
        ? 'true'
        : 'false'
    : 'true'
```

Don't need state or effect. Could have same functionality with value derived from props.

packages/linearprogress/src/react/index.tsx
```diff
@@ -35,22 +35,42 @@ const styles = {
...
+}) => {
   const themeName = useTheme()
+  const [busy, setBusy] = React.useState<'false' | 'true'>('false')
+  React.useEffect(() => {
+    if (value) {
+      value > 0 && setBusy('true')
+      value === 100 && setBusy('false')
+    }
+    setBusy('true')
+  }, [value])
```
jaketrent, line 54:
```suggestion
  const busy = value 
    ? value === 100
      ? 'false'
      : value > 0
        ? 'true'
        : 'false'
    : 'true'
```

Ditto. Just a var.

packages/linearprogress/src/react/index.tsx
```diff
@@ -35,22 +35,42 @@ const styles = {
 
 interface LinearProgressProps extends HTMLPropsFor<'div'> {
   value?: number
+  'aria-label': string
```
jaketrent, line 38: another breaking change because this is now required.

jaketrent (changes requested):
Most comments inline. 

I screen-read this with voiceover, and it didn't respond as I'd expect.

The text always just reported "busy" but never indicated value amount in the case of giving it a value:

Screen Shot 2021-06-15 at 9 35 33 AM

It does seem to have the aria attrs coded for:

Screen Shot 2021-06-15 at 9 36 52 AM

Perhaps this means that these attrs are not compatible with VoiceOver? Have you been able to verify your solution with any other accessibility tool?

I also tried the Animated story to hear the announced changes to the value. But I heard nothing. I'm not used to testing this, so maybe I'm doing it wrong. If you've confirmed that works in a screenreader, please indicate to me how it's accomplished so I can also verify it.

packages/circularprogress/src/react/index.tsx
```diff
@@ -80,18 +81,44 @@ const CircularProgress = React.forwardRef<
...
-          ? 'Loading...'
-          : `${value}% complete`}
-      </ScreenReaderOnly>
+    <div
+      ref={ref}
+      {...styles.circularprogress(themeName, props)}
+      {...rest}
+      role="progressbar"
+      {...ariaAttributes}
+      aria-busy={busy}
```
jaketrent, line 114: Reading more about aria-busy, I'm not sure it's what we want here.

> Indicates an element is being modified and that assistive technologies MAY want to wait until the modifications are complete before exposing them to the user.

> The default value of aria-busy is false for all elements. When aria-busy is true for an element, assistive technologies MAY ignore changes to content owned by that element and then process all changes made during the busy period as a single, atomic update when aria-busy becomes false.

https://www.w3.org/TR/wai-aria-1.1/#aria-busy

In the section about progressbar, however, it says aria-busy should be used, but on the (other) region of the page for which loading is being indicated:

> If the progressbar is describing the loading progress of a particular region of a page, the author SHOULD use aria-describedby to point to the status, and set the aria-busy attribute to true on the region until it is finished loading.

https://www.w3.org/TR/wai-aria-1.1/#progressbar

Based on that knowledge, I'd recommend removing it. Do you have other knowledge that says otherwise?

EdwardIrby: I followed Deque on this pattern 🤷🏿‍♂️
https://dequeuniversity.com/library/aria/progress-bar-bounded 
https://dequeuniversity.com/library/aria/progress-bar-unbounded

packages/circularprogress/src/react/index.tsx
EdwardIrby: My commits are always squashed I'll add breaking to the last one. I honestly don't think require something that should be there for a11y is breaking.

packages/circularprogress/src/react/__specs__/index.spec.tsx
EdwardIrby: No clue honestly I used what is available to me to use https://chrome.google.com/webstore/detail/screen-reader/kgejglhpjiefppelpmljglcjbhoiplfn. I also followed the Deque pattern. Once I got it reading what I wanted I stopped.

packages/circularprogress/src/react/index.tsx
jaketrent: If someone runs tsc in their build, in practical terms it will be breaking.

jaketrent: Also, squashed is fine to get rid of the WIP commit. But still want to be able to gen new versions for both circularprogress and linearprogress packages. That will require 2 commits.

packages/circularprogress/src/react/index.tsx
jaketrent: ^ Yeah, that's totally not what the ARIA guidelines say. Deque is trusted for a lot, though, so maybe this too.

Practically, if aria-busy can hide reading via screenreader, does removing it solve the other issue I had with not getting good screenreads? If so, venerable though Deque is, I'd still remove it.

This might be a good thing to pass by Amber to get another take.

EdwardIrby:
@jaketrent talked to Amber. I got one small recommendation and she approved current implementation.

jaketrent:
@EdwardIrby What was her recommendation? And did she screen read the current impl? If so, in what screenreader?

EdwardIrby:
> @EdwardIrby What was her recommendation? And did she screen read the current impl? If so, in what screenreader?

We walked through the experience pairing remotely and to see if the experience was correct using Macs assistive technology. She agreed it was but asked me to round numbers so I did.

jaketrent (approved):
Okey dokey, thanks for following up with Amber on the interaction experience.

Issue #1798, opened 2021-06-11
radio: JAWS doubles the count of buttons in a group


jaketrent:
If 5 buttons, says 10 buttons in JAWS. VoiceOver says 5. 

Radio has 2 elements, nested, that have aria label.

Suggestion: attempt aria-hidden on `<input />` tag.

From Roshni on Blanketsburg

jaketrent:
Roshni reports her test is still not merged. Still waiting for her successful test before adding.

Issue #1779, opened 2021-06-01
form: fix axe errors


jaketrent:
Such as:

- ARIA toggle fields have an accessible name 

Then enable new axe-core test.

PR #1810, opened 2021-06-15
axe fixes for switch and radio

jaketrent:
Resolves: #1779
I think this will also resolve: #1798


EdwardIrby (approved):
Should we modify the styles for radio's input to also use core's accessibility.screenReaderOnly like switch does?

Issue #1817, opened 2021-06-18
core: sync css vars with js vars in core


jaketrent:
There have been changes like removal of base colors and additions of "csv" colors on the js side of core. The css vars side has not been updated. These (and possibly other) deltas needs synced.

PR #1818, opened 2021-06-18
CSS build support and Avatar CSS conversion

jaketrent:
Fixed JS->CSS build which had been broken since ESM. Added scripts to Build for reuse. Exposed refactor helper in Core. Used Storybook postcss preset. 

What do you think of the utils?
How about the conversion itself?


packages/avatar/src/css/index.css
```diff
@@ -0,0 +1,56 @@
...
+  object-fit: cover;
+}
+.psds-avatar__initials {
+  width: auto;
+  height: 100%;
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  color: var(--psColorsTextIconHighOnDark);
+  /* TODO: revert to bg util color after it exists in css vars */
```
jaketrent, line 53: see https://github.com/pluralsight/design-system/issues/1817

EdwardIrby:
@jaketrent we should probably have @sweeting look at these going forward for visual errors that test can not catch.

EdwardIrby (changes requested):
Seeing the following warning message when running build in Avatar
`You did not set any plugins, parser, or stringifier. Right now, PostCSS does nothing. Pick plugins for your case on https://www.postcss.parts/ and use them in postcss.config.js.`

Further we need to build to a path relative to the index.js files in the esm and cjs directories. 
- `dist/cjs/css/index.css`
- `dist/esm/css/index.css`

Storybook wouldn't show these failing nor our test but for the distributed packages we want them at these locations.

EdwardIrby:
Visually looks the same as the glamor variant!

jaketrent:
@EdwardIrby Thanks for the review. Good catch on the postcss plugin missing. 

I'm not thrilled with having the css under the `dist/esm` and `dist/cjs` subdirs. The `index.css` file is not of that tech and thus seems wrongly categorized. I know that this is for the `dist` dir to be able to relative import like the `src` dir. I don't have another solution in mind yet.

PR #1821, opened 2021-06-21
simplify scroll restoration for docs

jaketrent:
Still works locally. Not sure why sessionStorage is acting weird in prod. 

This is an attempt to simplify and hopefully rule out more potentially problems in the differences between envs.

If this doesn't work, I may move toward a window.var approach.


sweeting (commented):
Works well from a user experience.

EdwardIrby (approved):

PR #1825, opened 2021-06-22
fix(avatar): prevent css from being tree-shaken out

jaketrent:
From:
https://webpack.js.org/guides/tree-shaking/#mark-the-file-as-side-effect-free

This means if you use something like css-loader in your project and
import a CSS file, it needs to be added to the side effect list so
it will not be unintentionally dropped in production mode


jaketrent:
Confirmed locally that this fix also fixes the current styles issue on DS docs prod

Screen Shot 2021-06-24 at 11 56 38 AM

Local prod build:

Screen Shot 2021-06-24 at 11 57 23 AM

jaketrent:
@danethurber since Edward is out, you're here, and the change is small and impactful, I wonder if you might not give this a second set of eyes?

danethurber (approved):
nice!

danethurber:
do you need to do this to every project?

jaketrent:
@danethurber yes, and we will. We have only converted Avatar atm.

Issue #1819, opened 2021-06-21
css: move to vanilla css


jaketrent:
Resolves: #1686 

## Components

- [x] #1857
- [x] #1856
- [x] convert css for avatar
- [x] #1833
- [x] #1834
- [x] #1841
- [x] remove glamor utils from build
- [x] #1842
- [x] #1846
- [x] #1847
- [x] #1859
- [x] #1865
- [x] convert css for collapsible
- [x] convert css for core
- [x] #1866
- [ ] #1867
- [x] #1869
- [x] #1871
- [x] #1873
- [x] #1891
- [x] #1892
- [x] convert css for featureflags
- [x] #1922
- [x] #1923
- [x] #1924
- [x] #1925
- [x] #1926
- [x] #1927
- [x] #1930
- [x] #1929
- [x] #1880
- [x] #1920
- [x] #1919
- [x] #1918
- [x] #1917
- [x] #1916
- [x] #1913
- [x] convert css for normalize
- [x] #1912
- [x] convert css for position
- [x] #1911
- [x] #1910
- [x] #1909
- [x] #1906
- [x] #1905
- [x] #1904
- [x] #1903
- [x] #1902
- [ ] convert css for storybook-addon-center
- [ ] convert css for storybook-addon-theme
- [ ] convert css for storybook-preset
- [x] #1899
- [x] #1898
- [x] #1897
- [x] #1896
- [x] #1895
- [x] #1879
- [x] #1878
- [x] #1877
- [x] #1876
- [x] #1875
- [x] #1863
- [x] convert css for util
- [x] #1861
- [x] #1862

## Steps

*** = manual step required

To get automated steps done, run:

```
cd myPackage
cp ../../scripts/upgradeCss.mjs .
node upgradeCss.mjs
```

1. Convert imported core vars in src/css/index.ts to
import { refactor: { cssVars: { named, imports } } } from '@pluralsight/ps-deisgn-system-core'
2. Run old build script
3. Cp dist/index.css to src/css
4. Adjust css script in package.json
"build:css": "postcss-compile"
5. Adjust style in package.json
"style": "dist/esm/css/index.css",
6. Replace "sideEffects": false with "sideEffects": ["**/*.css"],
7. Remove "glamor" from "peerDependencies"
8. Copy avatar/postcss.config.js to dir
9. *** Make any dashify adjustments in css as needed
10. *** Adjust in-react css usage to use:
- import stylesheet
- className in jsx
[...]

PR #1832, opened 2021-06-24
All CSS Conversions

jaketrent:
We're going to make PRs to this branch for continued conversions.

Resolves: #1833
and others...

Add a new script that does a lot but not everything.  To use,

```
cd myPackage
cp ../../scripts/upgradeCss.mjs .
node upgradeCss.mjs
```

See new instructions on issue https://github.com/pluralsight/design-system/issues/1819 for which steps the script covers and which are still manually required.

This PR now includes dashifyin core and each component as we go.


scripts/upgradeCss.mjs
```diff
@@ -0,0 +1,138 @@
...
+  const pkg = JSON.parse(readFile('package.json'))
+
+  pkg.scripts['build:css'] = 'postcss-compile'
+  pkg.style = 'dist/esm/css/index.css'
+  if (pkg.peerDependencies) {
+    const { glamor, ...rest } = pkg.peerDependencies
+    pkg.peerDependencies = rest
+  }
+  if (pkg.sideEffects.toString() === 'false') {
+    pkg.sideEffects = ['**/*.css']
```
EdwardIrby, line 58: What does setting this mean for tree shaking?

EdwardIrby (approved):
Love the automation

scripts/upgradeCss.mjs
jaketrent: It means that the css file should be considered to provide a side effect. It can't be tree-shaken and keep the program in tact -- in this case, the visuals.

jaketrent:
Hey Edward, would you please re-review. There is now a dashification of the CSS which means breaking change in Core.

EdwardIrby (approved):
Like the dashes

EdwardIrby (approved):

packages/typeahead/src/react/index.tsx
```diff
@@ -219,14 +213,19 @@ const Typeahead: TypeaheadFieldComponent = props => {
...
       })
     }
 
-    return <Field.Label {...getLabelProps()}>{label}</Field.Label>
+    return (
+      <Field.Label {...getLabelProps()} aria-label={_ariaLabel}>
+        {label}
+      </Field.Label>
+    )
   }, [label, getLabelProps])
```
EdwardIrby, line 229: @jaketrent aria-label flow I'm thinking

EdwardIrby:
This is a lot of commits. You should probably squash and merge this instead of committing it with all these commits.

jaketrent:
@EdwardIrby I think I'd rather merge with commits in tact. Git likes lots of commits. I do too. For many reasons, including easier to browse, cohesive, see incremental progress through history, better git-bisect experience. Why squash?

jaketrent:
Working on the master back-merge right now, btw.

sweeting (approved):
After your last commit to dashify CSS custom properties, I built the docs and clicked through all the pages. Everything looked good to me.

EdwardIrby (approved):
Other than needing to find and remove all files at `/packages/**/src/css/*.index.ts` it looks good.

jaketrent:
@EdwardIrby removed used src/css/*.ts files.

PR #1838, opened 2021-06-25
add changelog link to datepicker doc

jaketrent:
From user doc walkthrough that we noticed this.

Brian, I'm not sure if this is the right place to put this, but I think it's the best at the moment. The DatePicker docs are not the common case in what they contain or how they're organized.

Resolves: #1837

### DIFF
--- a/docs/src/content/components/datepicker.mdx
+++ b/docs/src/content/components/datepicker.mdx
@@ -15,6 +15,13 @@ import { layout } from '@pluralsight/ps-design-system-core'
 
 ## Components
 
+<Usage
+  install="npm install @pluralsight/ps-design-system-datepicker"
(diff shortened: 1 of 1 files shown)


sweeting (approved):
Yes, this works for now. 

I will look into how we can improve the hierarchy and ordering of the content sections of component pages for more consistency. The current structure can vary quite a bit from component to component, and I think it would be good to have a starting point template for new component docs pages with some basic content structure in place.

jaketrent:
@sweeting k, sounds good. Yeah, this component has more stuff exposed than many, like the "functions" section. And it requires so much boilerplate, except in the case of the convenience wrapper, DatePicker, that there was an attempt to explain parts of the process separately instead of full working examples. I like that there is space for variability between pages. But I do like the idea of consistent landmarks.

PR #1868, opened 2021-06-30
css/viewtoggle-typeahead

### LINKED ISSUE #1861: convert css for verticaltabs

### LINKED ISSUE #1862: convert css for viewtoggle

### LINKED ISSUE #1863: convert css for typeahead

EdwardIrby:
resolves #1862
resolves #1861 
resolves #1863


packages/verticaltabs/src/react/group.tsx
```diff
@@ -57,11 +48,11 @@ interface GroupStatics {
 interface GroupComponent
   extends RefForwardingComponent<GroupProps, HTMLLIElement, GroupStatics> {}
 
-const Group = React.forwardRef((props, ref) => {
+const Group = React.forwardRef(({ className, ...props }, ref) => {
```
jaketrent, line 51: We destructure `className` here? Seems to disallow styling of Group and nothing else.

packages/verticaltabs/src/react/index.tsx
```diff
@@ -49,7 +46,11 @@ const VerticalTabs = React.forwardRef((props, ref) => {
 
   return (
     <Context.Provider value={contextValue}>
-      <ul {...styles.verticaltabs()} ref={ref} {...rest}>
+      <ul
+        className={classNames('psds-verticaltabs', className)}
```
jaketrent, line 50: Your commit message says that there's a breaking change, moving the className to parent  node. I don't think that's the case. It was on the parent node already. This seems to also be true from the other subcomponents in this package that I observed. This is probably true for all 3 commits in this PR.

packages/verticaltabs/src/react/item.tsx
```diff
@@ -9,91 +9,39 @@ import {
...
 import { List, CollapsibleList } from './list'
-
-type StyleFn = (
-  themeName?: ValueOf<keyof typeof Theme.names>,
-  props?: Record<string, unknown>
-) => glamorExports.StyleAttribute
-
-const glamor = glamorDefault || glamorExports
+import '../css/index.css'
+import classNames from 'classnames'
```
jaketrent, line 17: Auto import bug? I think we want to use our own `classNames` fn from utils.

packages/verticaltabs/src/react/__specs__/index.spec.tsx
```diff
@@ -16,6 +16,14 @@ describe('VerticalTabs', () => {
...
   })
 
+  it('composes className', () => {
+    const { container } = render(<VerticalTabs className="compose-classname" />)
+
+    expect(container.firstChild).toHaveClass(
+      'psds-verticaltabs compose-classname'
+    )
+  })
+
```
jaketrent, line 26: Please add className compose tests for the subcomponents that are exposed as named exports or are attached to `VerticalTabs` function.

packages/verticaltabs/src/css/index.css
```diff
@@ -0,0 +1,311 @@
...
+  list-style: none;
+  margin: 0;
+  padding: 0;
+}
+.psds-verticaltabs__list--collapsible {
+  overflow: hidden;
+  transition: height var(--ps-motion-speed-normal);
+  visibility: hidden;
+  -webkit-transition: height var(--ps-motion-speed-normal);
+  -moz-transition: height var(--ps-motion-speed-normal);
```
jaketrent, line 29: We do not want vendor-specific attributes in the `src` css. That's why we build the css for publishing.

packages/typeahead/src/react/index.tsx
```diff
@@ -237,9 +231,14 @@ const Typeahead: TypeaheadFieldComponent = props => {
...
-        <div {...styles.wrapper(isOpen)}>
+        <div
+          className={classNames(
+            'psds-multi-select__wrapper',
+            isOpen && 'psds-multi-select__wrapper--open'
+          )}
+        >
           <Menu
-            {...styles.menu()}
+            className="psds-multi-select__menu"
```
jaketrent, line 241: Please add className prop composition to parent node of exposed components.

jaketrent (changes requested):
I came up with a few things to review and adjust. Let's discuss as needed.

packages/verticaltabs/src/react/item.tsx
EdwardIrby: looks like

packages/typeahead/src/react/index.tsx
EdwardIrby: Not sure that applies here, due to how we designed Position quite frankly. If I had made this thing it would of been a hook not a jsx Element and not something that wrapped what I see as the root node. From a render standpoint Field is the root and thus it already receives the rest spread.

jaketrent: I think the argument about choosing a different root node is a good one. Sometimes it's not super obvious. If it's Field, I'm good with that. Maybe I'm misreading this instance. In general, I think the parent node is usually the most worth while, because css selectors can have child selectors but not parent selectors.

packages/verticaltabs/src/react/__stories__/examples.story.tsx
```diff
@@ -65,23 +65,30 @@ export const AdminTools: Story<
...
-                      {item.title}
-                    </Tier2.Header>
-                  )
-
-                  return (
-                    <Tier2 active={active} header={itemHeader} key={itemKey} />
-                  )
-                })}
+                {
+                  // @ts-ignore: story
```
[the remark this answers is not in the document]
EdwardIrby: This should have been throwing before I looked at this file to make sure there were no linting errors resulting from refactor which these are not the result of. This is outside the scope of the refactor. Honestly I have no clue how it was not hit before this frankly.  I will create a card for this.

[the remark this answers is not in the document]
EdwardIrby: #1872

packages/verticaltabs/src/react/index.tsx
EdwardIrby: Copy paste error looking for what was a good default message for this rather repetitive tasks

packages/verticaltabs/src/react/group.tsx
EdwardIrby: Error on my part

packages/verticaltabs/src/react/divider.tsx
```diff
@@ -1,24 +1,20 @@
...
 
 const Divider: React.FC = () => {
   const themeName = useTheme()
 
-  return <hr {...styles.divider(themeName)} />
+  return (
+    <hr
+      className={classNames(
+        'psds-verticaltabs__divider',
+        `psds-theme--${themeName}`
```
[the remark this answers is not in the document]
EdwardIrby: Sure

packages/verticaltabs/src/css/index.css
jaketrent: Re-review: this vendor css still needs removed.

packages/verticaltabs/src/react/__specs__/index.spec.tsx
```diff
@@ -16,6 +18,90 @@ describe('VerticalTabs', () => {
...
+      'psds-verticaltabs__divider psds-theme--dark compose-classname'
+    )
+  })
+
+  it('composes className (CollapsibleGroup)', () => {
+    const { container } = render(
+      <CollapsibleGroup className="compose-classname" />
+    )
+
+    expect(container.firstChild).toHaveClass('compose-classname')
```
jaketrent, line 44: tip for here or the future: 

```
toHaveClass(
  'psds-verticaltabs__divider compose-classname'
)
```

is more valuable than:

```
toHaveClass(
  'compose-classname'
)
```

because it tests _composition_ in addition to props being properly attached to elements.

packages/verticaltabs/src/react/item.tsx
```diff
@@ -190,32 +140,52 @@ const Tier1: React.FC<Tier1Props> & { Header: typeof Tier1Header } = props => (
...
+      />
     )
   return (
-    <Tag {...styles.tier1Header()} {...rest} aria-expanded={!collapsed}>
+    <Tag {...rest} aria-expanded={!collapsed}>
       {icon &&
         React.cloneElement(icon, {
           size: CaretDownIcon.sizes.medium,
-          ...styles.itemIcon(),
+          className: classNames(styles.itemIcon()),
```
jaketrent, line 178: Do we want to be able to pass a className to the icon? If so, that would need merged here.

jaketrent (changes requested):
One cleanup still needed. Another question added.

packages/verticaltabs/src/react/__specs__/index.spec.tsx
EdwardIrby: unless of course there is no className on the root node as is the case here. I can rename test to ` properly takes className`

PR #1870, opened 2021-06-30
CSS Convert 4

jaketrent:
asdf


packages/dialog/src/css/index.css
```diff
@@ -0,0 +1,128 @@
...
+  content: ' ';
+  display: block;
+  height: 14px;
+  line-height: 0;
+  position: absolute;
+  transform: rotate(45deg);
+  white-space: pre;
+  width: 14px;
+}
+.psds-dialog--tailposition-topleft:after {
```
EdwardIrby, line 41: Should we  dash case these tail position classes. I don't feel strongly on it but it would be nice to decided on when we want to treat things as whole words versus hyphenated ones.

EdwardIrby (approved):

packages/dialog/src/css/index.css
jaketrent: Great question.

What I think we could do: (diff from current practice, as I paused to think about your question)

1. Insert dashes everywhere there are separate works (where there once were Capitals)
2. dashify in code the js vars that concat into selectors

For example:
```
import { dashify } from '......-util'
classNames(`psds-dialog--tail-position-${dashify(tailPosition)}`)
```

Today I'm converting like this usually:

```
classNames(`psds-dialog--tail-position-${tailPosition.toLowerCase()}`)
```

And in this case even missed a dash:

```
classNames(`psds-dialog--tailposition-${tailPosition.toLowerCase()}`)
```

So I think there is some refactoring work to do to use a util#dashify instead of toLowerCase. I'll plan to do that in my next PR.

jaketrent:
crud, accidentally back-merged master instead of convertScript branch. I've cherry-picked to jtcss5 branch and will do the dashify modification there.

packages/dialog/src/css/index.css
jaketrent: Talked together again. This is the more-full standard:

1. Dashify whole words (eg, `tail-position`)
2. Acronomyns are not dashed (eg, `svg`, `csv`)
3. Abbreviations are not dashed (eg, `xsmall`, `xxlarge`)

PR #1889, opened 2021-07-01
feat(tooltip): allowOffscreen rendering

### LINKED ISSUE #1793: position: off-screen tooltips in stories scroll with viewport
### Expected behavior

Positioned tooltips stay positioned relative to the offscreen elements. 

### Actual behavior

If an element if offscreen (below the fold), the tooltip for it sits at the bottom of the viewport and stays there until the offscreen element scrolls into view. Then it stays in the correct position.

See: https://github.com/pluralsight/design-system/pull/1786

### Steps to repro

1. Run Position stories
2. Ensure some positioned tooltips are below the fold in the viewport
3. Scroll
4. See tooltips move

EdwardIrby:
resolves #1793

### DIFF
--- a/docs/src/content/components/position.mdx
+++ b/docs/src/content/components/position.mdx
@@ -191,6 +191,70 @@ const Example: React.FC = props => {
 export default Example
 ```
 
+### Allow Offscreen
+The `allowOffscreen` property allows a developer to enable rendering content outside the viewport.
+This is useful when you want to render a tooltip open on page load and the tooltip should render below the fold.
+
+```typescript
+import React, { useState } from 'react'
+
+import Button from '@pluralsight/ps-design-system-button'
+import Tooltip from '@pluralsight/ps-design-system-tooltip'
+
+import { RightOf } from '@pluralsight/ps-design-system-position'
+
+const Example: React.FC = props => {
+  const [hovered, setHovered] = useState(true)
+
+  const hide = () => setHovered(false)
+  const show = () => setHovered(true)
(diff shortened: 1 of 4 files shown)


docs/src/content/components/position.mdx
```diff
@@ -215,6 +215,12 @@ import {
 />
 
 <TypesTable>
+   <TypesProp
+    name="allowOffscreen"
+    type="boolean"
+    desc="allow rendering of positioned element offscreen, useful when open on page load"
```
jaketrent, line 221: I don't understand the "useful" comment. Might benefit from an example section where recommendations like this could benefit from room for explanation.

packages/position/src/js/index.ts
```diff
@@ -261,7 +301,7 @@ export const leftOf: PositionFunction = target => {
 function overrideDefaultOpts(
   opts: Partial<StyleOptions> | undefined
 ): StyleOptions {
-  return { ...opts, ...defaultOptions }
+  return { ...defaultOptions, ...opts }
```
jaketrent, line 304: lol, nice catch

jaketrent (changes requested):
Works great. One important change: the commit needs to be scoped to `position`, not `tooltip`.

docs/src/content/components/position.mdx
EdwardIrby: Will do!

jaketrent (approved):

PR #1928, opened 2021-07-08
CSS Convert 7 - Field, Form, Halo, Icon, Layout

jaketrent:


sweeting (approved):
Visual check of
✅ Field
✅ Form
✅ Halo
✅ Icon
✅ Layout

packages/icon/src/react/index.tsx
```diff
@@ -35,9 +22,20 @@ export interface IconComponent
...
-  return <div {...style.icon({ color, size })} {...rest} ref={ref} />
+  const { className, size = sizes.medium, color, ...rest } = props
+
+  return (
+    <div
+      {...rest}
+      className={classNames(
+        'psds-icon',
+        `psds-icon--size-${size}`,
+        `psds-icon--color-${color}`,
```
EdwardIrby, line 33: There is no default color on icon so we should ```color && `psds-icon--color-${color}` ``` I noticed classes with undefined in the icon snapshot.

EdwardIrby (changes requested):

packages/icon/src/react/index.tsx
jaketrent: agreed. done.

EdwardIrby (approved):

PR #1933, opened 2021-07-13
feat: autoselect current date or selected date

### LINKED ISSUE #1836: datepicker: make keyboard accessible
### Expected behavior

- Tab-accessible days on the calendar
- Keyboard-openable calendar from datepicker
- More?

### Steps to repro

1. Open calendar in codesandbox, try to tab to day, can't
2. Open DS docs, try to open calendar from datepicker cal icon, can't

Notify `#dev-learner-workflow` when done.

EdwardIrby:
@sweeting Need to know focus styles for buttons as they are focusable and you can tab through them
@jaketrent want to pull in new css stuff before adding focus styles and tests

Using this for for the key events https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/datepicker-dialog.html
However I'm not using these rules want to run this through an a11y test later but we already have good key navigation and this seems to add unnecessary complexity plus the library we're using does it differently. https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/datepicker-dialog.html#rps_label_5

resolves #1836


jaketrent (commented):
Hey Edward, code so far looks fine. 

I'm a little unclear about the state of things when reading your comments. 

1. I think you're saying that the key events here are now implemented https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/datepicker-dialog.html
2. When I built and went through the stories, I didn't get any of those key events to work for me. I didn't see an autofocus, or at least I didn't think so, because tabbing didn't seem to change focused cal dates. Arrows or page up/down also didn't seem to work.
3. I'm good with waiting for css merging before focus styles and tests

So, I think this is undone and on purpose, but I'm not sure. At the very least, it seems that keyboard nav still is not working.

packages/datepicker/src/react/__specs__/index.spec.tsx
```diff
@@ -11,7 +11,8 @@ import {
   onRangeDateSelected,
   onMultiDateSelected,
   useDateSelectChange,
-  useRangeSelectChange
+  useRangeSelectChange,
+  DatePicker
```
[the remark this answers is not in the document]
EdwardIrby: Waiting on css refactor to merge this will clash

packages/datepicker/src/react/utils.ts
```diff
@@ -71,33 +71,38 @@ interface OnMultipleDatesSelected {
...
-        const firstTime = selected[0].getTime()
-        if (firstTime < dateTime) {
-          newDates.push(date)
-        } else {
-          newDates.unshift(date)
-        }
-        setSelected(newDates)
-      } else if (newDates.length === 2) {
-        setSelected([date])
+export const onRangeDateSelected = ({
```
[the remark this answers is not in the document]
EdwardIrby: yep

EdwardIrby:
So page up and page down song with the plus shift variants work. However struggled getting them to with like w3 of example. Think I should split that out and get this in. Will right test later this evening.

packages/datepicker/src/react/calendar.tsx
```diff
@@ -64,11 +69,22 @@ interface CalendarProps
...
 export const Calendar = React.forwardRef<HTMLDivElement, CalendarProps>(
   (
-    { calendars, getBackProps, getForwardProps, children, slide, ...rest },
+    {
+      calendars,
+      getBackProps,
+      getForwardProps,
+      children,
+      slide,
+      uniqueId = defaultUniqueId,
```
jaketrent, line 84: This is the current approach to solving this problem for snapshots. It's a bit painful, wouldn't you agree?

I would eventually love another solution. If you have ideas you like, I'd love to hear them.

(This is a musing, not an ask for change in this PR.)

packages/datepicker/src/react/calendar.tsx
```diff
@@ -103,6 +119,16 @@ export const Calendar = React.forwardRef<HTMLDivElement, CalendarProps>(
         return () => el.removeEventListener('animationend', updateOffset)
       }
     }, [])
+    const gridLabels = calendars.map((calendar, i) =>
```
jaketrent, line 122: `monthIds` seems more clear.

packages/datepicker/src/react/context.ts
```diff
@@ -1,6 +1,16 @@
...
+    dateObj: {
+      [x: string]: any
+    }
+  ) => React.KeyboardEventHandler<HTMLButtonElement>
+}
 const initialValue = {}
-// @ts-ignore: really hate this part of context and typescript
-export const DateContext = React.createContext<Calendar>(initialValue)
+export const DateContext = React.createContext<DateContextInterface>(
+  // @ts-ignore: really hate this part of context and typescript
```
jaketrent, line 14: This seems like a bad idea. Above you have the interface showing non-null types. Then here we suppress the error. If the fields here are potentially nullable, TS should know and then we will too when we go to code against context fields. 

What do you think? Would love to hear counter arguments, if you prefer.

packages/datepicker/src/css/index.ts
```diff
@@ -109,7 +109,10 @@ export default {
     backgroundColor: 'transparent',
     border: 'none',
     [`&:focus`]: {
-      outline: 'none'
+      borderRadius: '50%',
+      border: `2px solid ${colorsBlue[6]}`,
+      background: colorsBlue[2],
+      color: colorsTextIcon.highOnLight
```
sweeting, line 115:
```suggestion
      color: colorsTextIcon.highOnLight,
      outline: 'none'
```

@EdwardIrby Since we have a custom styles for focus state the default focus styles can be removed.

sweeting (changes requested):
Looks good. My only suggestion is to remove the default focus styles on the days.

packages/datepicker/src/react/use-key-events.ts
```diff
@@ -0,0 +1,84 @@
...
+  const updateFocused = () => {
+    focusedDate &&
+      document
+        .querySelector<HTMLButtonElement>(
+          `[aria-label="${format(focusedDate, 'EEE LLL dd yyyy')}"]`
+        )
+        ?.focus()
+  }
+
+  const focusNext = useDebounceCallback(updateFocused, 100)
```
jaketrent, line 29: Please help me understand why this needs debounced?

packages/datepicker/src/react/use-key-events.ts
```diff
@@ -0,0 +1,84 @@
...
+              days: shift
+            })
+          : sub(date, {
+              days: shift
+            })
+      setFocusedDate(nextDate)
+      if (!months.includes(getMonth(nextDate))) {
+        dir === 1
+          ? handleForwardClick(
+              evt as unknown as React.MouseEvent<HTMLButtonElement>
```
jaketrent, line 58: Not a huge deal based on today's implementation, I believe, but since a KeyboardEvent is not a MouseEvent and differs in its properties, it might not be great to pass one event to the other's handler.

I would suggest an event type-agnostic function like `moveForwardMonth` or something of that sort. What do you think?

packages/datepicker/src/react/use-key-events.ts
```diff
@@ -0,0 +1,84 @@
...
+        ;[' ', 'enter'].includes(key) && dateProps.onClick(evt)(evt)
+        key === 'arrowup' && handleDayShift(-1, 7, evt)
+        key === 'arrowdown' && handleDayShift(1, 7, evt)
+        key === 'arrowleft' && handleDayShift(-1, 1, evt)
+        key === 'arrowright' && handleDayShift(1, 1, evt)
+        key === 'home' && handleWeekEndStartFocus(-1)
+        key === 'end' && handleWeekEndStartFocus(1)
+      }
+    return handleKeyDown
+  }
```
jaketrent, line 81: Love it! Nice function.

Quick question on line 72: Is it an error to have onClick(evt)(evt) called like that? I'm having a hard time following dateProps back to where it originates from in order to check. Perhaps you could point that out, please?

jaketrent (commented):
Looks like code is coming along.

Styling is looking good in storybook. Did you change your mind about when you want to merge in relation to the css conversion merging?

Are you saying in your previous comment that you want to do the PageUp/Down month switching in a different issue? If so, I'm wondering, is there code already in this PR that is a part of that solution and could be removed until the point that new card is worked? If so, I'd like that so-as to make sure all the code here is absolutely needed by the code here. This component is getting rather complicated and more challenging to navigate.

EdwardIrby:
@jaketrent yeah just realized this was missing two commits. Please re-review it should make more sense.

packages/datepicker/src/react/use-key-events.ts
EdwardIrby: Switching between months causes a rerender such that we can't focus a button immediately in the new month view. Debouncing for 100ms worked better than waiting on RAF.

packages/datepicker/src/react/calendar.tsx
EdwardIrby: Not sure honestly my normal approach for this is to overwrite a global but I can't see a reasonable way around this other than mimicking downshifts approach which is an iterator I believe.

jaketrent: Yeah, I like the iterator better. Outside scope here, if wanted, but I would love to move to it.

packages/datepicker/src/react/calendar.tsx
jaketrent: I think code readability is usually a top priority. I was making a suggestion toward readability.

My thought flowed from this fact:

From the [docs](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-labelledby_attribute):

> The aria-labelledby attribute establishes relationships between objects and their label(s), and its value should be one or more element IDs.

Seem like they're ids, and they're applied to the element that is labelling a month. I didn't know what grid was.

jaketrent:
Hey @EdwardIrby, A couple questions:

I see one new commit, d97852c. It's small, and you said there'd be two. Is there another that didn't get pushed? From the two commits, what is the additional sense that you anticipate I'll be able to see?

Also, I have a couple outstanding questions that I would appreciate your response on, please:

- [Re mixing event types](https://github.com/pluralsight/design-system/pull/1933#discussion_r672675788)
- [Re tracing dateProps.onClick to its definition to confirm usage](https://github.com/pluralsight/design-system/pull/1933#discussion_r672680231)

packages/datepicker/src/react/use-key-events.ts
EdwardIrby: The library we're use only gives us this signature to use. It's an onClick I have to type cast like so to use this functionality to correctly move the month. I'm basing this off the libraries own code examples. It cannot be changed.

packages/datepicker/src/react/use-key-events.ts
EdwardIrby: This is outdated.

EdwardIrby:
> Hey @EdwardIrby, A couple questions:
> 
> I see one new commit, [d97852c](https://github.com/pluralsight/design-system/commit/d97852c1ba538e551c81bb81e6ed18a001bf55c7). It's small, and you said there'd be two. Is there another that didn't get pushed? From the two commits, what is the additional sense that you anticipate I'll be able to see?
> 
> Also, I have a couple outstanding questions that I would appreciate your response on, please:
> 
> * [Re mixing event types](https://github.com/pluralsight/design-system/pull/1933#discussion_r672675788)
> * [Re tracing dateProps.onClick to its definition to confirm usage](https://github.com/pluralsight/design-system/pull/1933#discussion_r672680231)

`dateProps.onClick` is outdated
`e mixing event type` is a necessary type cast

jaketrent (changes requested):
Hey @EdwardIrby I was doing some more keyboard nav testing. The arrow key nav is really nice. I really like that experience. Way to go on nailing that!

I found a couple keyboard nav issues:

1. Whether PageUp or PageDown, the month always decrements, never increments. I think one should go forward, the other backward.
2. I would enter a date in the DatePicker.  Then I'd close the calendar. Then I'd press enter in the text input to reopen the calendar. But the focus wouldn't move to the calendar days area. I couldn't tab to get there. Keyboard nav to select a date at that point was impossible. I think the calendar days should be tabbable to focus (and then arrow navable w/in the days). This would help in another case where one tabs to the month forward/back buttons, presses them, but then can't tab into the days area of the calendar.
3. Not sure on this one, but it feels like when the Calendar has been closed on the DatePicker that the focus should return back to the text input. Currently focus is lost. In storybook, it goes to the story iframe.

sweeting:
Related to what @jaketrent said about keyboard nav, when I navigate to different months with the keyboard, I'm unable to tab to focus on the first day of the month I'm in. 

Also comparing that with the [Dayzed examples](https://dayzed.netlify.app/examples/single), it seems both tab and arrow keys work to navigate between days.

EdwardIrby:
> Related to what @jaketrent said about keyboard nav, when I navigate to different months with the keyboard, I'm unable to tab to focus on the first day of the month I'm in.
> 
> Also comparing that with the [Dayzed examples](https://dayzed.netlify.app/examples/single), it seems both tab and arrow keys work to navigate between days.

Tabbing is not part of the spec actually for the date grid https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/datepicker-dialog.html#kbd_label_5. You're suppose to use the arrow keys. Originally before this PR let you do that but just didn't have focus styles.

jaketrent:
@EdwardIrby Arrow keys are fine to use on the days grid of the calendar. But how do you get there so that you can start using the arrows? I think there's an upstream problem currently, as I mentioned in my issue number 2. If the spec says you can't tab from day to day in the days grid, that's fine, but the problem is getting to the days grid first.

Are you able to repro that issue?

sweeting (commented):
On the "Calendar with Input" story, I'm unable to get the calendar to open using the keyboard.

For the following feedback I was using the Calendar with String Label story.

- When navigating to the next/previous month, the focus gets set on the first day of the month, but I expected focus to remain on the next/previous month button.
- I didn't expect tabbing to close the date picker dialog. I expected it to cycle between the date in the grid and the month navigation arrows.
- Voiceover speaks "Jul 2021" over and over again when the datepicker dialog is opened.

I was looking at this as reference for keyboard shortcuts and date picker a11y practices:
https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/datepicker-dialog.html

EdwardIrby:
@jaketrent this needs a non unique id to past test.
@sweeting having trouble getting it to focus back on month button @jaketrent can you give it a shot?

jaketrent:
Hey @EdwardIrby , yeah, I can work on month button focus if I get to it before you do. I'm doubtful I will, because I still see the NextJS blocker from our CSS release as the highest priority item, and that may take a bit, but maybe I'll get lucky.

EdwardIrby:
Will re-open once solution is is found in for CSS/NextJS bug

jaketrent (approved):
Keyboard nav is very impressive. It works as expected. Tabbing and arrowing stories seem solid.

One issue I noticed is on the DatePicker: When the calendar is closed after selecting a date, focus is not brought back to the input field. It moves to document.body. This seems like something we'd want to change.

The other thing I noticed was that sometimes the calendar doesn't visually contain months with 6 week rows.

Screen Shot 2021-09-01 at 10 54 26 AM

Great work!

sweeting (dismissed):
Calendar With Input story: Unable to open date picker

Date Picker String Label story: 
- Date picker opens (that's good).
- Tab key selects previous month only. Unable to use Tab to get to next month button or days.
- Shift+Tab is the only way to get back to days once previous month button has focus.
- Using the arrow keys to navigate to other months I'm unable to use tab to get to previous/next month buttons.

EdwardIrby:
> Calendar With Input story: Unable to open date picker
> 
> Date Picker String Label story:
> 
> * Date picker opens (that's good).
> * Tab key selects previous month only. Unable to use Tab to get to next month button or days.
> * Shift+Tab is the only way to get back to days once previous month button has focus.
> * Using the arrow keys to navigate to other months I'm unable to use tab to get to previous/next month buttons.

Wasn't able to reproduce the  Calendar With Input story error. However that story is not the Datepicker but rather used as an example of how to build your own custom datepicker with an input so it's okay to ignore. As for focus try it now.

sweeting (commented):
@EdwardIrby 

- I can navigate between previous month, next month, and days now. 👍 
- There's a flash of focus from the previous month button to the next month button when hitting "Enter" to navigate to the next month. Not sure if there's anything that can be done about that.
- If I tab to the days and use the arrow keys to navigate, I'm unable to use the tab key again to switch focus between previous month, next month, and days, and I'm unable to hit "Escape" to close the date picker dialog.

I recorded a video of this behavior and sent to you in Slack.

sweeting (commented):
- Hitting enter on Next month no longer shows a flash of focus on the previous month button first 👍 

The steps I take to reproduce the bug where you can't use the tab key or hit escape to close the window are:
1. Open the dialog by hitting the enter key.
2. Use any arrow keys to move day focus 1 or more times.
3. Hit tab key to switch focus to previous/next buttons (nothing happens)
4. Hit escape key (dialog doesn't close)

This might not be a common scenario, but I can reproduce it consistently.

sweeting:
Another concern is how our datepicker reads using a screen reader compared to the [WAI-ARIA example](https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/datepicker-dialog.html#kbd_label_5). This is using VoiceOver on my Mac, but is very different reader experience.

https://user-images.githubusercontent.com/48542/132547972-cf9c96d6-e92e-4129-a2e3-447f88079245.mp4

PR #1936, opened 2021-07-15
css/navitem-multiselect

EdwardIrby:
resolves #1916 
resolves #1917 
resolves #1918 
resolves #1919 
resolves #1920

PR #1940, opened 2021-07-19
Fix multiselect two-liner

jaketrent:
Resolves: https://github.com/pluralsight/design-system/pull/1936#pullrequestreview-708750359

### DIFF
--- a/packages/multiselect/src/css/index.css
+++ b/packages/multiselect/src/css/index.css
@@ -79,11 +79,8 @@
   margin: calc(4px / 2);
 }
 
-.psds-multi-select__input {
+.psds-multi-select__input-container input {
   min-height: 24px;
   min-width: 50px;
 }
 
-.psds-multi-select__input:disabled {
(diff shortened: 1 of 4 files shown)


packages/multiselect/src/css/index.css
```diff
@@ -79,11 +79,8 @@
   margin: calc(4px / 2);
 }
 
-.psds-multi-select__input {
+.psds-multi-select__input-container input {
   min-height: 24px;
   min-width: 50px;
 }
 
-.psds-multi-select__input:disabled {
```
jaketrent, line 87: This was unused afaik. Field component has this attribute on its input always.

EdwardIrby: Yeah I prefer not to ever use a non direct selector or reference a tag. I already resolved this using a better selector see my PR

jaketrent:
Fixed elsewhere apparently. Closing.

Issue #1952, opened 2021-07-28
typeahead: add forwardRef


jaketrent:
related: https://pluralsight.slack.com/archives/C70HPSJPP/p1627494545201500

jaketrent:
This is already possible with `renderInputTag`:

```
import Field from '@pluralsight/ps-design-system-field'

export const Autofocus: Story = () => {
  const options = React.useMemo(() => periodicElements, [])
  const [value, setValue] = React.useState<string>()
  const ref = React.createRef<HTMLInputElement>()

  React.useEffect(() => {
    ref.current?.focus()
  })

  return (
    <Typeahead
      renderInputTag={React.forwardRef((props, forwardedRef) => {
        return (
          <Field.Input
            {...props}
            ref={ref}
            placeholder="tag from render prop"
          />
        )
      })}
      onChange={(
        evt: React.ChangeEvent<HTMLInputElement> | null,
        selectedItem?: {
          label: React.ReactText
          value?: React.ReactText
        }
      ) => {
        setValue(evt?.target.value || '')
      }}
      options={options}
      value={value}
    />
  )
}
```

PR #1955, opened 2021-07-30
feat(typeahead): add forwardRef to the HTMLInputElement

jaketrent:
Resolves: #1952


packages/typeahead/src/react/index.tsx
```diff
@@ -84,215 +81,224 @@ const defaultFilterFunc = (
...
+              : selectedItem || undefined
+          )
+        setInputItems(filterFunction(options, inputValue))
+      }
+    })
+    React.useEffect(() => {
+      onChange && onChange(null, activeItem || undefined)
+    }, [activeItem, onChange])
+    const { value: inputValue, ...inputProps } = getInputProps({
+      ref: forwardedRef,
```
jaketrent, line 163: This is the key line change in the body of the function.

The rest is formatting from the identation change because of the forwardRef wrapping.

EdwardIrby (approved):

Issue #1950, opened 2021-07-27
all: find a solution for loading DS components with CSS on NextJS


jaketrent:
Somehow we missed this in prep for the CSS transition. But this is a blocker for NextJS users. 

In short, our components can't import their own CSS on NextJS as of this version (11.1).

For more context, see: https://pluralsight.slack.com/archives/C70HPSJPP/p1627340844162300

Jake Set 1 (merged)

- avatar
- button
- icon

Jake Set 2 (merged)

- dialog
- textinput
- halo
- searchinput
- textarea
- circularprogress
- text

Jake Set 3 (in PR)

- actionmenu
- badge
- breadcrumb
- carousel
- card
- checkbox

Jake Set 4 (WIP) 
- datepicker
- field
- dropdown
- form
- layout

Remaining 
- appframe (scrollable dep)
- banner (text dep)
- datawell (text dep)
- drawer (row, text dep)
- emptystate (tag dep)
- errors (text dep)

Edward will work on
- linearprogress
- link
- menu
- multiselect
- navbar
- navbrand
- navcookiebanner
- navitem
- navuser
- note
- radio

Edward Set 1
- row
- screenreaderonly
- scrollable
- select
- starrating
- steps
- switch
- tab
- table
- tag
- tagsinput
- tooltip
- typeahead
- util
- verticaltabs
- viewtoggle

jaketrent:
Potential solutions, in order of my personal desirability:

1. next.config.js workaround that allows css-loader to load global css from components imported from node_modules 
2. Separate exports in the DS for components and stylesheets separately; leave export default in tact and combined
3. Change CSS implementation; Move to CSS modules; export selector map to allow overrides from CSS-in-JS.

jaketrent:
Attempting num 2 with this approach...

Next wants these separate:

import Button from 'button/react'
import 'button/styles.css'

(1) Move internal component css import to src/index:
import './css/index.css'
export { default } from './react/index'
export * from './vars/index'

Has to be separate all the way down (for DS component dependencies, eg, Button -> Icon)

(2) Rewrite the other internal styled DS component imports
In button/react/index.tsx:
import Icon from 'icon/react'

(3) Add *their* CSS import to src/index.tsx
import 'icon/styles.css'
import './css/index.css'
export { default } from './react/index'
export * from './vars/index'

(4) Create convenience path files to support these naming conventions
In button/:
echo "module.exports = require('./dist/cjs/react/index.js')" > react.js
[...]

PR #1960, opened 2021-08-10
nextjs 1: enable separate style imports

jaketrent:
Subset of what's required to address: https://github.com/pluralsight/design-system/issues/1950

Design decision details: https://github.com/pluralsight/design-system/issues/1950#issuecomment-889291797

Tested default usage in CRA. Tested separate usage in NextJS. Both work in my tests.

Want to merge, release and send to Will to test in his app before doing too many more of these.


packages/avatar/package.json
```diff
@@ -25,7 +28,8 @@
     "test:watch": "yarn test --watchAll"
   },
   "sideEffects": [
-    "**/*.css"
+    "**/*.css",
+    "src/index.ts"
```
jaketrent, line 32: required to avoid tree-shaking out

packages/avatar/package.json
```diff
@@ -9,6 +9,9 @@
   "main": "dist/cjs/index.js",
   "module": "dist/esm/index.js",
   "files": [
+    "react.d.ts",
+    "react.mjs",
+    "styles.css",
```
jaketrent, line 14: files back in package root dirs again

packages/avatar/react.d.ts
```diff
@@ -0,0 +1,23 @@
+import {
```
jaketrent, line 1: cp and adjust of dist/esm/index.ts

should probably create a script to do the copy and adjust in a yarn build

packages/avatar/react.mjs
```diff
@@ -0,0 +1,3 @@
+export { default } from './dist/esm/react/index'
```
jaketrent, line 1: cp and adjust of src/index.ts (no css import)

packages/avatar/src/index.ts
```diff
@@ -1,3 +1,5 @@
+import './css/index.css'
```
jaketrent, line 1: mv css import here for `import Button from 'button'` (default/easy) usage.

packages/avatar/src/react/__stories__/index.story.tsx
```diff
@@ -1,7 +1,7 @@
 import { Meta, Story } from '@storybook/react/types-6-0'
 import React from 'react'
 
-import Avatar from '../index'
+import Avatar from '../../index'
```
jaketrent, line 4: use src/index.ts (default/easy/styled file) instead of src/react/index.ts (unstyled)

packages/avatar/react.d.ts
EdwardIrby: I concur on this

packages/avatar/src/react/__stories__/index.story.tsx
EdwardIrby: This pattern might present a problem when we write stories for sub-components that are not re-exported named or as statics. What do you think? Is this necessary for the next.js solution?

EdwardIrby (approved):
Get test passing and then cool let's roll with this if you can answer my questions about components.

packages/avatar/src/react/__stories__/index.story.tsx
jaketrent: This is not necessary for nextjs. This line is necessary for storybook to get the styled version of the component. 

If there are stories for sub-components, we'll want to import the `src/index.ts` file alongside. It's my prediction that will usually already be done.

PR #1962, opened 2021-08-13
feat(*): export additional component types

aj-foster:
This PR exports additional types and interfaces from a number of components.

The goal of these additional exports is to enable further manipulation of the components. For example, exporting the `DropdownComponent` type makes it much easier to [hoist the statics](https://github.com/mridgway/hoist-non-react-statics) from `Dropdown` into a new component.

While it is debatable whether folks **should** do that sort of thing, this change will enable it. :) 

No guarantees that this covers all of the types, and I don't have enough context to know if this will ruin everything.

### DIFF
--- a/packages/actionmenu/src/react/index.tsx
+++ b/packages/actionmenu/src/react/index.tsx
@@ -38,7 +38,7 @@ interface ActionMenuProps
   onClose?: () => void
   origin?: ValueOf<typeof vars.origins>
 }
-type ActionMenuComponent = RefForwardingComponent<
+export type ActionMenuComponent = RefForwardingComponent<
   ActionMenuProps,
   HTMLUListElement,
   ActionMenuStatics
--- a/packages/appframe/src/react/index.tsx
+++ b/packages/appframe/src/react/index.tsx
@@ -66,7 +66,7 @@ interface AppFrameStatics {
   TopNav: typeof TopNav
 }
 
-interface AppFrameComponent
+export interface AppFrameComponent
   extends RefForwardingComponent<
     AppFrameProps,
     HTMLDivElement,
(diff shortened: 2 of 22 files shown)


EdwardIrby:
@jaketrent  See discussion at https://pluralsight.slack.com/archives/C70HPSJPP/p1628879010022900

jaketrent:
If this makes life easier for you as a user, AJ, this works for me. It expands our public API, but not very much. I've never loved our `*Component` interface naming, but I don't have a better idea yet. Making this name public will make it concrete and an important interface to follow. Consistency-wise, not every component has one of these interfaces defined yet.

I still don't quite see the picture of why you're manipulating the type and need access to the statics, but maybe it's for extending components to do something additionally special. Maybe it's a Dropdown wrapper that you want to hoist Dropdown statics onto.

Sounds like this PR was to start a conversation. I don't have significant reservations on this point.

The PR seems incomplete. Need to cover the other components. Need to export it from src/index.ts too. Maybe something else? AJ, do you plan on finishing this work or hoping that we prioritize and complete this work? If us, do you have a timeline you were hoping for? And what effort would this work support? 

Thanks for your engagement!

aj-foster:
Thanks for engaging in the conversation. :) 

We are not blocked by not having the changes from this PR, so it would be my preference that we're 100% comfortable with extending the public API in this way before we move forward. I'll let you prioritize the topics of renaming and bringing consistency to these types as you see fit. When the time comes, consider me a rubber duck and an extra set of hands where needed.

jaketrent:
Closing for now since this doesn't seem like an active PR. Sounds like there's plenty of time and space to work on this in the future. Will not create a current priority for this item at the moment given existing priorities and capacity. Will attach a "revisit" tag. AJ, if you feel more urgency around this, please bring it up and feel free to add upon your already-made contributions.

PR #1964, opened 2021-08-17
Example/webpack css modules

### LINKED ISSUE #1953: Investigate Webpack Config issues with new global CSS component files
There appears to be a problem with using Webpack with our global styles. In the example inside our of our repo we don't see this. However two slack threads lead me to believe that the problem may be with css-modules being used on the consumer side. 

https://pluralsight.slack.com/archives/C70HPSJPP/p1627407466167300

https://pluralsight.slack.com/archives/C70HPSJPP/p1627494761201900

Both of them use css-modules and miniExtractPlugin. Which differs from our example. However the erros they are receiving are different.

Screen Shot 2021-07-29 at 10 32 14 AM

Carolina's team is seeing a file-type issues indicating it's a loader that's missing for what is a vanilla CSS file. 
Andy's team is seeing a document not defined error in the postcss-loader.  .

EdwardIrby:
resolves #1953


examples/bundle-css-modules-webpack/webpack.config.js
```diff
@@ -0,0 +1,57 @@
...
+module.exports = {
+  mode: 'development',
+  entry: './src/index.js',
+  output: {
+    path: path.resolve(__dirname, 'public'),
+    filename: 'app.js',
+    clean: true,
+    publicPath: '/',
+  },
+  // optimization: {
```
jaketrent, line 13: not sure if you want this in there.

jaketrent (approved):

PR #1970, opened 2021-08-23
Re-add old camelCase vars to Core

jaketrent:
Will allow teams to handle partial upgrade situations, such as that recently experienced by Flow.

For example, in a conversation with Flow devs:

This is a problem with our Core package, where we changed the variable names. Then, you upgraded some components and not others. And you worked with another team who updated some code but not others. And the single set of new variable names was insufficient to handle this partial upgrade situation. So, the most likely option was to rollback until you could upgrade everything, but that seems difficult to do at once, especially with other teams.

I thought of a proposed solution this morning and wondered what you thought. What if we put the old vars side by side with the new vars in the latest version of the core package? This would allow everything to work, whether it needed the old or new vars, once you upgrade Core to latest. The old vars in that package would be immediately deprecated to show that they’d be eventually removed. I don’t know when that removal would be, but it should give a reasonable window for all code to be upgraded gradually and then also not remain a drag in the DS codebase for a long time.

The team was agreed that this is a helpful solution.

### DIFF
--- a/packages/core/src/css/accessibility.css
+++ b/packages/core/src/css/accessibility.css
@@ -1,3 +1,4 @@
+.screenreader-only, /* DEPRECATED */
 .psds-screenreader-only {
   position: absolute;
   top: auto;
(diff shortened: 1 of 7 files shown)


jaketrent:
Solved instead by clients importing old and new side by side as needed. See https://pluralsight.slack.com/archives/C70HPSJPP/p1629741102040200

PR #1972, opened 2021-08-24
Css/refactor exports row viewtoggle

EdwardIrby:
CSS import refactor 
See issue #1950

    field
    menu
    row
    screenreaderonly
    scrollable
    select
    starrating
    steps
    switch
    tab
    table
    tag
    tagsinput
    text
    textarea
    textinput
    tooltip
    typeahead
    util
    verticaltabs
    viewtoggle


EdwardIrby:
Forgot to copy over definition files and do mjs/js files. Just a second.

jaketrent (changes requested):
Made it through `select` component, and I'm seeing a few things that need covered that seem like widespread needs:

- Great job removing `./index.css` imports in react/index.tsx files. But the other DS component styles need removed too. Use the `component/react` import path to do this.
- Once those styles are removed from that file, they need added to two other files: 1) `src/index.ts` and `styles.css`. Use the `component/styles.css` to do this.
- The `react.mjs`, `react.js` and `react.d.ts` files are generally missing. Use `src/index.ts` as the base for `react.mjs`; use `button/react.js` or another component as the base for `react.js`; and use `dist/esm/react/index.d.ts` as the base for `react.d.ts`.
- While making these src adjustments, please also test with `yarn test` and `yarn storybook` and `tsc` if it's not run during test. This will ensure new imports are working. Spot checking storybook on `select` showed broken visuals for me.

EdwardIrby:
@jaketrent I've got failing test because I'm getting undefined Icons? amd I not suppose to do something like this.

`import { CheckIcon } from '@pluralsight/ps-design-system-icon/react'`

jaketrent:
Edward, yes, the named icon import should be as you mention above. Tests use cjs, and there's a bug in the exports in icon/react.js. My latest PR has a fix for icon.

EdwardIrby:
> Edward, yes, the named icon import should be as you mention above. Tests use cjs, and there's a bug in the exports in icon/react.js. My latest PR has a fix for icon.

Should we get yours in first then ? Or should I submit my PR against your branch instead?

packages/select/src/index.ts
```diff
@@ -1,2 +1,7 @@
-export { default, useListbox, UseListboxProps, useMenuRef } from './react/index'
+import '@pluralsight/ps-design-system-icon/styles.css'
+// import '@pluralsight/ps-design-system-menu/styles.css'
```
EdwardIrby, line 2: @jaketrent So I ran storybook in Select with this uncommented and it broke. When I comment them out it works and icons look correct. I thought we were suppose to do this. I can't figure out what is going on with this.  I know menu also imports `icon/styles.css`

jaketrent: Yes, we are supposed to include all the dependant component styles. This is for user convenience, so they only have to import the top-level component /styles.css.

EdwardIrby:
May re-open if we can't find a solution in #1977

PR #1976, opened 2021-08-26
Separate imports, set 3

jaketrent:
- feat(actionmenu): allow separate import for react and css
- feat(badge): allow separate import for react and css
- feat(breadcrumb): allow separate import for react and css
- feat(carousel): allow separate import for react and css
- feat(card): allow separate import for react and css
- feat(checkbox): allow separate import for react and css


packages/breadcrumb/src/index.ts
```diff
@@ -1 +1,7 @@
-export { default } from './react/index'
+import '@pluralsight/ps-design-system-button/styles.css'
```
EdwardIrby, line 1: So  if you run storybook on this it will error. I ran into the same thing on Select. If you import a component styles that also imports another component styles it throws the error below in storybook. 

Screen Shot 2021-08-27 at 11 13 30 PM

EdwardIrby (changes requested):

jaketrent:
Pausing in favor of #1977.

PR #1993, opened 2021-09-14
fix/drawer-checkbox-css

EdwardIrby:
fixes: https://pluralsight.slack.com/archives/C70HPSJPP/p1630095874000800


sweeting (changes requested):
Checkbox on light theme looks good

Still seeing the extra border on the row in drawer example on light theme

EdwardIrby:
> Checkbox on light theme looks good
> 
> Still seeing the extra border on the row in drawer example on light theme
> 
> CleanShot 2021-09-14 at 10 15 07

Yea I used the docs for debugging. Selectivity is slightly different in storybook. We really do need a stack component where layout and borders can be managed like you said during standup" It's fixed now.

sweeting (approved):
Drawer with row on light theme looks good to me now.

PR #2002, opened 2021-09-23
fix(position): fix scroll jank caused by the position  component

EdwardIrby:
https://pluralsight.slack.com/archives/C70HPSJPP/p1632416018084000

### DIFF
--- a/packages/position/src/react/index.tsx
+++ b/packages/position/src/react/index.tsx
@@ -160,7 +160,10 @@ export const Position = React.forwardRef<HTMLElement, PositionProps>(
             )}
 
         {createUniversalPortal(
-          <div style={{ visibility: shownOnce ? 'visible' : 'hidden' }}>
+          <div
+            style={{ display: shownOnce ? 'initial' : 'none' }}
+            aria-live="polite"
+          >
             {when && showEl}
           </div>,
           inNode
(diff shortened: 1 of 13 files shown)


packages/position/src/react/index.tsx
```diff
@@ -160,7 +160,10 @@ export const Position = React.forwardRef<HTMLElement, PositionProps>(
             )}
 
         {createUniversalPortal(
-          <div style={{ visibility: shownOnce ? 'visible' : 'hidden' }}>
+          <div
+            style={{ display: shownOnce ? 'initial' : 'none' }}
```
danethurber, line 164: 👍  guessing this is to hide the scrollbar. was changing the overflow an option at all

packages/position/src/react/index.tsx
```diff
@@ -160,7 +160,10 @@ export const Position = React.forwardRef<HTMLElement, PositionProps>(
...
+            aria-live="polite"
```
danethurber, line 165: is this to limit the amount of updates read out to the user while scrolling?

danethurber (approved):
LGTM. I asked a couple questions for my own edification

packages/position/src/react/index.tsx
EdwardIrby: I wanted to let the screen reader know something will eventually appear here after a user interaction. Initially on mount this is an empty div. We should let assistive technology know its dynamic and may render content at some point (politely).

packages/position/src/react/index.tsx
EdwardIrby: Overflow not really. Other than that single style this `div` had no other styles applied to it. It's just the mount point for a positioned element. We do positioning on the `showEl`  we're cloning. I could probably try add some checks to make sure `showEl` is rendering on screen the first pass but that's more JS for a simple css fix.

PR #2003, opened 2021-09-23
fix(drawer): border-top on row on light-theme removed

EdwardIrby:
https://pluralsight.slack.com/archives/C70HPSJPP/p1630095874000800


sweeting (approved):
👍 Docs light theme example of a drawer with a row looks good to me.

PR #2027, opened 2021-11-23
Checkbox and Radio - remove readOnly attributes

dannytannertantrum:
Per [MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/readonly), "The (readOnly) attribute is not supported or relevant to `<select>` or input types that are already not mutable, such as checkbox and radio..."

### What we're solving
This fixes an issue in Safari where keyboard navigation was unable to reach these input elements.

### Steps
- Ensure this works in all browsers, but in particular, make sure it works in Safari.
- Visit an [empty search query page](https://app.pluralsight.com/search) in prod.
- In Chrome and Firefox, notice how you can reach the checkbox and radio button filters by using keyboard navigation.
- Try the same thing in Safari and notice how it does not work.
- On this branch, run storybook and note how you can now interact with these input elements in Safari.


packages/checkbox/src/react/index.tsx
```diff
@@ -30,6 +30,9 @@ interface CheckboxProps
   value: string | number
 }
 
+/* eslint-disable-next-line */
+const noop = () => {}
```
caseybaggz, line 34: We typically want to avoid adding any hacks that force us to use eslint disables as it results in a code smell. I'll check this out further to see what else we can do to get this fixed and stay within out eslint boundaries.

caseybaggz (changes requested):

packages/checkbox/src/react/index.tsx
dannytannertantrum: Sounds good! For additional context, I paired with @danethurber on this and I think we decided to go that route because checkbox and radio input elements look for `onChange` events; we're using an `onClick` event for the radio button and lacked any kind of `click` or `change` event on the checkbox. I believe we were seeing warnings logged in the tests, too. Because there's a refactor/overhaul of DS in the works, we decided to utilize a quick solution rather than refactor a bunch of code.

I'm happy with whatever you and Dane feel is best. Thanks!

caseybaggz: Awesome! Thanks for the context @dannytannertantrum . I went ahead and just extended the component to allow people to use an `onChange` if they wanted to (but not required) which resolved that issue. Thanks so much for the PR! 💪

dannytannertantrum: Great - thank you! I like that solution; nice work! 👏

caseybaggz (approved):

Issue #2051, opened 2021-12-17
[Bug?]: CircularProgress does not allow value of 0


caseybaggz:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Current behavior 😯

Due to the math being used in the component a value of `0` will incorrectly show a progress of ~25%.

### Expected behavior 🤔

Progress should allow a `value` prop of `0` and follow the a11y guidelines for a progress bar:
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_progressbar_role

### Steps to reproduce 🕹

Steps:

1. Open codesandbox example from docs
2. Add value `0` to static progress

### Context 🔦

Skills are using this component to display the progress of a course - via "@meg" / Meghan Kline-Tuls

PR #2058, opened 2022-01-18
Fix/2051/circular progress zero

Rykus0:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

#2051 - Value of `0` is visually displaying 25% complete

## What is the new behavior?

Value of `0` visually displays 0% complete

## Other information

Previously was checking if value was falsy and if so using the default value for indeterminate progress.  Since `0` is falsy, it was displaying that progress value (25).  Updated to only do this for null/undefined values.


package.json
```diff
@@ -39,6 +39,7 @@
     "publish:all:beta": "yarn publish:all --npm-tag=beta",
     "start": "yarn docs",
     "start:button": "yarn workspace @pluralsight/ps-design-system-button run storybook",
+    "start:circularprogress": "yarn workspace @pluralsight/ps-design-system-circularprogress run storybook",
```
caseybaggz, line 42: noice.

packages/circularprogress/src/react/__specs__/index.spec.tsx
```diff
@@ -38,6 +38,27 @@ describe('CircularProgress', () => {
...
+        data-testid="zero-test"
+      />
+    )
+
+    // Get the SVG path to test
+    const el = screen
+      .getByTestId('zero-test')
+      .querySelector('.psds-circularprogress__fg')
+
+    // Get the attribute value to test
```
caseybaggz, line 55: Future reference, there's no need for these comments. The code is readable enough. We typically like to roll with using the Clean Code standards - no comments, unless the code is written in a way that requires them - if so, refactor the code to make it cleaner and more readable first...if unable too, leave comment. 👍

caseybaggz (approved):

packages/circularprogress/src/react/__specs__/index.spec.tsx
Rykus0: Thanks, that makes sense.

PR #2068, opened 2022-01-26
refactor: closes #1839 - replace RefFor with React.RefObject

### LINKED ISSUE #1839: Refactor or remove alias for RefFor
## Purpose
To align with new typing patterns, the RefFor alias in util needs to be updated. Currently it is most commonly used for type casting in the ref attr in JSX and  React.useImperativeHandle.

### Refactored
- `export type RefFor<T extends HTMLElement> = React.RefObject<T>`
- `export type RefFor<T extends HTMLElement> = React.Reft<T>`
### Removed and replaced with
- `React.RefObject<HTMLElement>`
-  `React.Reft<T>`

See: https://github.com/pluralsight/design-system/pull/1811#discussion_r658055487

Rykus0:
## PR Checklist

- [ ] Tests for the changes have been added (for bug fixes / features)

## PR Type

- [ ] Bugfix
- [ ] Feature
- [x] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

Refactor to use `React.RefObject` directly instead of `RefFor` as per #1839


packages/focusmanager/src/react/use-focus-manager.ts
```diff
@@ -16,7 +16,7 @@ const FOCUSABLE_SELECTORS = [
 ]
 
 export default function useFocusManager(
-  ref: RefFor<'div'>,
+  ref: React.RefObject<HTMLDivElement>,
```
caseybaggz, line 19: In the future, steer clear of using the global React Object which is now an anti-pattern for current React settings which we will eventually move to. Once we do, we will have to update every instance in each package to use the named exports which will dramatically decrease the bundle size for our users. 😄 

This lib is about a 2 yrs behind in React conventions.

packages/util/src/types/index.ts
```diff
@@ -1,5 +1,8 @@
 import React from 'react'
 
+/**
+ * @deprecated
```
caseybaggz, line 4: Nice. Also would be a good idea to add a note with this in the future about when it happened and when we will officially remove it.

Typically, this is normally not enough since it's really just an IDE feature. Would be better to rename it using the common convention of 
```
RefFor_DEPRECATED
```

However, no need to do it now. Just future reference.

caseybaggz (approved):
🌮

packages/focusmanager/src/react/use-focus-manager.ts
Rykus0: Makes sense

PR #2094, opened 2022-02-08
Deps/upgrade gatsby

Rykus0:
## PR Checklist

- [ ] Tests for the changes have been added (for bug fixes / features)

## PR Type

- [ ] Bugfix
- [ ] Feature
- [x] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

Gatsby 2

## What is the new behavior?

Upgraded to Gatsby 4 (including related plugins)


docs/src/components/code-block/actions.tsx
```diff
@@ -6,7 +6,7 @@ import { CopyToClipboard } from 'react-copy-to-clipboard'
 
 import pkg from '../../../package.json'
 import usePrevious from '../../hooks/use-previous'
-import styles from './styles.module.css'
+import * as styles from './styles.module.css'
```
caseybaggz, line 9: How come we had to make this a wildcard?

docs/src/components/frame/frame.module.css
```diff
@@ -9,23 +9,27 @@
...
   flex-direction: column;
   min-height: 100vh;
+}
 
-  & .container {
-    display: flex;
-    flex-direction: row;
-    min-height: calc(100vh - var(--top-nav-height));
-  }
+.frame .container {
```
caseybaggz, line 14: Just for future reference - like when we get to TVA stuff, let's avoid nested selectors. That is a really bad pattern that by nature was created from preprocessors which we now know to be really bad.

Since this is the old docs, I'm not worried about it, but definitely want to get it out of your system to not do it in the future. 👍

docs/src/components/motion/easing.module.css
```diff
@@ -153,9 +153,9 @@ b {
 }
 
 @media (width >= 1280px) {
-  .function {
+  /* .function {
```
caseybaggz, line 156: Let's kill this code. We typically want to avoid pushing code we aren't using to keep the files clean.

docs/src/components/code-block/actions.tsx
Rykus0: I was getting errors that `styles` was undefined.  The wildcard syntax was what I found in the [Gatsby documentation](https://www.gatsbyjs.com/docs/how-to/styling/css-modules/#css-module-example).

Rykus0: I can look into it more - maybe we can go back

docs/src/components/frame/frame.module.css
Rykus0: Gotcha.  That's good to know.  In this case I just flattened the SCSS syntax, since it was throwing errors.

docs/src/components/motion/easing.module.css
Rykus0: Ah, yeah, I usually do too.  I forgot to clean this one up.

caseybaggz (approved):

docs/src/components/code-block/actions.tsx
caseybaggz: If it's in the docs, then don't worry about it. It's just an anti-pattern for imports...but...gatsby be gatsby... 🙈

PR #2096, opened 2022-02-08
fix(switch): remove readOnly attribute from input and add optional onChange handler

dannytannertantrum:
## PR Checklist

## PR Type

What kind of change does this PR introduce? It adds keyboard accessibility for Safari users interacting with the Switch component.

- [x] Bugfix

The Switch component is not currently accessible via keyboard navigation in Safari.

## What is the new behavior?

The new behavior adds keyboard accessibility for Safari users.


packages/switch/src/react/index.tsx
```diff
@@ -26,17 +28,18 @@ interface SwitchProps
...
 }
 
 interface SwitchComponent
   extends RefForwardingComponent<
-    SwitchProps,
-    HTMLInputElement | HTMLLabelElement,
-    SwitchStatics
-  > {}
+  SwitchProps,
+  HTMLInputElement | HTMLLabelElement,
```
dannytannertantrum, line 40: Ugh - sorry. My `format-on-save` switched that 🤦

packages/switch/src/react/index.tsx
```diff
@@ -108,7 +112,7 @@ const Switch = React.forwardRef<HTMLInputElement, SwitchProps>(
               `psds-theme--${themeName}`,
               `psds-switch__track--size-${size}`,
               checked &&
-                `psds-switch__track--checked psds-switch__track--color-${color}`
+              `psds-switch__track--checked psds-switch__track--color-${color}`
```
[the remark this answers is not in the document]
dannytannertantrum: Gah! Sorry about that, @Rykus0 - must have missed that with my auto-formatter.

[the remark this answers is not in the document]
Rykus0: No worries!  Thanks!

packages/switch/src/react/index.tsx
```diff
@@ -26,6 +28,7 @@ interface SwitchProps
   name?: string
   onBlur?: React.FocusEventHandler
   onClick?: (checked: boolean) => void
+  onChange?: DefaultChangeFn
```
caseybaggz, line 31: Instead of adding another prop to the interface, I would prefer we just kept it on the component as you have it, but just remove it from the props/interface. 

That way, the component API doesn't change and you don't have to worry about updating the docs as well.

dannytannertantrum: Sounds good - just pushed!

caseybaggz (approved):

Rykus0 (approved):

PR #2132, opened 2022-03-30
fix(dialog): Adding dark theme to dialog

js-park92:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

The current dialog ignores theme provider's theme setting and force light theme on all dialogs.

## What is the new behavior?

The new change will read the theme from theme provider and change the colors to appropriate colors.

## Other information

There is one accessibility check failing on dark theme, but it's because of the `<a>` tag color. I assumed that `<a>` tags should be replaced with `ps-design-system-link`, and that'd make it accessible. Let me know if I should change the `<a>` tag color to `--ps-colors-primary-action-text-dark-theme` in storybook.

Thanks! 🙏🏼

### DIFF
--- a/packages/dialog/src/css/index.css
+++ b/packages/dialog/src/css/index.css
@@ -7,7 +7,6 @@
 .psds-dialog {
   animation: psds-dialog__keyframes__fade var(--ps-motion-speed-fast) ease-out
     forwards;
-  background: var(--ps-colors-white);
   border-radius: 2px;
   box-shadow: 0 1px 2px rgba(0, 0, 0, 0.5);
   display: inline-flex;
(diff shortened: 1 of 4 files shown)


caseybaggz:
This was a branding decision for the pre-2022 brand and this change will create a massive breaking change for all the teams and overall product (from skills to flow). The Flow team also very specifically requires a white dialog until they move to the v2 branding.

This library is also officially in maintenance mode which means we are not adding any new features. We are only merging bug fixes or a11y enhancements.

Closing.

